PR #180, opened 2014-12-09
Drop sqlite attempt 2

### LINKED ISSUE #64: Remove sqlite from Crossbar core

wilbertom:
Drop `sqlite` by making the `PersistentCookieStorage` class save cookies into a file.

This PR is a follow up on #118 to close #64.

This code saves cookies to a `JSON` formatted file. Updates like `setAuth` are lazy in that they only save the to the file if a changed occurred. 

@oberstet if you see anything else you want changed let me know. Your last feedback on not doing the `temp` file magic was invaluable. I learned a lot from it. Sorry I took so long to finish this, contracting has been taking a lot of my spare time.

@malinoff thank you for the [comment](https://github.com/crossbario/crossbar/issues/64#issuecomment-64530161). I didn't know about the `shelve` module and it will save me a lot of trouble next time I have to do something similar. If we want to use it I'm up for refactoring for it.

If this gets merged and any bugs arise `@` ping me. I'll take care of it. Along this lines, do we have a unit testing convention in this project? Doing so would make me a lot more confident with the changes.

A side effect of this pull request is that we don't have to check for `_HAS_ADBAPI` when importing the class, it will always be available.

Another note on string formatting. In this little module I used `%` because, the module used it. However I just saw `.format` in another part of the project. If we prefer `.format` I can update the source code for it.

### DIFF
--- a/crossbar/crossbar/router/cookiestore.py
+++ b/crossbar/crossbar/router/cookiestore.py
@@ -26,13 +26,6 @@ from six.moves import http_cookies
 
 from twisted.python import log
 
-try:
-   from twisted.enterprise import adbapi
-   _HAS_ADBAPI = True
-except ImportError:
-   ## Twisted hasn't ported this to Python 3 yet
-   _HAS_ADBAPI = False
-
 
 from autobahn import util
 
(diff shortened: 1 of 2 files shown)


oberstet:
@wilbertom sorry for the sluggish response on my side .. well, better late than never;) thx for your contributions!

wilbertom:
@oberstet no problem! I know merging stuff is a big responsability. I'll be looking for another issue to fix this weekend.

PR #196, opened 2015-01-05
prevent value error

voidfuture:


oberstet:
Oh, good catch!

Would merge right away, but we need a CAA: https://github.com/crossbario/crossbar/wiki/Contributing-to-the-project

Sorry for the hassle .. needs only be done once though ..

voidfuture:
@oberstet sure... just sent over

oberstet:
got it! thx .. also for the fx!

voidfuture:
happy to contribute

PR #209, opened 2015-01-16
Fix bug into checkconfig.py

dimddev:
2015-01-16 17:28:21+0200 [Controller  14034]   File "/home/dimd/projects/middleware/env/local/lib/python2.7/site-packages/crossbar-0.10.0-py2.7.egg/crossbar/common/checkconfig.py", line 899, in check_paths
2015-01-16 17:28:21+0200 [Controller  14034]     if not _WEB_PATH_PAT_STR.match(p):
2015-01-16 17:28:21+0200 [Controller  14034] AttributeError: 'str' object has no attribute 'match'


oberstet:
Will be handled in https://github.com/crossbario/crossbar/issues/210

PR #218, opened 2015-01-23
RouterSession exception handling refactoring

Paranaix:
This pull request shall be seen as a hotfix and provides a minimalistic solution, this problem might require a better solution.

When a exception was thrown inside RouterSession.onHello and RouterSession.onAuthenticate, nothing but a print happend. Not only was the log output very simplistic and lacked crucial informations (e.g file name and line number) but it also caused the session to stay alive without any reply to the client.

### DIFF
--- a/crossbar/crossbar/router/router.py
+++ b/crossbar/crossbar/router/router.py
@@ -429,13 +429,12 @@ class RouterSession(FutureMixin, BaseSession):
          if self.debug:
             print("exception raised in onError callback: {0}".format(e)) 
 
-      self.leave(None, u"Internal server error")
+      self.leave(message = u"Internal server error")
 
     
    def onError(self, err):
       """
       Overwride for custom error handling.
-      Shall return a tuple (reason, message)
       """
       if self.debug:
          print("Catched exception during message processing: {0}".format(err.getTraceback())) # replace with proper logging
--- a/crossbar/crossbar/router/router.py
+++ b/crossbar/crossbar/router/router.py
@@ -429,7 +429,13 @@ class RouterSession(FutureMixin, BaseSession):
          if self.debug:
             print("exception raised in onError callback: {0}".format(e)) 
 
-      self.leave(message = u"Internal server error")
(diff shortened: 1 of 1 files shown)


oberstet:
Yep, there is an issue with not sufficiently processing errors here. The fix would look different, e.g. it exposes new API to override. We don't want that here.

Another general thing: we need a CAA to take in contributions https://github.com/crossbario/crossbar/wiki/Contributing-to-the-project

Paranaix:
The CAAs are out. 

I don't know how or even whether you want ot change the proposed solution (as you haven't elaborated much) but I would highly appreciate if you would take another look at error handling. This topic seems to me to have been rather ignored by you, as another related issue I found shows: Exceptions thrown in Container worker are completely unlogged and the Controller won't notice a thing.
Such issues are major if not critical and should not be ignored.

With a "better solution" I meant a thorough refactoring of the whole error handling meachanisms in order to unify. E.g with this patch there would be currently two error handling mechanisms in a single class. One can't prevent this, as it is simply not possible to delegate an exception from a Deferred to its iniator AFAIK. Thus one should consider adding a method which can be used in both cases. I don't know how exceptions inside ApplicationSessions are handled but I suspect the same issue solved with this pull request is also present there. Transports on the other hand have again they very own exception handling AFAIK. Therefor a thorough refactoring would primarily involve making error handling consistent over the whole architecture.

Besides that I don't really have time for this, I also believe that this exceeds my scope by far. As already stated, this pull request was meant as a hotfix.

PS: There is currently no (easy) way to set any of the debug properties of the RouterSession (whether directly codewise or with Crossbar). One could write his own factory or set in his own session but both those options should not be required. Besides that, the debug properties are currently not documented at all AFAIK.

oberstet:
I agree that error handling needs to improve. No question. Please give me some time to review this and come up with a solution that fits into the overall design. I do appreciate your detailed comments and your code. It's just touching deep down internals (and adds new exposed API) ..

PR #228, opened 2015-02-04
fixed wrong URIs

Paranaix:
The used URIs were wrong according to the WAMP documentation. Maybe having a dedicated file inside autobahn for all predfined URIs from the WAMP specification might be a good idea to prevent such mistakes. Its would also be quite convenient for autobahn users.

### DIFF
--- a/crossbar/crossbar/router/session.py
+++ b/crossbar/crossbar/router/session.py
@@ -522,14 +522,14 @@ class CrossbarRouterSession(RouterSession):
 
       ## dispatch session metaevent from WAMP AP
       ##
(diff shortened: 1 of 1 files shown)


oberstet:
I have recently started to clean up and extend the spec. And that involved changing URIs. But Crossbar.io is still doing the "old thing" (also documented https://github.com/crossbario/crossbar/wiki/Session-Metaevents).

As we are breaking API here, we should do it in one go - means to together with fixing/adding more meta stuff.

And even before this: finish the spec first on WAMP meta things.

PR #232, opened 2015-02-06
Allow dynamic authenticators to specify a authid

Paranaix:
Possible use cases of this feature:
* WAMPCRA authentication with an username but an internal (user) id as authid for simpler managment
* Ticket authentication without an authid for authentication (e.g read the authid from a db based on the ticket)

### DIFF
--- a/crossbar/crossbar/router/session.py
+++ b/crossbar/crossbar/router/session.py
@@ -166,11 +166,13 @@ class CrossbarRouterSession(RouterSession):
                            }
                            d = service_session.call(cfg['authenticator'], realm, details.authid, session_details)
 
-                           def on_authenticate_ok(user):
+                           def on_authenticate_ok(response):
+
+                              authid = response.get("authid", details.authid)
 
                               ## construct a pending WAMP-CRA authentication
                               ##
-                              self._pending_auth = PendingAuthWampCra(details.pending_session, details.authid, user['role'], u'dynamic', user['secret'].encode('utf8'))
+                              self._pending_auth = PendingAuthWampCra(details.pending_session, authid, response['role'], u'dynamic', response['secret'].encode('utf8'))
 
                               ## send challenge to client
                               ##
@@ -181,15 +183,15 @@ class CrossbarRouterSession(RouterSession):
                               ## when using salted passwords, provide the client with
                               ## the salt and the PBKDF2 parameters used
                               ##
-                              if 'salt' in user:
-                                 extra[u'salt'] = user['salt']
-                                 extra[u'iterations'] = user.get('iterations', 1000)
-                                 extra[u'keylen'] = user.get('keylen', 32)
+                              if 'salt' in response:
+                                 extra[u'salt'] = response['salt']
+                                 extra[u'iterations'] = response.get('iterations', 1000)
+                                 extra[u'keylen'] = response.get('keylen', 32)
 
                               return types.Challenge(u'wampcra', extra)
 
                            def on_authenticate_error(err):
-                           
+
                               error = None
                               message = "dynamic WAMP-CRA credential getter failed: {}".format(err)
 
@@ -390,9 +392,10 @@ class CrossbarRouterSession(RouterSession):
                service_session = self._router_factory.get(self._pending_auth.realm)._realm.session
                d = service_session.call(self._pending_auth.authprovider, self._pending_auth.realm, self._pending_auth.authid, signature)
 
-               def on_authenticate_ok(role):
-                  return types.Accept(authid = self._pending_auth.authid,
-                     authrole = role,
+               def on_authenticate_ok(response):
+                  authid = response.get("authid", self._pending_auth.authid)
+                  return types.Accept(authid = authid,
+                     authrole = response["role"],
                      authmethod = self._pending_auth.authmethod,
                      authprovider = self._pending_auth.authprovider)
 
@@ -403,7 +406,7 @@ class CrossbarRouterSession(RouterSession):
                   if isinstance(err.value, ApplicationError):
                      error = err.value.error
                      if err.value.args and len(err.value.args):
-                        message = err.value.args[0]
+                        message = str(err.value.args[0])
 
                   return types.Deny(error, message)
 
--- a/crossbar/crossbar/router/session.py
+++ b/crossbar/crossbar/router/session.py
@@ -393,9 +393,15 @@ class CrossbarRouterSession(RouterSession):
                d = service_session.call(self._pending_auth.authprovider, self._pending_auth.realm, self._pending_auth.authid, signature)
 
                def on_authenticate_ok(response):
-                  authid = response.get("authid", self._pending_auth.authid)
+                  if isinstance(response, dict):
+                     authid = response.get("authid", self._pending_auth.authid)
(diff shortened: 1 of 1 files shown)


oberstet:
This changes the API, and the authid should be part of the challenge already

Paranaix:
Would you mind elaborating a bit more? Authentication is the process where I acquire and validate the identity of a user. What you just stated does not just violates this basic principle but it also renders the WELCOME message completely useless. Could you maybe tell me, how you think above use cases should be done without completely ignoring WAMPs authnetications model by doing it completely on our own?

oberstet:
The `authid` is an ID already provided by the client when the client starts authentication. It is the ID the client wants to be authenticated for. Like when you login to something, you say "here is my name, pls check that I am in fact the one". It is not something a backend dynamically assigns.

> WAMPCRA authentication with an username but an internal (user) id as authid for simpler managment

=> use `username` as `authid`.

> Ticket authentication without an authid for authentication (e.g read the authid from a db based on the ticket)

=> not supported today

We can continue discussion if you file an issue. I want to close this PR for now since I'm gonna move the whole codebase to PEP8 (as I just did with AutobahnPython) - it won't merge afterwards anyway ...

Thanks for critical input and contributions though!

Paranaix:
First of all: As I would implement this feature again after the refactoring and file in a new pull request anyways, I see little point in describing everything again in an issue and loosing the whole history. I think this PR is just fine for discussion and tfiguring things out.

I know what authentication is, you dont have to explain this to me. What you miss is the fact that this feature has very little to do with the actual authentication (that is validating the identity of an entity). We just assign an alias to somone who is already authenticated, like "Hey your name is dingle dangle dons dans? I will just call you dingle for short". An application can in fact do this herself (e.g by keeping a mapping in a database), but thats introduces alot of trouble and overhead. As your answere shows you haven't even fully understood why such a feature might even useful. Im well aware of the the fact that authid is currently e.g the usenrame for wampcra and that one can work with this. However this is simply inconvenient. Entities are usually referenced by some kind of identifier. While an username is for sure an identifier as well, nearly every application out there will primarly use simple integers or maybe even UUIDs (for good reasons). When designing services one now has to provide two APIs: one that accepts the primarly used identifiers and one that accepts usernames. To make it plain: This sucks if we could easily convert the more troublesome username into a simple id once and for all during authentication and work with what ever we want instead.

Besides those use cases mentioned already by me, there are of course alot more. I think one of the most obvious is the integration of 3rd party authentication (e.g via facebook, google etc.). Once completed you probably dont want to keep some totally irrelevant userid of a 3rd party service or a token but your internal id. This can even be insecure. Another use case could e.g be the generation of ids in highly distributed systems with alot of peers (roboter swarms for example).

What I simply can not understand is your refusal to implement this feature. This is nothing very complicated and nothing which changes a lot of things, in fact the WELCOME message already supports this. The WAMP spec also does not state that you mustn't do something like this, One can easily solve a lot of different problems with such a feature which is easily implemented. Why do you want to restrict developers in such a drastic way? Its totally up to the server(/application) how he wants to call and organizes its users/peers. Hell he could even call them `Idiot1`, `Idiot2` etc. The problem however is, that while the router is part of our architecture, you can't really modify it to ones liking. Therefor the router has to provide an API for such a feature an that exactly what is introduced with this PR. While I can understand that you might have concerns about the actual API, I can't really understand why you oppose the feature itself so vehemently.

oberstet:
@Paranaix Cool down;) My problem is: the change you propose is a change in API of Crossbar.io. There are people using this stuff for their apps - which then break. Further, Crossbar.io is not the only WAMP router, so we need to make sure the spec has it covered. After discussion with other WAMP implementors.

The WAMP spec requires `authid` in WAMP-CRA in the `HELLO` message: https://github.com/tavendo/WAMP/blob/master/spec/advanced.md#challenge-response-authentication

Paranaix:
Based on your comment im not sure whether you have noticed but my latest commit fixed backward compatibility.

The WAMP spec is in fact not violated, nor changed in any way. The only things this PR should enable is to firstly change the authid sent in the `WELCOME` message and secondly change the associated authid of a RouterSession. This is not a thing other routers MUST implement but rather an (additional) feature.

As you explicitely mentioned the `HELLO` message i will shortly outline the whole authentication process with this PR:

1) Peer sends `HELLO` with username
2) Server responds with CHALLENGE, the contained authid will be changed according to the authenticator, as required from the spec (`The authentication ID the client will be authenticated as when the authentication succeeds.`)
3) If the signatures matchs the server will send a `WELCOME` message with the custum authid once he received the signature.

This is however only optional. An authenticator doesn't have to pass an authid to the router. In that case wampcra just proceeds regulary and the requested username from the client will be used as final authid.

oberstet:
Ah, yes. That's right. With WAMP-CRA, it doesn't change the spec (the text there already allows to do that). It does _extend_ the Crossbar.io API for WAMP-CRA dynamic, but upwards compatible.

I have transcribed your proposal (for WAMP-CRA) to the current code:

https://github.com/crossbario/crossbar/commit/ac1e30c9282d424d1b839f9d335c436dc8684249

The docs need update now https://github.com/crossbario/crossbar/wiki/WAMP-CRA-Authentication#dynamic-credentials

Plus: I know, you want the same for ticket ..

oberstet:
alright: https://github.com/crossbario/crossbar/commit/590185e1d0a5c59fb128ec03c2b0d5bbe89986a5

Paranaix:
Thank you very much, this was quite important for me.

Notice that there is a small leftover though: https://github.com/crossbario/crossbar/commit/590185e1d0a5c59fb128ec03c2b0d5bbe89986a5#commitcomment-9727032

oberstet:
Yes, thanks. Removed.

I am wondering about the following though:

With WAMP-CRA, the `CHALLENGE` message already contains the `authid` that will be effective when authentication later succeeds.

With WAMP-Ticket, the `CHALLENGE` message is something like `[4, "ticket", {}]` and does NOT contain an `authid` - neither any provided in `HELLO`, nor the one that will effective later.

oberstet:
Well, no, forget it. WAMP-CRA does provide the `authid` only in the _serialized_ challenge (which happens to be JSON serialized so 3-legged auth does work). All fine.

Paranaix:
Wouldn't this require changing the spec? At some point I got the feeling that it might be more generic having this whole mechanism separate from the actual authentication (e.g by calling another uri which performs the mapping). That would make the API more uniform. However this comes with two issues: We cant include the authid in the challenge and a small performance penality if we go with the simplest way and would call the uri just in `onAuthenticate`. If we want to call it already in `ònHello` the implementation would be alot more complicated (must keep an internal state etc.).

I think the actual question is whether the authid contained in the challenge should be the requested or the assigned authid. For a simple peer its not really important IMO as he shouldn't need to parse it anyways. He gets informed about his actual authid by the `WELCOME` message nevertheless. Im more concerned about three-legged-authentication. Another service might not be able to deal with an internal id (which really could be anything). It might thus make more sense, to maybe include the requested authid in the challenge. 

However this something which I really believe is part of the spec and should maybe discussed by a broader audience. Can you give any insight on how that particular definition was chosen? Was there already a discussion about this topic?

PR #235, opened 2015-02-08
Fixed assertion on (service) wamp call

Paranaix:
A wamp call from the router failed with an assertion if the calle requested to disclose callers because the role of the internal service session was an ascii string.


oberstet:
oh, thx! good catch.

Issue #251, opened 2015-02-18
Cache control on static web resources


taion:
Currently static resources are served via `t.w.s.File`. This doesn't set any response headers related to caching, which means that most browsers end up caching the static resources indefinitely.

In general it'd be nice to have some non-infinite cache timeout on files. Especially when running with `--debug` it'd be ideal to disallow caching static assets by the browser entirely.

oberstet:
Yes, I agree. Cache control headers would be useful ..

taion:
What do you think is an appropriate default cache timeout? Flask uses 12 hours by default for static assets (even when running in debug mode, apparently, which I wasn't previously aware of).

oberstet:
@taion I sane finit default seems to be better than infinite. But we could do even better I think (make it configurable)

Static file serving resources are created here:

* https://github.com/crossbario/crossbar/blob/master/crossbar/crossbar/worker/router.py#L660
* https://github.com/crossbario/crossbar/blob/master/crossbar/crossbar/worker/router.py#L870

Config for static service is checked here:

* https://github.com/crossbario/crossbar/blob/master/crossbar/crossbar/common/checkconfig.py#L594

Settings cache headers is done like this:

* https://github.com/tavendo/AutobahnPython/blob/master/autobahn/autobahn/twisted/longpoll.py#L589

We could have an option that is an integer: If negative, no caching at all. If 0, infinite caching. Else, caching time in seconds or what. Probably steal ideas of what Nginx or Apache allows here ..

taion:
That was my plan. I was actually going to go with 0 for no caching (cache for 0 seconds) and negative (or maybe null) for infinite duration caching.

I believe it's considered best practice to set both the `Expires` and `Cache-Control` headers, though.

I'll get a PR up when I get a chance.

oberstet:
@taion awesome! yeah, I think your value mapping makes more sense (less suprise for users). `null` instead of negative is what I like even more. and when cache value is missing in the config, we can use a sane finite default (like some hours or what)

shaworth:
I like where this is headed.  I would argue that a missing value should
result in no caching.  IMHO this is less likely to trip up new developers.

taion:
I don't really care that much what the default is. I think 12 hours is reasonable, because that's what Flask does. Using 0 would be dangerous in production, but you should probably use nginx or something to serve static assets anyway. The unlimited caching is just causing issues for me right now.

taion:
Thanks for merging!

PR #255, opened 2015-02-20
Add cache control headers for static resources

taion:
For crossbario/crossbar#251

As discussed:

* `cache_timeout` is specified in seconds
* Default is 12 hours
* There's no special-casing for 0 timeout (i.e. I don't explicitly set `no-cache`) but you should get essentially no caching when `cache_timeout` is set to 0
* `null`/`None` preserves the current behavior, which sets no cache control headers, and lets the browser cache for as long as it wants

### DIFF
--- a/crossbar/crossbar/common/checkconfig.py
+++ b/crossbar/crossbar/common/checkconfig.py
@@ -594,6 +594,7 @@ def check_web_path_service_static(config):
         check_dict_args({
             'enable_directory_listing': (False, [bool]),
             'mime_types': (False, [dict]),
+            'cache_timeout': (False, list(six.integer_types) + [type(None)])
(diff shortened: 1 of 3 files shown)


taion:
This is a bit fiddlier than I thought, with the `createSimilarFile` override. It probably makes sense to add a test case for this, but I couldn't see anything set up that would just let me plug one in, and I don't quite have the time at the moment to set up something more extensive.

oberstet:
Thanks for working on this! What is the `createSimilarFile` doing / why is it needed?

taion:
Added a comment documenting why the `createSimilarFile` override is necessary. `t.w.s.File` makes a near-copy of itself to serve up files under the directory, and we need to set the cache timeout on the copy as well.

PR #265, opened 2015-03-10
Further REST bridge testing

hawkowl:


hawkowl:
Getting to a good place :) 

```
crossbar/adapter/rest/caller                   31      1      4      1    94.29%
crossbar/adapter/rest/common                  121      3     56      1    97.74%
crossbar/adapter/rest/publisher                35      5      4      1    84.62%
```

PR #266, opened 2015-03-13
Remaining two WAMP roles for the REST bridge

hawkowl:


hawkowl:
@oberstet This implements the other two roles, with the first idea for settings as per the email. I think the possibility for a different URL per message will be more flexible in the end.

I've also got a better test transport -- I have not tested "wildcard" publishing with it, though. To test the wildcard section of the config, I'll have to flesh that out (and figure out where in Crossbar/Autobahn such expansion is done).

I'm not happy with the naming of the two new classes, though. I picked some names that describe it, but it might be better to replace them with more consistent ones.

Docs/examples are forthcoming but will be my focus next week :)

crossbar/adapter/postgres/__init__.py
```diff
@@ -27,3 +27,9 @@
 #  with this program. If not, see <http://www.gnu.org/licenses/agpl-3.0.en.html>.
```
oberstet: I there a specific reason to import, and export things again in `__init__.py`?

hawkowl: It makes the class path shorter + friendlier :)

Issue #268, opened 2015-03-16
Support Python 3


oberstet:
For users that want to run Crossbar.io as an embedded part of their applications, the fact that Crossbar.io currently only runs on Python 2 introduces pain when their application supports only Python 3, or when it supports both Python 2 and 3.

The reason that Crossbar.io right now only supports Python 2 is that Crossbar.io depends on Twisted, and Twisted porting to Python 3 is unfinished. Most of the parts that Crossbar.io uses from Twisted are ported already (e.g. I know that the routing core will run perfectly on Twisted / Python 3), but other auxiliary parts are missing (mostly process and web related things).

> Note that AutobahnPython supports Python 2 and 3 as well as Twisted and asyncio. So users can of course write WAMP app components in any of the former and use them with Crossbar.io. Users can run such components by starting them independently or let them be started and monitored by Crossbar.io as *guest workers*. What user currently *can't* do is run such components in Crossbar.io *container workers* or side-by-side with router in *router workers*. This is only possible when then app component is written for Twisted and runs from the same Python as Crossbar.io - and that currently means Python 2.

So the task is:

* **port missing parts from Twisted used by Crossbar.io to Python 3**
* fix up any Python 2 relicts in Crossbar.io itself

Note that the former does not necessarily means port complete Twisted to Python 3. The latter should be only some occasional, minor stuff like print's or such.

avanov:
Upvote this ticket. I'd like to see the functionality ported as soon as possible. Thanks for the efforts!

hawkowl:
@avanov Hi! I've been hard at work on porting the sections of Twisted that Crossbar uses to Python 3 -- so it is very much underway. :)

If you would like to assist, there are several tickets in code review on the Twisted issue tracker -- https://twistedmatrix.com/trac/ticket/5987, https://twistedmatrix.com/trac/ticket/6176, and https://twistedmatrix.com/trac/ticket/6177 are the ones blocking the rest being submitted for review. Once these are reviewed and issues fixed, the next batch can be put up for review and can make it into a released Twisted which Crossbar can use.

We're (Twisted) always looking for new reviewers, and anyone can review -- and if you have Python 3 experience, it would be invaluable for these porting reviews. :)

hawkowl:
@oberstet This is more or less done, I think -- there's some small things not ported (like the REST bridge), but Crossbar's unit tests pass on 3.3+3.4, and it's now on Travis, so it'll keep working.

Should we close this ticket?

[...]

PR #269, opened 2015-03-17
Get Crossbar mostly importing on Python 3

hawkowl:
ref #268

### DIFF
--- a/crossbar/adapter/rest/test/request_mock.py
+++ b/crossbar/adapter/rest/test/request_mock.py
@@ -21,7 +21,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from StringIO import StringIO
+from io import BytesIO as StringIO
 
 from twisted.web import server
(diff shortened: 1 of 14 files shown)


crossbar/adapter/rest/test/request_mock.py
```diff
@@ -47,16 +47,16 @@ def _render(resource, request):
...
 
-def _requestMock(path, method="GET", host="localhost", port=8080, isSecure=False,
+def _requestMock(path, method=b"GET", host=b"localhost", port=8080, isSecure=False,
                  body=None, headers=None, args=None, reactor=None):
 
     if not headers:
         headers = {}
 
-    headers["Date"] = ["Tue, 01 Jan 2014 01:01:01 GMT"]
+    headers[b"Date"] = ["Tue, 01 Jan 2014 01:01:01 GMT"]
```
Zopieux: Shouldn't the date value be `bytes` too?

hawkowl: Good catch... nothing in Crossbar uses the HTTP date so this slipped through, I'll fix it next py3 pass.

Issue #253, opened 2015-02-19
Way to check if a process is really a crossbar node


Paranaix:
It sometimes happens that crossbar exists uncleanly and leaves the *.pid* file. This would actually not be a big deal if not other unrelated processes can get the same PID resulting in crossbar unable to restart as it assumes that there is still a node running (this already occured multiple times to me!). An user then has to remove the *.pid* file manually.

To fix this issue I suggest to implement a way to check whether a process is really a crossbar node.
https://github.com/giampaolo/psutil might provide the required functionality.

herm:
This is a big problem when trying to start crossbar using systemd. After an unclean shutdown crossbar sometimes starts and sometimes doesn't start. As the order in which services are started is not fixed  sometimes a service with the same pid crossbar had last time is already running.  

Test case: 
```
echo '{"pid": 1}' > .crossbar/node.pid 
crossbar start
Crossbar.io is already running from node directory /root/crossbar-apps/.crossbar (PID 1).
```

Paranaix:
After considering this a bit more I think we also need a way to check whether a process is a SPECIFIC crossbar node. This need arises if we got multiple nodes running on the same host. Maybe using some form of IPC is maybe better as we might need it for other features as well.

oberstet:
This is how it works right now: upon startup, Crossbar.io checks if there is a PID file. If there is, it checks if there is a process with PID equal to the one written in the PID file. If there is, it assumes that Crossbar.io already runs.

Is there a way to get the executable of a process running under a given PID? This would allow us to improve robustness.

meejah:
/proc/PID/exe links to the executable running, but that's not cross-platform of course. And this will just be the python interpreter. ``psutil`` includes ability to examine the command-line -- the second arg should be the path to the crossbar script. I'm not sure how cross-platform this may be, though.

hawkowl:
@meejah can we close this, it seems to work on trunk?

meejah:
Yes, this can be closed I think (for the record, #322 was introduced by this enhancement, but that was fixed already at least for the linux platforms I tested).

KrishnaPG:
Facing this problem in the latest docker image:

![image](https://cloud.githubusercontent.com/assets/4610831/16520340/1e271512-3fad-11e6-81bb-7a40ed32b34d.png)

sebastianwahn:
I found a workaround for the problem occurring in Docker:

[...]

PR #284, opened 2015-03-30
Double-checks that the PID is really running crossbar

meejah:
Possible fix for #253 

This only works when psutil is installed, and examines the
commend-line to determine if the second thing is a crossbar
script. The command-lines (on Linux, anyway) end up looking
like:

  .../python ..../crossbar other args

Should be tested on other platforms.


oberstet:
ah, nice!

Issue #278, opened 2015-03-25
SIGINT sometimes handled twice upon Ctrl+C


estan:
Most of the time I get a nice

```
^C2015-03-25 10:32:57+0100 [Controller  31863] Received SIGINT, shutting down.
2015-03-25 10:32:57+0100 [Controller  31863] Main loop terminated.
```

upon stopping the node with Ctrl+C.

But sometimes I get

```
^C2015-03-25 10:32:10+0100 [Controller  31833] Received SIGINT, shutting down.
2015-03-25 10:32:10+0100 [Container   31842] Received SIGINT, shutting down.
2015-03-25 10:32:10+0100 [Container   31842] Connection to node controller lost.
2015-03-25 10:32:10+0100 [Container   31842] Unhandled Error
2015-03-25 10:32:10+0100 [Container   31842] Traceback (most recent call last):
2015-03-25 10:32:10+0100 [Container   31842] Failure: exceptions.RuntimeError: No transport, but disconnect() called.
2015-03-25 10:32:10+0100 [Container   31842] 'While firing onLeave'
2015-03-25 10:32:10+0100 [Controller  31833] Main loop terminated.
```

Notice the *two* `Received SIGINT, shutting down.` messages. It seems the handler for `SIGINT` runs twice for some reason, which I guess is the reason for the subsequent `RuntimeError`. Sometimes this double-handling results in

```
^C2015-03-25 10:31:36+0100 [Controller  31801] Received SIGINT, shutting down.
2015-03-25 10:31:36+0100 [Container   31810] Received SIGINT, shutting down.
2015-03-25 10:31:36+0100 [Container   31810] Connection to node controller lost.
2015-03-25 10:31:36+0100 [Container   31810] Unhandled Error
2015-03-25 10:31:36+0100 [Container   31810] Traceback (most recent call last):
2015-03-25 10:31:36+0100 [Container   31810] Failure: exceptions.RuntimeError: No transport, but disconnect() called.
2015-03-25 10:31:36+0100 [Container   31810] 'While firing onLeave'
2015-03-25 10:31:36+0100 [Container   31810] Unhandled Error
2015-03-25 10:31:36+0100 [Container   31810] Traceback (most recent call last):
2015-03-25 10:31:36+0100 [Container   31810] File "/usr/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
[...]

PR #285, opened 2015-03-31
Process exit cleanup

meejah:
The previous patch for #278 isn't sufficient in cases where the router process has already died when the controller shuts down. It would also throw an extra exception if the reactor was already stopping in loseConnection().


estan:
I'm not that well versed in Twisted, so take this with a grain of salt: Shouldn't the `cleanup_worker` somehow return a deferred, one that is called back on once the worker process finally exits, to hold off the controller reactor stop until that happens?

estan:
I would also expect something similar to be done for guests? (This is just for native workers right?)

PR #292, opened 2015-04-03
single serializer passed to WampRawSocketServerFactory

tigeraniya:
need to pass serializer as iterator to WampRawSocketServerFactory


oberstet:
RawSocket revision is finished and this no longer applies

PR #311, opened 2015-04-20
Component stopping and on_component_stop

meejah:
This fixes up a few things relating to stopping components:

 - on_component_stop published whenever a connection to a component dies
 - log errors when instantiating a component fails
 - proper shutdown in stop_container
 - while doing "static config initialization", if any component dies within 2 seconds, we shutdown crossbar
 - if details.caller was None, this wasn't dealt with properly (could rebase this out; not strictly related...)

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -215,7 +215,6 @@ class Node:
         call_options = CallOptions(disclose_me=True)
 
         for worker in config.get('workers', []):
-
             # worker ID, type and logname
             #
             if 'id' in worker:
@@ -365,6 +364,21 @@ class Node:
 
                     component_no = 1
 
+                    # if components exit "very soon after" we try to
+                    # start them, we consider that a failure and shut
+                    # our node down. We remove this subscription 2
+                    # seconds after we're done starting everything
+                    # (see below). This is necessary as
+                    # start_container_component returns as soon as
+                    # we've established a connection to the component
+                    def component_exited(info):
+                        dead_comp = info['id']
+                        log.msg("Component '{}' failed to start; shutting down node.".format(dead_comp))
+                        if self._reactor.running:
+                            self._reactor.stop()
+                    topic = 'crossbar.node.{}.worker.{}.container.on_component_stop'.format(self._node_id, worker_id)
+                    component_stop_sub = yield self._controller.subscribe(component_exited, topic)
+
                     for component in worker.get('components', []):
 
                         if 'id' in component:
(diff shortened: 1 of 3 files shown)


crossbar/controller/process.py
```diff
@@ -528,7 +528,10 @@ def on_ready_success(id):
                 'who': worker.who
             }
 
-            self.publish(started_topic, started_info, options=PublishOptions(exclude=[details.caller]))
+            options = None
+            if details.caller is not None:
+                options = PublishOptions(exclude=[details.caller])
+            self.publish(started_topic, started_info, options=options)
```
meejah: Might be nicer to just filter out ``None`` entries from the ``exclude`` list in Autobahn|Python

crossbar/worker/container.py
```diff
@@ -235,9 +238,15 @@ def start_container_component(self, id, config, reload_modules=False, details=No
...
+        # ultimately, this gets called once the connection is
+        # establised, from onOpen in autobahn/wamp/websocket.py:59
         def create_session():
-            return create_component(componentcfg)
+            try:
+                return create_component(componentcfg)
+            except Exception as e:
+                # AutobahnPython swallows exceptions from onOpen
+                log.err(_why="Instantiating component failed")
+                raise
```
meejah: We could, I think, just call create_component() right here, and pass it to the factory via ``lambda: component`` instead of giving it ``create_session``; then we would get any instantiation errors immediately (and could, at least for this particular use-case, get rid of the 2-second stuff in controller/node.py)

crossbar/controller/node.py
```diff
@@ -365,6 +364,21 @@ def _run_node_config(self, config):
...
+                    # we've established a connection to the component
+                    def component_exited(info):
+                        dead_comp = info['id']
+                        log.msg("Component '{}' failed to start; shutting down node.".format(dead_comp))
+                        if self._reactor.running:
+                            self._reactor.stop()
+                    topic = 'crossbar.node.{}.worker.{}.container.on_component_stop'.format(self._node_id, worker_id)
+                    component_stop_sub = yield self._controller.subscribe(component_exited, topic)
+
                     for component in worker.get('components', []):
```
meejah: This smells a bit hacky, but handles any case where the component dies very shortly after starting...

crossbar/worker/container.py
```diff
@@ -235,9 +237,15 @@ def start_container_component(self, id, config, reload_modules=False, details=No
...
+        # ultimately, this gets called once the connection is
+        # establised, from onOpen in autobahn/wamp/websocket.py:59
         def create_session():
-            return create_component(componentcfg)
+            try:
+                return create_component(componentcfg)
+            except Exception:
+                # AutobahnPython swallows exceptions from onOpen
+                log.err(_why="Instantiating component failed")
+                raise
```
meejah: Instead of the "2 second" thing in node.py, we could instead call `create_component()` right away and have `create_session` be a closure that just returns the instance; then we could fail right away on instantiation problems. This would work for constructor problems, but not anything else.

oberstet:
something went wrong .. mmh. merge conflict ..

crossbar/controller/process.py
meejah: I'll rebase and re-submit, as code like this can now be simplified back to what it was thanks to recent AutobahnPython merge.

Issue #316, opened 2015-05-05
Details passed to wamp.session.on_join subscribed asyncio method have bytes keys


meejah:
As reported by @terrycojones in the Autobahn|Python repository [ticket 338](https://github.com/tavendo/AutobahnPython/issues/388):

If I have a python2 router session that subscribes to `wamp.session.on_join`, the associated method is called with a details `dict` whose keys are strings.

However, I just converted the same code to be a python3 session and was surprised to get `KeyErrors`. It turns out that the details passed to the method have bytes keys:

```
{b'authid': 'coordinator',
 b'authmethod': 'wampcra',
 b'authprovider': 'dynamic',
 b'authrole': 'coordinator',
 b'realm': 'realm1',
 b'session': 978859022,
 b'transport': {b'http_headers_received': {'cache-control': 'no-cache',
                                           'connection': 'Upgrade',
                                           'host': '127.0.0.1:8080',
                                           'pragma': 'no-cache',
                                           'sec-websocket-key': 'Zb3bDO+efx133r1EOMriQQ==',
                                           'sec-websocket-protocol': 'wamp.2.msgpack.batched,wamp.2.msgpack,wamp.2.json.batched,wamp.2.json',
                                           'sec-websocket-version': '13',
                                           'upgrade': 'WebSocket',
                                           'user-agent': 'AutobahnPython/0.10.3'},
                b'http_headers_sent': {},
                b'peer': b'tcp4:127.0.0.1:64774',
                b'protocol': b'wamp.2.msgpack.batched',
                b'type': b'websocket'}}
```

I could probably fix this, but I don't know where to look. Grepping for `wamp.session.on_join` and `on_join` didn't turn up anything obvious.

meejah:
Hmm, actually this only happens using msgpack; json serializers are fine.

terrycojones:
Hi @meejah, thanks for the reply.

I don't think I'm using msgpack.  My code is a slight simplification of that found in the example at https://github.com/tavendo/AutobahnPython/blob/master/examples/asyncio/wamp/basic/client.py with the "websocket" transport. Roughly:

```python
from autobahn.asyncio.wamp import ApplicationSessionFactory
from autobahn.asyncio.websocket import WampWebSocketClientFactory
from autobahn.wamp import types

session_factory = ApplicationSessionFactory(types.ComponentConfig(realm=args.realm))
session_factory.session = MyComponent  # My session app.

[...]

PR #317, opened 2015-05-05
Unicode dict keys

meejah:
A unit-test of sorts for #316 and switch the dict keys for anything we publish to be unicode strings.

I'm not 100% sure this is the *right* fix, now that I look at the diff.

WAMP says dicts should have string keys, so perhaps this really was an Autobahn|Python bug after all? That is, a trip through the serializers should have made all they keys the appropriate string type for a Python3 client instead of bytes()...

### DIFF
--- a/crossbar/router/test/test_router.py
+++ b/crossbar/router/test/test_router.py
@@ -64,7 +64,6 @@ class TestCrossbarSessions(unittest.TestCase):
         session.onJoin(details)
 
         calls = session._router._realm.session.method_calls
-        print(calls, calls[0][1])
         self.assertEqual(1, len(calls))
         for (k, v) in six.iteritems(calls[0][1][1]):
             self.assertTrue(type(k) == six.text_type)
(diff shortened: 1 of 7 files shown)


crossbar/router/test/test_router.py
```diff
@@ -32,13 +32,42 @@
...
+        session._router._realm = Mock()
+        session._router._realm.session = Mock()
+        session._transport = Mock()
+        session._transport._transport_info = 'nothing to see here'
+        details = types.SessionDetails(u'test_realm', 1234)
+
+        session.onJoin(details)
+
+        calls = session._router._realm.session.method_calls
+        print(calls, calls[0][1])
```
hawkowl: Maybe don't need the print :)

meejah: Ah, nice catch. See also a different (more robust) fix, in autobahn under [PR 390](https://github.com/tavendo/AutobahnPython/pull/390).

Issue #322, opened 2015-05-10
Process 'crossbar-controller' not recognised as a crossbar process.


jwaterhouse:
I'm using version 0.10.4 installed from pip, on CentOS 7.

After running crossbar and then attempting to get the status, I get the following error:
```
"/home/jon/dev/wamp-router/.crossbar/node.pid" points to PID 6744 which is not a crossbar process:
  crossbar-controller
Verify manually and either kill 6744 or delete /home/jon/dev/wamp-router/.crossbar/node.pid
No Crossbar.io instance is currently running from node directory /home/jon/dev/wamp-router/.crossbar.
```

Crossbar runs fine, but the error prevents getting the status or stopping the service through cli.

Seems the offending like is crossbar/controller/cli.py, line 121. If statement checks the PID's process name does not end with 'crossbar', which succeeds as the process is apparently named 'crossbar-controller'.

jwaterhouse:
Actually with a little more investigation, it seems to fail when checking the number of args. It's expecting at least 2, but my cmdline only contains a single element, the process name: `['crossbar-controller']`.

If I change the code to just look at the first element in cmdline, and match it to 'crossbar-controller', it appears to fix the issue above.

Submitted a fix here: https://github.com/crossbario/crossbar/pull/323

s-trooper:
Ubuntu 14.04 exactly the same problem. 
Any workaround so i can use crossbar in production? Should i kill crossbar-controller and remove node.pid manually is that the way to go?

meejah:
Yes, you can just kill the process and the start will detect the stale node.pid

meejah:
What if we relax the checking to just be: `"crossbar" in process_name`?

The original reason this was changed was so we didn't accidentally kill a process that unluckily re-used the PID found in a stale node.pid and the above relaxed check seems more than sufficient for this purpose to me...

meejah:
#338 should fix this; if that works for you use-case can you please close this issue?
thanks.

jwaterhouse:
Sorry for the late response, but the latest code does appear to fix this issue. Thanks heaps!

PR #323, opened 2015-05-10
Fix for issue #322

jwaterhouse:
Fixed check of process name working incorrectly, and did a full string comparison for the process name, seems more exact than using startswith/endswith?


oberstet:
This might be platform specific .. maybe we should rather relax identifying a crossbar process. not sure .. needs testing.

However, in the meantime, for merging stuff, we would need a CLA:

* https://github.com/crossbario/crossbar/blob/master/legal/contributors.md 
* http://crossbar.io/docs/Contributing-to-the-project/

meejah:
@jwaterhouse can you confirm that #338 fixes this issue for you?
Thanks.

PR #325, opened 2015-05-11
Further Python 3 work

hawkowl:
This gets Crossbar starting with a basic websocket routing node + static serving from a Twisted checkout with some in-flight patches applied.

(Still not final, but it's getting there :) )

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -35,6 +35,8 @@ import json
 import re
 import six
 
+from twisted.python.compat import unicode
+
 from pprint import pformat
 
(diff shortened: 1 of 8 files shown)


hawkowl:
cc @oberstet

crossbar/common/checkconfig.py
```diff
@@ -35,6 +35,8 @@
 import re
 import six
 
+from twisted.python.compat import unicode
+
```
meejah: looks unused?

hawkowl: Something else down the file uses `unicode` :)

PR #333, opened 2015-05-23
More Python 3 porting 

hawkowl:
This also improves the tox.ini by making it get the dependencies from the current setup.py, not the one published on PyPI.

### DIFF
--- a/crossbar/router/observation.py
+++ b/crossbar/router/observation.py
@@ -31,7 +31,7 @@
 from __future__ import absolute_import
 
 from pytrie import StringTrie
-from ordered_set import OrderedSet
+from oset import oset as OrderedSet
 
 from autobahn import util
 
--- a/crossbar/router/observation.py
+++ b/crossbar/router/observation.py
@@ -31,7 +31,6 @@
 from __future__ import absolute_import
 
 from pytrie import StringTrie
-from collections import MutableSet
 
 from autobahn import util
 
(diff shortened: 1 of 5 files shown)


oberstet:
why a different ordered set? I just removed it;)

hawkowl:
I stopped using the other one when I realised why it was failing :)

oberstet:
well, we don't need any external dep for that .. I am wondering why this PR isn't in conflict, since I just yesterday removed the other ordered_set. anyway, I'm not gonna merge it like this

hawkowl:
I merged it forward, to get rid of the dependency :) -- the diff shouldn't have it there.

crossbar/router/observation.py
```diff
@@ -43,8 +43,8 @@ def is_protected_uri(uri):
 
 class OrderedSet(set):
 
-    def __init__(self, iter=None):
-        super(set, self).__init__(iter)
+    def __init__(self):
```
oberstet: why?

oberstet:
changes to tox.ini: are you sure these changes work?

crossbar/router/observation.py
hawkowl: set() on Python 3 works differently, and we don't instantiate it with any arguments in the codebase (it also doesn't put iter in `self._list`).

hawkowl:
I ran them all locally, and Travis is giving all-green. `skip_install` under flake8 requires a recent tox (1.9, current is 2.0.1), but it makes it run massively quicker (11s Travis job for flake8, master takes 54sec).

PR #339, opened 2015-05-27
More logging improvements

hawkowl:
This implements:
- Loglevels being applied to log file output (Py2 only, can't write to log output on Py3)
- Worker containers communicating logs by JSON through stdout, to give richer and multi-line logs
- "trace" debug entries, usable by ``--loglevel=trace``, which outputs things like WAMP calls routed, for super debugging.

cc: @oberstet

### DIFF
--- a/crossbar/_logging.py
+++ b/crossbar/_logging.py
@@ -78,6 +78,9 @@ def makeStandardOutObserver(levels=(LogLevel.info, LogLevel.debug),
     @provider(ILogObserver)
     def StandardOutObserver(event):
 
+        if event.get("cb_level") == "trace":
+            return
+
         if event["log_level"] not in levels:
             return
 
--- a/crossbar/_logging.py
+++ b/crossbar/_logging.py
@@ -32,11 +32,13 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import sys
+import json
 
 from zope.interface import provider
 
 from twisted.logger import ILogObserver, formatEvent, Logger, LogPublisher
 from twisted.logger import LogLevel, globalLogBeginner, formatTime
+from twisted.logger import FileLogObserver, eventAsJSON
 
 from twisted.python.reflect import qual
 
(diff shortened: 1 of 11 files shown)


crossbar/_logging.py
```diff
@@ -150,6 +153,70 @@ def StandardErrorObserver(event):
...
+
+    recordSeparator = u"\x1e"
+    return FileLogObserver(
+        outFile,
+        lambda event: u"{0}{1}".format(_make_json(event), recordSeparator)
+    )
+
+
+def makeLegacyDailyLogFileObserver(path, logoutputlevel):
+    """
```
meejah: everything else in crossbar uses under_score style names; maybe more consistent to do so with this _logging module as well?

hawkowl: True, I'll fix that up.

crossbar/controller/processtypes.py
```diff
@@ -104,34 +107,23 @@ def log(self, childFD, data):
         """
         assert(childFD in self._log_fds)
 
-        if _PY3 and type(data) != str:
+        if type(data) != unicode:
             data = data.decode('utf8')
```
meejah: I think `if type(data) != six.text_type` or `if not isinstance(data, six.text_type)` for consistency with rest of crossbar

meejah:
To ditto my other comment about method names, I would change `camelCase` variable names to `under_score` style to match the rest of crossbar (especially `logPublisher` -> `log_publisher` since it's part of the API)

crossbar/controller/processtypes.py
hawkowl: Changed to use six instead

hawkowl:
@meejah I think I've got all of the non-Twisted public API that `_logging` exposes in underscore style, in the latest diff.

meejah:
cool, looks good. I didn't actually try running it, however ;)

hawkowl:
I'll make that a variable that's imported, and provide a link to why it is that specifically (there's a JSON log stread RFC for it).

hawkowl:
(I wish you could reply to comments in github :( )

meejah:
yeah, threaded comments would be nice...

PR #344, opened 2015-06-02
File upload resource for crossbar web service

markope:
# File Upload Resource

The uploader provides a means to upload files over the web using POST requests. This resource provides a backend for an upload request using HTML forms. The backend was inspired by the [resumable.js](http://www.resumablejs.com/) library which can be used as a fitting web front end for the task.

## Features

* Expects the File transfered in chunks. (size configured by client)
* Resumes uploads of a file even over brower or user sessions changes.
* Publishes all upload related events (including progress on file detail) over a configurable URI. This enables all clients and the backend to watch the progress when one client posts an upload.
* incomplete files are stored in a separate folder on the server to simplify the usage of inotify on the final upload folder. (configurable)
* provide a GET response to check wether a chunk of a file has already been uploaded, to skip uploading a file chunk.
* The file identifier is created by the client. This can be done simply by using the file_name or e.g. by using an md5 or similar together with the browsers FileAPI. Files with the same identifiers are treated as "the same file" by this resource.

Right now uploading files are shared accross web client sessions. That means, that if another session uploads the same file in the same moment, then this is detected and the upload is stopped. 

## Configuration
The uploader config is under the path property of the web transport. It needs a config object as follows:

```javascript
"upload": {
               "type": "fileupload",
               "form_fields": {
                  "file_id": "resumableIdentifier",
                  "file_name": "resumableFilename",
                  "mime_type": "resumableType",
                  "total_size": "resumableTotalSize",
                  "chunk_number": "resumableChunkNumber",
                  "chunk_size": "resumableChunkSize",
                  "total_chunks": "resumableTotalChunks",
                  "content": "file"
               },
               "directory": "../web_upload",
               "temp_directory": "../temp",
               "file_owner": {
                  "owner": "upload_owner",
                  "group": "somegroup",
                  "permissions": "0444"
               },
               "max_file_size": 209715200,
               "file_types": [".csv", ".txt"],
               "file_progress": {
                  "realm": "realm1",
                  "uri": "re.upload"
               }
            }
```

Every property up to temp_directory is mandatory.
The file_owner is optional. If you don't provide the owner then user by whom crossbar is started will be used.

### DIFF
--- a/crossbar/twisted/resource.py
+++ b/crossbar/twisted/resource.py
@@ -44,6 +44,8 @@ from autobahn.twisted import longpoll
 
 import crossbar
 
+from crossbar.adapter.rest.common import _CommonResource
+
 try:
     # triggers module level reactor import
     # https://twistedmatrix.com/trac/ticket/6849#comment:4
@@ -79,8 +81,7 @@ class JsonResource(Resource):
         return self._data
 
 
-class FileUploadResource(Resource):
-
+class FileUploadResource(Resource, _CommonResource):
 
     """
     Twisted Web resource that handles file uploads over HTTP post requests.
@@ -96,10 +97,8 @@ class FileUploadResource(Resource):
         self.fileProgressURI = file_progress_URI
 
     def render_POST(self, request):
-        # request.setHeader(b'content-type', b'text/text; charset=UTF-8')
         log.msg( 'file uploads --------POST--------')
         headers = request.getAllHeaders()
-        log.msg(json.dumps(headers))
 
         content = cgi.FieldStorage(
         fp = request.content,
@@ -120,37 +119,36 @@ class FileUploadResource(Resource):
         totalChunks = int(content['resumableTotalChunks'].value)
         fileContent = content['file'].value
 
-        print fileId,chunkNumber, chunkSize, cchunkSize, totalSize, fileType, filename, relPath, totalChunks
+        # print fileId,chunkNumber, chunkSize, cchunkSize, totalSize, fileType, filename, relPath, totalChunks
 
-        # check request header for file upload
         # check file size
-        print 'max file size', self.max_file_size, 'total', totalChunks
 
         if int(totalSize) > self.max_file_size: 
-            print 'what??'
-            request.setResponseCode(500)
+            request.setResponseCode(413)  #request entity too large
             return 'max filesize exceeded'
 
+        extension = os.path.splitext(filename)[1]
+        if extension not in self.fileTypes:
+            request.setResponseCode(400) 
+            return 'file extension not accepted'
 
-        # check mime type
-        # check file name suffix (file type)
+        # TODO: check mime type
 
-        # if first of many chunks: create temp file in temp_dir for uncomplete uploads.
         fileTempDir = os.path.join(self.tempDir,fileId)
         chunkName = os.path.join(fileTempDir, 'chunk_' + str(chunkNumber))
-        print fileTempDir
-        if not (os.path.exists(os.path.join(self.tempDir, fileId)) or os.path.exists(fileTempDir)):
+
+        if not (os.path.exists(os.path.join(self.dir, fileId)) or os.path.exists(fileTempDir)):
             # first chunk of file
             # publish file upload start to file_progress_URI     
             if totalChunks == 1: 
-                print 'doing first and only chunk'
+                # print 'doing first and only chunk'
                 # only on chunk overall
                 # write file directly
                 chunk = open(os.path.join(self.dir, fileId), 'wb') 
                 chunk.write(fileContent)
                 chunk.close
             else:
-                print 'doing first of more chunks'
+                # print 'doing first of more chunks'
                 # first of more chunks
                 os.makedirs(fileTempDir)
                 chunk = open(chunkName, 'wb') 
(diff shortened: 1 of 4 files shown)


markope:
Here is some demo.html code.
Besides this demo.html you need a config as in the pull request and you need to make sure that the upload directories from the config exist relative to the crossbar node folder.

```javascript
<!DOCTYPE html>
<html>

<head>
   <script src="bower_components/resumablejs/resumable.js"></script>
   <script src="bower_components/autobahn/autobahn.min.js"></script>
</head>

<body>
   <a href="#" id="browseButton">Select files</a>

   <script>
   var connection = new autobahn.Connection({
      url: 'ws://127.0.0.1:8089/ws',
      realm: 'realm1'
   });

   connection.onopen = function(session, details) {
      console.log("Connected", details);
      absession = session;

      absession.subscribe('re.upload', function(res) {
         absession.log('upload event received', res);
      });

      var r = new Resumable({
         target: 'upload',
         chunkSize: 10 * 1024 * 1024,
         simultaneousUploads: 4,
         testChunks: true
      });

      r.assignBrowse(document.getElementById('browseButton'));

      r.on('fileSuccess', function(file) {
         console.debug(r.files);
         // enable repeated upload since other user can delete the file on the server
         // and this user might want to reupload the file
         file.cancel();
      });
   };
   connection.open();
   </script>
</body>
</html>
```

oberstet:
@markope thanks!! this already looks very good. only 3 things:

* PEP8: flake8 fails (see Travis CI above): https://travis-ci.org/crossbario/crossbar/jobs/65073526 (you can test with `make flake8`)
* would be cool to have an example here https://github.com/crossbario/crossbarexamples - so I can test it and other have a jump start (to contribute: same as here, fork, do a PR)
* docs: https://github.com/crossbario/crossbarwww/blob/master/website/crossbario/docs/Web-Transports-and-Services.md and plus 1 new page similar to eg https://github.com/crossbario/crossbarwww/blob/master/website/crossbario/docs/HTTP-Bridge-Services-Publisher.md  (to contribute: same as here, fork, do a PR)

oberstet:
ah, and we need http://crossbar.io/docs/Contributing-to-the-project/#what-we-need-you-to-do

oberstet:
more of a note: changing owner/permissions of an uploaded file will only work when CB is run as `root` - which is strongly discouraged.

crossbar/worker/router.py
```diff
@@ -1055,6 +1056,57 @@ def create_resource(self, path_config):
...
+
+            file_progress_URI = ''
+            # If fileupload events are not desired the publish function does nothing then.
+            def fileupload_publish(payload):
+                return ''
+
+            if 'file_progress' in path_config:  
+                file_progress_URI = path_config['file_progress']['uri']
+
+                fileupload_session_config = ComponentConfig(realm=path_config['file_progress']['realm'], extra=None)
```
oberstet: This needs to be `realm=root_config['realm']`. You can't connect the service component to a different realm. Also: pls remove the respective config attribute ..

oberstet:
rgd `file_progress` publish URI being defined in the config: that means, all frontends subscribe to that topic and all frontends receive progress events for all uploaded files. Is that what you want?

I could imagine scenarios, where the progress event should only be received by the file uploading frontend instance ...

In general, it's important that we get the config "right". Since this is kind of public API of CB, and we want to have stability there ...

crossbar/worker/router.py
markope: ok with ownership, but I can change the permissions. I'll only keep the permission config then.
And yes, we should definitely get the config right.

markope:
Regarding the event publishing. What do you think of having one more optional form_field which holds a topic. Then the publish topic can be controlled by the client code on a granular level. Then I would remove the file_progress part from the config.

markope:
how should I get the `root_config['realm']`from within the `create_resource`function ? There I only have the `path_config`. I could hack, but I better ask you ;-)

markope:
I can't see the realm issue ?! At least it works with the configured realm like that. Also it seems that each realm has a service session and which one is picked as the "web transport" realm? Does that make sense?

markope:
except for the issue with the realm I'm ok now.

crossbar/twisted/resource.py
```diff
@@ -77,6 +79,244 @@ def render_GET(self, request):
...
+                # publish file upload progress to file_progress_URI
+
+                fileupload_publish({
+                                   "fileId": fileId,
+                                   "fileName": filename,
+                                   "totalSize": totalSize,
+                                   "status": "FINISH",
+                                   "progress": 1
+                                   })
+
```
meejah: Sorry for nit-picking, but I think it would be more-consistent to use the 4-space indenting (you've done this in some places for dicts). This also matches other indenting in crossbar, and IMO is more "normal" for Python. That is, like:

```
fileupload_publish({
    "key": value,
    "foo": bar
})
```

Also, not sure if it *really* matters for published information, but crossbar/autobahn use `under_score` naming, in general, not `camelCase` (Twisted uses camelCase, of course, so there's some of that in the code too :/).

But, for "internal" method calls like `removeTempDir` and variables (e.g. `fileTempDir`), I think `under_score` naming should be preferred to keep things consistent.

oberstet:
thx! note: I'm gonna adjust a couple of things now ..

PR #348, opened 2015-06-05
Fix the logger output on Windows

hawkowl:


oberstet:
thanks! looks better on windows now. no color though - should the change have fixed that too?

hawkowl:
That'll be separate -- need to move the colorama code to click. Now that I have a working Windows install, I'll be able to do that with certainty :)

tlausch:
hey amber, for windows you might try using colorama.init. after importing colorama in the crossbar/_loging.py:72
it doesnt look like the colors are correct on windows, but at least the output is colored and not scrambled with the color code representations ;-)

greetings.

PR #367, opened 2015-06-12
config.extra is supposed to be a dict

meejah:
Sometimes it is an `argparse.Namespace()` in crossbar usage, but Autobahn
docs say it should be a dict()

### DIFF
--- a/crossbar/common/process.py
+++ b/crossbar/common/process.py
@@ -127,10 +127,10 @@ class NativeProcessSession(ApplicationSession):
         """
         """
         if not hasattr(self, 'cbdir'):
-            self.cbdir = self.config.extra.cbdir
+            self.cbdir = self.config.extra['cbdir']
 
         if not hasattr(self, '_uri_prefix'):
-            self._uri_prefix = 'crossbar.node.{}'.format(self.config.extra.node)
+            self._uri_prefix = 'crossbar.node.{}'.format(self.config.extra['node'])
 
         self._started = datetime.utcnow()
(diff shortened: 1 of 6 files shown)


oberstet:
yes, we should cleanup / handle this systematically.

FWIW, the `extra` was originally thought as holding arbitrary user type (not only dict) - but I am fine with the PR.

But why does it work? An `argparse.Namespace()` allows access both by element and dict-like access?

meejah:
Ah, I see; well I could instead just change the documentation ;)

It only works along with the crossbar PR I also put in changing all access to ``self.config.extra`` to use it as a dict, and used ``vars(arguments)`` to conver the Namespace to a dict (as per argparse docs)...so if it's "supposed to work" with any user-supplied object, perhaps changing the docs and closing these two PRs is better.

oberstet:
Ah, I know where that dict use comes from: there is the option where you can have your component configured from `extra` coming from the node configuration. In that case, we only have a parsed JSON part, and no native Python object. But looking at `checkconfig.py`, `extra` can by anything JSON. A single number. An array. Or a dict. The latter often being the most useful practically.

When the component isn't started by a CB container worker (which will get the extra from node config), a user can provide native Python objects for extra as well.

I think we use this ability within CB itself: where we provide a Deferred in extra that is resolved in onJoin.

So: while there is a difference in what `extra` effectively can be, depending on how the component is started, I think making it "dict only" has some consequences.

What do you think?

PR #376, opened 2015-07-14
some fixes

hawkowl:
cc @oberstet 

fixes the container not starting up because checkconfig checks it with the wrong config checker
fixes non-serializable stuff going through the JSON serialiser
fixes and reenables some logger tests


oberstet:
could you remove the changes to crossbar/common/checkconfig.py from the PR? I merge then ..

PR #377, opened 2015-07-15
checkconfig improvements

hawkowl:
This does the following:

- Makes checkconfig raise its own exception for invalid config, which the cli then makes look nicer (no need to print a useless traceback). If checkconfig raises an actual unhandled exception, it'll still print the full stacktrace for us.
- Replace some prints with log.debugs
- Rename `check_components` `check_router_components`, and add `check_container_components`.
- Add some `checkconfig` regression tests, which fail before these changes.


oberstet:
awesome! no more cryptic stacktraces for invalid configs .. will be appreciated by users

PR #380, opened 2015-07-20
Test the CLI + make sure file logging actually works (especially on Py3) + make the CLI logger start on all functions

hawkowl:


hawkowl:
This changeset improves the py27 coverage from 27.21% to 32.36%, and sets the groundwork for future changes to the CLI to have tests to verify the changes.

PR #387, opened 2015-08-12
Better content type handling in the REST adapter

### LINKED ISSUE #364: REST bridge denying valid content-type
When a HTTP/POST has a content type `application/json; charset=utf-8`, the REST bridge will deny that, as it looks for `application/json` - and only that.

hawkowl:
Ought to fix #364


hawkowl:
The ideal case would be using something like https://pypi.python.org/pypi/negotiator/1.0.0 , but this should fix it for right now. (Using Negotiator will also let us do alternate format handling, like YAML or JSON or Cap'n Proto or whatever through the REST interface).

PR #391, opened 2015-08-17
Make the REST bridge handle encodings better

hawkowl:
This makes the REST bridge handle UTF-8 encoding better.


hawkowl:
cc @oberstet this is ready for review

oberstet:
- the only supported encoding should be UTF8 (not ascii, and not anything else)
- it does not use the AB UTF8 checker

hawkowl:
Okay, this uses just UTF-8 (with the assumption of that if it's not included), and uses the AB UTF8 checker to verify before loading it.

Issue #352, opened 2015-06-07
Remove use of twisted.python.log and twisted.python.compat


oberstet:
* `twisted.python.log` should be replaced with the new logging.
* `twisted.python.compat` should be replaced with `six`

oberstet:
@hawkowl: agree? want to do it?

hawkowl:
@oberstet Yep, I can take point on this.

oberstet:
@hawkowl should be fixed, right?

hawkowl:
Still a handful of `twisted.python.log` using modules:

```
crossbar//adapter/rest/subscriber.py:from twisted.python import log
crossbar//controller/guest.py:from twisted.python import log
crossbar//router/role.py:from twisted.python import log
crossbar//twisted/tlsctx.py:from twisted.python import log
crossbar//worker/container.py:from twisted.python import log
crossbar//worker/worker.py:from twisted.python import log
```
I can sort that out, though.

oberstet:
@hawkowl thx! I am trying to collect all open ends in the 0.11.0 milestone that relate to: logging, error handling, and cleanups like these. but I want to do a release soon ..

hawkowl:
#400 and #401 should close this up.

hawkowl:
I think that's it? grep returns the following things importing from `twisted.python`:

```
crossbar/_logging.py:from twisted.python.constants import NamedConstant
crossbar/_logging.py:from twisted.python.reflect import qual
crossbar/controller/cli.py:from twisted.python.reflect import qual
crossbar/twisted/resource.py:    from twisted.python.filepath import FilePath
crossbar/twisted/sharedport.py:from twisted.python.runtime import platformType
crossbar/worker/process.py:    from twisted.python.reflect import qual
```

No more logs, no more compat, no more tears :)

oberstet:
very good;) thanks!

PR #400, opened 2015-08-24
Remove the remainder of twisted.python.log

hawkowl:
Would fix one half of #352.


hawkowl:
cc @oberstet this is ready for review

PR #401, opened 2015-08-24
Remove the remainder of twisted.python.compat

hawkowl:
Would fix the other half of #352.


hawkowl:
cc @oberstet this is good for review now

Issue #328, opened 2015-05-16
SIGINT results in unclean exit


Paranaix:
This is fatal, as this effectively prevents one from restarting nodes (without restarting the router):

```
^C2015-05-16 16:18:42+0200 [Controller   7038] Received SIGINT, shutting down.
2015-05-16 16:18:42+0200 [Controller   7038] sending TERM to subprocess 7047
2015-05-16 16:18:42+0200 [Controller   7038] waiting for 7047 to exit...
2015-05-16 16:18:42+0200 [Container    7047] Ignoring SIGINT in worker.
2015-05-16 16:18:42+0200 [Container    7047] Received SIGTERM, shutting down.
2015-05-16 16:18:42+0200 [Container    7047] Connection to node controller lost.
2015-05-16 16:18:42+0200 [Container    7047] Unhandled Error
2015-05-16 16:18:42+0200 [Container    7047] Traceback (most recent call last):
2015-05-16 16:18:42+0200 [Container    7047] File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 295, in addCallbacks
2015-05-16 16:18:42+0200 [Container    7047] self._runCallbacks()
2015-05-16 16:18:42+0200 [Container    7047] File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 577, in _runCallbacks
2015-05-16 16:18:42+0200 [Container    7047] current.result = callback(current.result, *args, **kw)
2015-05-16 16:18:42+0200 [Container    7047] File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 429, in _continueFiring
2015-05-16 16:18:42+0200 [Container    7047] callable(*args, **kwargs)
2015-05-16 16:18:42+0200 [Container    7047] File "/usr/lib/python2.7/dist-packages/twisted/internet/base.py", line 624, in disconnectAll
2015-05-16 16:18:42+0200 [Container    7047] failure.Failure(main.CONNECTION_LOST))
2015-05-16 16:18:42+0200 [Container    7047] --- <exception caught here> ---
2015-05-16 16:18:42+0200 [Container    7047] File "/usr/lib/python2.7/dist-packages/twisted/python/log.py", line 88, in callWithLogger
2015-05-16 16:18:42+0200 [Container    7047] return callWithContext({"system": lp}, func, *args, **kw)
[...]

PR #406, opened 2015-08-25
shutdown + log improvements

meejah:
Relating to #328, a few minor logging improvements and a couple changes to make shutdown more robust: ignore TransportLost while trying to publish on_component_stop and stop checking if the reactor is running.

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -415,7 +415,8 @@ class Node(object):
                     # we've established a connection to the component
                     def component_exited(info):
                         dead_comp = info['id']
-                        self.log.info("Component '{}' failed to start; shutting down node.".format(dead_comp))
+                        self.log.info("Component '{dead_comp}' failed to start; shutting down node.", dead_comp=dead_comp)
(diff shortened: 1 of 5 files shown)


oberstet:
looks good. could you have a look at https://travis-ci.org/crossbario/crossbar/jobs/77244427#L305

crossbar/controller/node.py
```diff
@@ -415,9 +415,12 @@ def _startup(self, config):
                     # we've established a connection to the component
                     def component_exited(info):
                         dead_comp = info['id']
-                        self.log.info("Component '{}' failed to start; shutting down node.".format(dead_comp))
-                        if self._reactor.running:
+                        self.log.info("Component '{dead_comp}' failed to start; shutting down node.", dead_comp=dead_comp)
```
hawkowl: minor thing, you could pass in info and iirc use "{info.id}" here and "{info.config}" in the formatting string. I need to do this more often when I touch things...

meejah: Ah, yes, I keep forgetting about that syntax. I'll change it.
And the travis failures ;)

meejah:
@oberstet should be good for merge now, fixed a utest and hawlowl's suggestion

Issue #111, opened 2014-08-28
Better logging


oberstet:
We really need better logging. In particular related to application components and problems with importing, loading, connecting and running those.

E.g. start of a WAMP Python app component might fail because of

* PYTHONPATH issues
* syntax errors in the component
* failing imports in the component
* ...

hawkowl:
I'll definitely have to come up with some test cases for failing components and make sure (useful) logs are produced...

meejah:
The imports are easy: if you define a component of type ``class`` and just put a syntax error in the ``.py`` file you don't see the error in crossbar logs. I have *some* code for one of these from some debugging I did recently; I'll try with the most recent updates.

hawkowl:
#407 should fix up the final edge case -- it'll raise up the syntax error.

hawkowl:
This particular ticket, I think, can be closed -- if there's any further issues, I think they're best done as new tickets with a particular focus.

PR #407, opened 2015-08-26
Some misc things

hawkowl:


hawkowl:
cc @oberstet 
this will fix up #111 (unless there's any edge cases I'm unaware of, but it should catch them all)
it also improves coverage/the code of a bit of the CLI stuff -- not much to go until it's fully covered, a lot of it is low hanging fruit as far as coverage % goes, and it'll be quite rock solid at that point.

hawkowl:
This also brings the 'hello' template in line with the one from the examples (using the new logging system).

tox.ini
```diff
@@ -1,13 +1,16 @@
...
+ignore = E501
+exclude = crossbar/worker/test/examples/syntaxerror.py
 
 [testenv:flake8]
 skip_install = True
 deps =
    flake8
 commands =
-   flake8 --ignore=E501 crossbar
+   flake8 crossbar
```
oberstet: how can that work since we do have long lines in the codebase?

hawkowl: Flake8 looks at the [flake8] section in tox.ini automatically -- so we don't need to define it again. This means we can run flake8 manually and it'll pick up the settings.

hawkowl: (from http://flake8.readthedocs.org/en/latest/config.html#per-project )

oberstet:
@hawkowl ah, right. thanks for explaining!

PR #411, opened 2015-08-27
WAMP over alternate FD in container worker

hawkowl:


oberstet:
this PR seems to collect both general cleanups / logging improvements (which I'd like to merge if Travis when Travis is green) and FD stuff (which I don't want to merge now). could you separate that please? cool!

hawkowl:
@oberstet it's based on https://github.com/crossbario/crossbar/pull/407 , which should contain most of the cleanups

hawkowl:
@oberstet the FD stuff works, the test for it fails under PyPy. Oh, this'll be fun to debug...!!!

hawkowl:
Oh, it's because PyPy + Coverage adds 10s to startup time... :(

hawkowl:
Woo +5.72%!

oberstet:
I've merged #407 .. could you please rebase this PR?

hawkowl:
@oberstet rebased! *watches Travis*

hawkowl:
4.88%, now that the others are in. This changes it to FD3, and starts up a very basic Crossbar service to test this -- so we can make sure it works, and also make sure that starting up a service works at all!

@oberstet happy for comments at this point (end of the work day here :) )

oberstet:
One suggestion: not touch the standard FDs 0, 1 and 2 at all (other than logging what comes from 1 and 2), but use two dedicated FDs: 3 (controller->container) and 4 (container->controller) WAMP traffic. With the FDs hardcoded.

We have to test this on different platforms .. as I could imagine support for using non-standard FDs could differ between ..

hawkowl:
The problem with controller->container is that ProcessEndpoint.transport.write() automatically goes to FD0. This means we have to a) alter Autobahn to support using `transport.writeToChild()` for this one case (or rather, not making it always call transport.write(), but some other function, theoretically), b) wrap the transport so that write() goes to the FD we want, or c) continue using FD0. I tried to do b) but ran into some issues, I could probably get it to work if I take another stab at it with a fresh night of sleep :)

Up to you which one :)

bin/crossbar
```diff
@@ -3,7 +3,8 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.abspath(os.getcwd()))
+path = os.path.join(os.path.split(os.path.abspath(__file__))[0:-1])
+sys.path.insert(0, path[0] + "/../")
```
oberstet: um, ok;) I don't have to understand that, but it works, yes?

crossbar/controller/process.py
```diff
@@ -346,7 +347,7 @@ def _start_native_worker(self, wtype, id, options=None, details=None):
 
         # all native workers (routers and containers for now) start from the same script
         #
-        filename = pkg_resources.resource_filename('crossbar', 'worker/process.py')
+        filename = FilePath(crossbar.__file__).parent().child("worker").child("process.py").path
```
oberstet: curious: using Twisted `FilePath` is what I want?

crossbar/controller/process.py
```diff
@@ -405,9 +401,12 @@ def _start_native_worker(self, wtype, id, options=None, details=None):
 
         self._workers[id] = worker
 
-        # create a (custom) process endpoint
+        # create a (custom) process endpoint. We define a special FD3, which is
+        # used by the WAMP connection with the worker.
         #
```
oberstet: How about:

```
# The communication between controller and container workers is using WAMP running over 2 pipes.
# For controller->container traffic this runs over FD 0 (`stdin`) and for the
# container->controller traffic, this runs over FD 3.
# Note: We use FD 3, not FD 1 (`stdout`) or FD 2 (`stderr`) for container->controller traffic, so
# that components running in the container which happen to write to `stdout` or `stderr` do not
# interfere with the container-controller communication.
```

crossbar/controller/process.py
```diff
@@ -405,9 +401,12 @@ def _start_native_worker(self, wtype, id, options=None, details=None):
...
-        ep = WorkerProcessEndpoint(self._node._reactor, exe, args, env=worker_env, worker=worker)
+        ep = WorkerProcessEndpoint(
+            self._node._reactor, exe, args, env=worker_env, worker=worker,
+            childFDs={0: "w", 1: "r", 2: "r", 3: "r"})
```
oberstet: Problem is: not sure if `childFDs` are supported on Windows ..

crossbar/controller/processtypes.py
```diff
@@ -117,10 +117,21 @@ def _dump_remaining_log(self, result):
...
     def log(self, childFD, data):
         """
-        FIXME: line buffering
+        Handle a log message (or a fragment of such) coming in.
         """
         assert(childFD in self._log_fds)
 
+        system = "{:<10} {:>6}".format(self.LOGNAME, self.pid)
+
+        if childFD == 1:
```
oberstet: shouldn't we treat FD 2 the same?

crossbar/controller/test/test_cli.py
```diff
@@ -137,6 +144,131 @@ def test_start(self):
...
+               }
+            }
+         ]
+      }
+   ]
+}
+            """ % (code_location,))
+
+        with open(code_location + "/test.py", "w") as f:
+            f.write("""#!/usr/bin/env python
```
oberstet: convenient that we don't have windows testing;) I guess I am fine.

crossbar/controller/process.py
hawkowl: It's not. We'll just have to degrade back to the old behaviour on Windows, or not support it. I'm guessing you want the former?

crossbar/controller/process.py
hawkowl: Sorry, you'll have to rephrase. The `FilePath` usage here is because os.path has no `parent()` equiv, and using `__file__` rather than pkg_resources means that we use the one in the source tree that Crossbar is being ran from -- so if you've got it installed, it won't inadvertently use the installed copy (which is why I had to change it, it was using the one in system packages).

crossbar/controller/process.py
hawkowl: Good idea -- changed.

crossbar/controller/processtypes.py
hawkowl: FD2 is our special JSON stuff right now -- but maybe it should have its own pipe too, yeah.

PR #426, opened 2015-08-31
Add some semi integration tests for the Container

hawkowl:
Once it's finished it'll integrate https://github.com/crossbario/crossbarexamples/tree/master/containers/failures


hawkowl:
cc @oberstet this does 1-7 of the container ones, does not integrate the side-by-side ones yet.

crossbar/controller/test/test_run.py
```diff
@@ -0,0 +1,891 @@
...
+class ContainerRunningTests(CLITestBase):
+
+    def setUp(self):
+
+        CLITestBase.setUp(self)
+
+        # Set up the configuration directories
+        self.cbdir = os.path.abspath(self.mktemp())
+        os.mkdir(self.cbdir)
+        self.config = os.path.abspath(os.path.join(self.cbdir, "config.json"))
```
meejah: should get deleted in e.g. tearDown()

crossbar/controller/test/test_run.py
```diff
@@ -0,0 +1,891 @@
...
+            "Component 'component1' failed to start; shutting down node."
+        ]
+
+        self._start_run(config, myapp, expected_stdout, expected_stderr,
+                        _check)
+
+    def test_failure6(self):
+
+        config = """{
+   "controller": {
```
meejah: might make indenting/clarity/editing better to make "config" a Python dict, and do ``json.dump`` to write the configuration file?

meejah: also syntax highlighting :)

meejah: maybe also there could be a "common" config object at the top, or in CLIBaseTest and each test could just add its own "bits that are different"? (it's just the "component" section that's different I think?)

oberstet:
looks good to me!

crossbar/controller/test/test_run.py
hawkowl: The reason for the duplication is so that each test case is independent -- they all match right now, but we might add some more test cases later :)

`json.dumps` is also a good idea. I'll fix that up.

PR #428, opened 2015-09-02
Only always add pythonpath on native workers

hawkowl:
cc @meejah


meejah:
For the record and the search-bots, this would cause exceptions like this in Python3 guests:

```
Fatal Python error: Py_Initialize: Unable to get the locale encoding
File "/tmp/cts_Qyeb1c/lib/python2.7/encodings/__init__.py", line 123
raise CodecRegistryError,\
^
SyntaxError: invalid syntax
```
```

PR #432, opened 2015-09-03
don't publish meta events to lost sessions (fixes #431)

### LINKED ISSUE #431: Don't publish meta events to lost sessions
migrated from: https://github.com/tavendo/AutobahnPython/issues/459

oberstet:


oberstet:
alright. I am sure this fixes the issue, and pretty sure it doesn't introduce a new one, but I'd like to improve on my unit test skills;)

@hawkowl @meejah how would I write a unit test for this?

the critical line to provoke the original issue is this: https://github.com/crossbario/crossbar/pull/432/files#diff-8945283a6bce56d3d36c1032b21bcd78R264

if you take out the test for `receiver._session_id` not being `None`, it'll fail ..

oberstet:
here is the fun part: the PR consists of 3 commits. all our tests succeeded at every stage. but without the subsequent 2 commits, we would have had 2 new, hard to track down issues.

I've done quite some testing stuff at the functional/integration/fuzzing area. much less at unit test level.

and I am wondering what's to right approach to prohibit issues like above (resulting from missing those 2nd and 3rd commits in the PR) ..

or: I am trying to convince myself more of the value of unit tests ..

oberstet:
another area I'd like to improve my understanding is coverage.

looking at the relevant lines (before PR):

https://codecov.io/github/crossbario/crossbar/crossbar/router/session.py?ref=93829480571cea3812563a9f470ff0c196e8c490#l-367

https://codecov.io/github/crossbario/crossbar/crossbar/router/broker.py?ref=93829480571cea3812563a9f470ff0c196e8c490#l-264

I am wondering: say we would have had *some* unit test which would make above lines green and covered.

however, *some* unit test that leads to coverage would have been no guarantee that it would have caught the original issue, nor the bugs I nearly introduced if I hadn't done those 2 follow up commits, right?

meejah:
I can confirm this works against my test-case also.

I will take a crack at a unit-test, though, and see what it might look like ;)

oberstet:
@meejah awesome, thanks!

meejah:
Regarding coverage: no, of course coverage is no guarantee that you covered *all* the cases, or the *right* cases. However, you can turn on "branch coverage" to get yellow lines on conditionals where you don't cover all the branches.

For a language like Python, "100% line coverage" is ensures you have no syntax errors -- but of course doesn't ensure you've "got all the bugs". That comes from: find bug; write failing test; fix test...

crossbar/router/broker.py
```diff
@@ -261,7 +261,7 @@ def on_authorize_success(authorized):
                             for receiver in receivers:
                                 if me_also or receiver != session:
                                     # the receiving subscriber session might have been lost in the meantime ..
-                                    if receiver._transport:
+                                    if receiver._session_id and receiver._transport:
                                         receiver._transport.send(msg)
```
meejah: Would it be better here to use an existing interface? E.g. is it possible to say "all observers implement ITransport" or similar, and call ``.isOpen``.

oberstet: I agree, would be cleaner. firing up pycharm and the code checking there spills out tons of private var access =( at some point, I'd like to clean that up ..

meejah:
Regarding unit-tests, see #434 for some working-but-impure ones. Plus some bonus Broker tests I found lying around.

PR #449, opened 2015-09-08
various py2/3 unicode/bytes issues fixed

oberstet:


crossbar/adapter/rest/subscriber.py
```diff
@@ -59,21 +59,29 @@ def onJoin(self, details):
...
         debug = self.config.extra.get("debug", False)
-        method = self.config.extra.get("method", u"POST").encode("utf8")
+        method = self.config.extra.get("method", u"POST")
         expectedCode = self.config.extra.get("expectedcode")
 
         @inlineCallbacks
         def on_event(url, *args, **kwargs):
 
             headers = Headers({
+                # FIXME: are keys/values supposed to be byte strings?
```
hawkowl: Yes, they should be. (This file has not been ported because of the Treq requirement -- see https://github.com/twisted/treq/pull/110 for blockers on that)

crossbar/controller/cli.py
```diff
@@ -487,7 +487,14 @@ def run_command_start(options, reactor=None):
                 'options': {x: y for x, y in options_dump.items()
                             if x not in ["func", "argv"]}
             }
-            fd.write("{}\n".format(json.dumps(pid_data, sort_keys=False, indent=3, separators=(',', ': '))))
+            fd.write("{}\n".format(
```
hawkowl: This should have a `.encode('utf8')` on it (since we're opening it in binary mode).

hawkowl: (oh and an `ensure_ascii=False`)

hawkowl:
Added a few comments - they should fix the Python 3 failures. Then I think it's good to merge :)

crossbar/adapter/rest/subscriber.py
oberstet: Not sure I get it: so it should read like the following already _now_? For this PR?

```python
headers = Headers({
    u"Content-Type": [u"application/json"]
})
```

hawkowl: no, it should be `b"Content-Type": [b"application/json"]`. Headers are bytestrings (for various reasons...).

oberstet: Ok. thx

hawkowl:
LGTM :+1:

PR #450, opened 2015-09-09
type of .signaure should be unicode

meejah:
Elsewhere we always demand that the signature is unicode.

### DIFF
--- a/crossbar/router/auth.py
+++ b/crossbar/router/auth.py
@@ -101,8 +101,9 @@ class PendingAuthWampCra(PendingAuth):
         }
 
         # challenge must be bytes
-        self.challenge = json.dumps(challenge_obj, ensure_ascii=False).encode('utf8')
-        self.signature = auth.compute_wcs(secret, self.challenge)
+        self.challenge = json.dumps(challenge_obj).encode('utf-8')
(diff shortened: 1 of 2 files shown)


oberstet:
@meejah I believe this might have overlapped with https://github.com/crossbario/crossbar/pull/449 .. can't work on it right now, but asked @hawkowl to take a look ..

hawkowl:
I am not sure that it should be Unicode -- a signature is a signed hash (or similar), which is a bunch of encoded bytes. I think `onChallenge` might need to be updated instead? (although it's a WAMP thing so is always unicode anyway? hmm).

meejah:
Yeah, intuitively it seemed to me it should also be ``bytes`` (which I why I assert()-ed that in the original PR). It's entirely possible to instead insist that the signature should be ``bytes``, but that's more changes (and e.g. would change what user-code is supposed to return from ``onChallenge``)...@oberstet?

In either case, I can re-base this and re-push (or re-do it as "signatures are bytes too").

oberstet:
@meejah can we close this PR (as other stuff has been merged already)?

meejah:
yes, should be closed I believe.

PR #459, opened 2015-09-17
Don't repr() all guests

hawkowl:


meejah:
works for my test-case

PR #467, opened 2015-09-21
No Python 3.x on Windows and more tests

hawkowl:


hawkowl:
cc @oberstet @meejah  -- this fails on `-abcurrent` on Python 2.7, but passes on `-abtrunk`. It'll only pass when 0.11 is released, with all the Unicode fixes.

Issue #490, opened 2015-10-24
Guest watcher starts new guest before old guest shuts down


Kostronor:
I am using a java guest application with crossbar.io 0.11.1:

```
Automatically choosing optimal Twisted reactor
Running on Linux and optimal reactor (epoll) was installed.
     __  __  __  __  __  __      __     __
    /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/

 Crossbar.io        : 0.11.1
   Autobahn         : 0.10.9
     UTF8 Validator : autobahn
     XOR Masker     : autobahn
     JSON Codec     : stdlib
     MsgPack Codec  : msgpack-python-0.4.6
   Twisted          : 15.4.0-EPollReactor
   Python           : 2.7.10/PyPy-2.6.1
 OS                 : Linux-3.13.0-042stab111.12-x86_64-with-debian-jessie-sid
 Machine            : x86_64
```

My guest worker is similar to the hello template and the issue can be reproduced with it. My config.json:

```
{
      "type": "guest",
      "executable": "java",
      "arguments": [
        "-Djava.util.logging.config.file=./.crossbar/logging.properties",
        "-jar",
        "./bin/hello.jar",
        "ws://127.0.0.1:7777/ws",
        "realm"
      ],
      "options": {
        "workdir": "..",
        "watch": {
          "directories": [
            "../bin",
            "../sync/client"
          ],
          "action": "restart"
        },
        "stdout": "log",
        "stderr": "log"
      }
    }
```

Because the file watcher only gets notified when there is a new file, I have added this code in my build script to trigger a restart:

```
#all moved, pinging restart now
##first the client
touch ./sync/client/RESTART
##cleanup
rm ./sync/client/RESTART
```

My log looks like this:

```
2015-10-24T09:11:46+0200 [Controller  22560] Using guest worker executable '/usr/bin/java' (executable path detected from environment)
2015-10-24T09:11:46+0200 [Controller  22560] Starting Guest with ID 'worker2'...
2015-10-24T09:11:46+0200 [Controller  22560] Guest with ID 'worker2' and PID 22570 started
2015-10-24T09:11:46+0200 [Controller  22560] Guest 'worker2': started
2015-10-24T09:11:46+0200 [Guest       22570] INFORMATION: Session status changed to Disconnected
2015-10-24T09:11:46+0200 [Guest       22570] INFORMATION: Session status changed to Connecting
[...]

PR #491, opened 2015-10-24
Fixed watcher restart problems

Kostronor:
This is a quick hacky fix for issue #490

### DIFF
--- a/crossbar/controller/process.py
+++ b/crossbar/controller/process.py
@@ -767,7 +767,10 @@ class NodeControllerSession(NativeProcessSession):
 
                         if options['watch'].get('action', None) == 'restart':
                             self.log.info("Restarting guest ..")
-                            self._node._reactor.callLater(0.1, self.start_guest, id, config, details)
+                            # Add a timeout large enough (perhaps add a config option later)
+                            self._node._reactor.callLater(10.1, self.start_guest, id, config, details)
+                            # Shut the worker down, after the restart event is scheduled
+                            worker.stop()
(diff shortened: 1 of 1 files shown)


meejah:
Hi, thanks for the patch! I can't find the list of contributers, but if you've not signed something, you'll have to look at this: http://crossbar.io/docs/Contributing-to-the-project/

Otherwise, yes a timeout isn't ideal but also better than "doesn't work" ;)

meejah:
*Ideally* I think what we'd do is send the SIGTERM (like now), wait a configurable number of seconds for the process to die and if it's still not dead, SIGKILL it. Then do the restsart. 

``_cleanup_worker`` actually does something similar to this as well, so it  might be worth factoring out code to do this (i.e. given a ``ProcessProtocol``, SIGTERM it and wait until it dies, or give up and SIGKILL it after X second and fire the returned Deferred after the process is finally dead/killed).

Kostronor:
Hey, thanks for the heads up, will sign the CAA today (if my scanner works ;-) ) 

If I remember correctly, cleanup worker also checks every second if the process is dead now, this is highly preferred because we can have large timeouts without using the full timeout every time.

meejah:
Yes, ``_cleanup_worker`` checks every second.
A non-polling way of accomplishing this would be even better, of course, but this works fairly well.

oberstet:
@Kostronor CAA received - thanks! ah, and thanks for the actual patch;)

PR #501, opened 2015-11-01
Added caller_role and caller_id to rpc details

Kostronor:
Depends on crossbario/autobahn-python#540


meejah:
You can use ``wamp.session.get`` with the session-id given to you via ``caller`` to get these details already.

Since this changes the API a bit (extends the Details dict) I'm adding 'needs-discussion' label (e.g. why not add all the session information?)

Kostronor:
I think having to call `wamp.session.get` and join the call so that the original rpc waits for `wamp.session.get` to return feels like a hack, sending the information (if exists) makes sense (like already done by authorize and authenticate rpc calls, where there are loads of information passed to the call that "could" be helpful. 

If there are more details relevant to the callee then they should also be included, I would say.

goeddea:
Just thinking out loud:

I see the motivation for getting the information that is required in a particular application without additional round-trips to the router. On the other hand we want the wire traffic to be minimal.

Since there is no transmission of any information about the caller without explicit configuration (set by the caller now), it could be argued that we only increase the wire traffic in cases where this is necessary anyway, and why not give all relevant information in this case?

On the other hand, where does relevant information end? Do we include transport information? Any other information which the router may have about the session/caller after future extensions? 

The everything-to-everybody solution would, of course, be to have configurable details: When a callee registers a procedure, it tells the router which details it requires, and these are then sent with every call. Not sure about the implications within Crossbar/WAMP routers in general however.

Kostronor:
Some thoughts, maybe not all on topic:

Disclosing should be something managed by the callee. 

When registering an rpc function, you should be able to specify your need for data. I would propose three possible options:

`{IGNORED,PREFERRED,REQUIRED}`

For simplicity i will use `add2` as the function. Now when we register, we can specify our need for some information, `IGNORED` is default if we do not specify anything.

lets say (in pseudo-code) we want to register a procedure:
```
register("add",callback,{
"details":{
    "caller_session": PREFERRED,
    "caller_id": REQUIRED,
    "transport": IGNORED
}
});
```

In this example, we require the `caller_id` for our function to work. We also would like to have the `caller_session`, but we can work without it. We do not need any `transport` information and explicitly ignore it. Also `caller_role` is implicitly ignored. This way each function can say that it needs some information and adding new information does not break anything.

But what if a caller does not want to be disclosed?

Well, if the callee said, some information is preferred, then he simply does not get it. He said, his procedure works without it and thats what happens.

If the callee requires some information, then there are two options for the caller: Give the information or no call. A simple model for this would be the existing `disclose_me` flag, but we change it to default true. If false, all calls to functions requiring some information that the caller does not want to disclose will fail. If true, only when information is required, then it will be send.

For the case that a function will ignore some information, we simply do not send it, saving bandwidth.

Extra addition:

Also we could give app developers the option to add new details that wll be filled by guest workers. Perhaps something like custom session variables. But that would be something else to discuss some time else...

meejah:
I would add: what are you trying to do that you need ``auth_role`` sent per-RPC? This seems like the authorization APIs aren't helping you achieve what you want...?

meejah:
@Kostronor can you describe the use-case that led to you wanting this feature? And specifically: why ``authrole`` and ``authid``?

Kostronor:
@meejah The more I think about it, there are no "good" reasons for `auth_role`... But perhaps there are some cases where it is needed. Need to think about that a bit more, but it would not be necessary for my use case.

My use case: An online game

You have many clients and one server (perhaps broken into multiple instances doing different things but transparent for each user)

The server (at the moment) authenticates itself as `backend` and is a guest worker on crossbar.io

Clients authenticate as their username as `auth_id`, so I can use the authentication feature crossbar.io provides and am not "rolling my own login/crypto". I have custom authenticate and authorize procedures registered for the backend, so that I can send user targeted events without making them public, like `event.user.kostronor.message.receive` since there is no option to just notify one user (or call the rpc on specifically this user).

So I have to have clients register accounts, so there exists a default auth role `guest` that can only call `register` and create an account that way. 

But once logged in, the backend needs to know the id of a user and that on every call that the user makes. Having every single rpc call make a round trip seems like a whole other issue.

Ultimately I would love to have my custom authentication procedure assign each session it's user id (as in Integer) and then use this more efficient key to query my data, but having the username and querying the id from that seems fine enough.

At the moment I would have to send an error code every time a caller does not disclose himself. I think I have chosen `INVALID_PARAMETERS` but none would really fit...

You sometimes need a single always-on instance of trust while keeping the rest in a "separation of concerns" model. And reusing the authorization APIs is a major argument for or against using crossbar.io

(In my case users can play games p2p where the server only records and validates them, wamp is perfect here)

meejah:
I wouldn't suggest actually calling ``wamp.session.get`` on each call, but instead you could map ``session_id`` -> relevant application information. If part of what you need to know per-session is the user-id, then you'd only need to grab it once? I guess all I'm suggesting is: instead of looking up things via user-id, you look them up via session-id. That does take some book-keeping locally, but presumably you need to keep some state anyway (so you *would* need to add "session 1234 is user X" to that, via e.g. listening to ``wamp.session.on_join``).

Another way to organize this is via the RPC namespace: you could, for example, put the user-id's into your URIs, like ``com.example.user.<user_id>.challenge_to_game`` or similar -- then the RPC call knows which user called it from the URI. You may already do some of this, based on the "targeted events" comment?

BTW, you *can* give an explicit list of session-IDs which shall receive a given publish, via ``eligible=`` to ``autobahn.wamp.types.PublishOptions``.

Kostronor:
Edit:
It seems like listening to `wamp.session.on_join` is the way to go then, will try that out first. I still think it would be good to have an optional `auth_id` detail, but with some bookkeeping it should work the same.

Old message:
The problem with `session_id` is that every page reload generates a new `session_id`. Also I can never make a connection between user and id, as authentication calls are for "someone WANTS to log in, give me his credentials" and not "someone has successfully logged in with session xxx"

meejah:
``session_id`` is a nice way to tie things together: it doesn't assume anything about your app (e.g. maybe it's okay to have many connections from one user, like one from their phone, one from their desktop, etc) and maps one-to-one to core WAMP concepts.

As @goeddea hints at, different apps are interested in different session-details; so ``session_id`` is definitely a "lowest common denominator" in a way but also is common through the APIs (e.g. the ``eligible=`` and ``exclude=`` options take lists of session-IDs).

oberstet:
until the WAMP spec is finalized rgd this aspec, we are not going not change this

PR #503, opened 2015-11-02
use efficient pattern matcher for wildcard subscriptions

yoch:
The current implementation for pattern-based subscriptions is very inefficient.

Here a pull request with a trie-based pattern matcher, tailored for crossbar.io usage. Should be very fast, even with many subscriptions.


meejah:
Thanks, @yoch!

I won't have time to look closely for a couple days but if we were to merge this, you'd first need to sign a contributer agreement for crossbar: https://github.com/crossbario/crossbar/blob/master/legal/individual_caa.pdf

See here for more: http://crossbar.io/docs/Contributing-to-the-project/

oberstet:
Thanks, but this already works using a Trie data structure ...

yoch:
@oberstet : Only prefix subscriptions are managed by StringTrie, but wildcard subscriptions not.

oberstet:
Correct. And wildcard matching is optimized for the 99% case where you have 1 to a few wildcards only. Do you have performance numbers comparing the existing code with the PR?

yoch:
I plan to make a benchmark. But trie-based approach performs obviously better with great number of subscriptions.

The real question is how such patterns to consider.

yoch:
Here benchmark results : 

```text
test on adding 76 wildcards, and searching 100 corresponding uris
	 Old : add	 0.0 s
	 Old : search	 0.015599966049194336 s
	 Old : remove	 0.0 s
	 ----------------------------------------
	 New : add	 0.0 s
	 New : search	 0.0 s
	 New : remove	 0.0 s
	 ----------------------------------------
test on adding 707 wildcards, and searching 1000 corresponding uris
	 Old : add	 0.0 s
	 Old : search	 0.062399864196777344 s
	 Old : remove	 0.0 s
	 ----------------------------------------
	 New : add	 0.015599966049194336 s
	 New : search	 0.015599966049194336 s
	 New : remove	 0.0 s
	 ----------------------------------------
test on adding 3550 wildcards, and searching 5000 corresponding uris
	 Old : add	 0.031200170516967773 s
	 Old : search	 0.2858009338378906 s
	 Old : remove	 0.015599966049194336 s
	 ----------------------------------------
	 New : add	 0.04679989814758301 s
	 New : search	 0.07800102233886719 s
	 New : remove	 0.04679989814758301 s
	 ----------------------------------------
test on adding 7017 wildcards, and searching 10000 corresponding uris
	 Old : add	 0.046801090240478516 s
	 Old : search	 0.6084010601043701 s
	 Old : remove	 0.015599966049194336 s
	 ----------------------------------------
	 New : add	 0.10920000076293945 s
	 New : search	 0.15699982643127441 s
	 New : remove	 0.06240081787109375 s
	 ----------------------------------------
test on adding 14175 wildcards, and searching 20000 corresponding uris
	 Old : add	 0.12479996681213379 s
	 Old : search	 1.2022018432617188 s
	 Old : remove	 0.031199932098388672 s
	 ----------------------------------------
	 New : add	 0.23400115966796875 s
	 New : search	 0.31200098991394043 s
	 New : remove	 0.12479996681213379 s
	 ----------------------------------------
test on adding 35115 wildcards, and searching 50000 corresponding uris
	 Old : add	 0.37540102005004883 s
	 Old : search	 2.9514050483703613 s
	 Old : remove	 0.09360098838806152 s
	 ----------------------------------------
	 New : add	 0.6084010601043701 s
	 New : search	 0.8424010276794434 s
	 New : remove	 0.3432011604309082 s
	 ----------------------------------------
```

http://pastebin.com/dFs5j8Qg

oberstet:
@yoch alright, Now we are talking;) Thanks backing up with numbers. So there could be something to gain. add/remove gets slower, while search gets faster.

From looking at your test code: this suggests that we should be able to have 2 classes in `pmatcher.py`, one for the old and one for the new matcher. This would then allow to switch between implementations easily and also create unit tests that demonstrate the behavior of the old/new are identical. Maybe we should also rename the file to `wildcard.py` (with 2 exported classes `WildcardTrieMatcher` and `WildcardMatcher` - the native existing one).

If you could come up with a PR that retains both the old/new and let's us switch by changing a single like, I review and merge. Ah, and we need:

* All _current_ Travis tests pass / green. But you already did that. Thanks!
* Please send us a CAA (we only need this once .. pls see @meejah comment)
* Any new files must get the copyright header exactly as in all the other files

Bonus: add new unit tests that demonstrate a) wildcard matching works as expected and/or b) the behavior of the old and new wildcard matcher is identical. That is, unit test that only work on `wildcard.py` (there are a bunch on `observation.py` already).

Thanks for your work!

yoch:
@oberstet 
Here benchmark modified according to your proposal : http://pastebin.com/1yTq6pDC

It's also possible to speedup `WilcardTrieMatcher.__contains__()` by adding keys into a `set` (at memory cost).

Note also that the longer are the uris, more significant is the difference between the two approaches. In other words, WilcardTrieMatcher can allow more expressive patterns than wamp currently supports.

*I plan to clean up the code and add tests before submitting a PR.*

yoch:
CAA sent

goeddea:
received the CAA and added you to the contributors list in the repo

oberstet:
@yoch awesome, thanks!

PR #498, opened 2015-11-01
Store challenge JSON in a serializable format

jvdm:
Storing `challenge` in `PendingAuthWampCra` as a UTF-8 JSON instead of
bytes is more natural; json spec states that json strings are utf-8,
so it usually will be required to be used throughout as such. For
example, the field is currently only required to be bytes during
signature calculation, which is done in its initialization.

This would fix #489.

Issue #489, opened 2015-10-21
WAMPCRA fails on python3 because challenge is not json serializable


jvdm:
I'm exercising WAMPCRA using crossbar-0.11.1 on python3.

During `HELLO` handling, crossbar is throwing an exception in a
Deferred, the client is never getting the `CHALLENGE` response.

    $ crossbar start
    2015-10-21T19:09:32-0200 [Controller   3442] Automatically choosing optimal Twisted reactor
    2015-10-21T19:09:32-0200 [Controller   3442] Running on Linux and optimal reactor (epoll) was installed.
    2015-10-21T19:09:32-0200 [Controller   3442]      __  __  __  __  __  __      __     __
    2015-10-21T19:09:32-0200 [Controller   3442]     /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    2015-10-21T19:09:32-0200 [Controller   3442]     \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/
    2015-10-21T19:09:32-0200 [Controller   3442]
    2015-10-21T19:09:32-0200 [Controller   3442]     Version: 0.11.1
    2015-10-21T19:09:32-0200 [Controller   3442]
    2015-10-21T19:09:32-0200 [Controller   3442] Starting from node directory /home/jvdm/.crossbar
    2015-10-21T19:09:32-0200 [Controller   3442] Loading node configuration file '/home/jvdm/.crossbar/config.yaml'
    2015-10-21T19:09:33-0200 [Controller   3442] Entering reactor event loop...
    2015-10-21T19:09:33-0200 [Controller   3442] Warning, could not set process title (setproctitle not installed)
    2015-10-21T19:09:33-0200 [Controller   3442] Process utilities not available
    2015-10-21T19:09:33-0200 [Controller   3442] Joined realm 'crossbar' on node management router
    2015-10-21T19:09:33-0200 [Controller   3442] No WAMPlets detected in enviroment.
    2015-10-21T19:09:33-0200 [Controller   3442] Starting Router with ID 'worker1'...
    2015-10-21T19:09:33-0200 [Router       3446] Worker running under CPython-EPollReactor
    2015-10-21T19:09:33-0200 [Router       3446] Process utilities not available
    2015-10-21T19:09:33-0200 [Controller   3442] Router with ID 'worker1' and PID 3446 started
    2015-10-21T19:09:33-0200 [Controller   3442] Router 'worker1': realm 'realm1' (named realm1) started
[...]

PR #510, opened 2015-11-07
WAMP-CRA auth

meejah:
For #489. Although previous discussion concluded that the challenge should be a ``bytes``, that can't work very well, as the json module won't serialize a ``bytes``. So, this switches both the signature and challenge to be a UTF8 string.

This works with the client-side code given in the current crossbar.io docs.

This is probably incomplete; it's failing on some log unittest for some reason in particular tox environments and I want to see if it runs on travis or not.

### DIFF
--- a/crossbar/router/auth.py
+++ b/crossbar/router/auth.py
@@ -100,9 +100,8 @@ class PendingAuthWampCra(PendingAuth):
             'timestamp': util.utcnow()
         }
 
-        # challenge must be bytes
-        self.challenge = json.dumps(challenge_obj, ensure_ascii=False).encode('utf8')
-        self.signature = auth.compute_wcs(secret, self.challenge)
+        self.challenge = json.dumps(challenge_obj, ensure_ascii=False)
+        self.signature = auth.compute_wcs(secret, self.challenge.encode('utf8')).decode('ascii')
 
 
 class PendingAuthTicket(PendingAuth):
--- a/crossbar/router/session.py
+++ b/crossbar/router/session.py
@@ -461,7 +461,7 @@ class _RouterSession(BaseSession):
 
         DO NOT attach to Deferreds that are returned to calling code.
         """
-        self.log.failure("Internal error (2): {log_failure.value}", failure=fail)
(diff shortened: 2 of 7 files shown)


oberstet:
@meejah have you tested this with AutobahnJS, e.g. the examples from https://github.com/crossbario/crossbarexamples/tree/master/authenticate ?

meejah:
@oberstet I have not, I will do that and report back.

meejah:
I also need to figure out why some of my local Tox environments fail the tests.

meejah:
Both the 'wampcra' and 'wampcradynamic' examples work out-of-the box on this branch (running under either Python2 or 3.5).

meejah:
My local tox failures are some logging failures when using crossbar master vs. v0.10.9 (between then and master, we moved logging to txaio). I think the only implication of this is that we need a new Autobahn release before we can release crossbar master (which is true with or without the changes in this pull-request AFAICT).

So, +1 from me to merge this and close #498 unmerged thus fixing #489

oberstet:
running locally at least, I get a failing test:

```
[FAIL]
  File "/home/oberstet/scm/crossbario/crossbar/crossbar/router/test/test_authorize.py", line 93, in test_wamp_cra_challenge
    self.assertTrue(isinstance(pend.challenge, six.text_type))
  File "/home/oberstet/pypy400_1/site-packages/twisted/trial/_synctest.py", line 400, in assertTrue
    super(_Assertions, self).assertTrue(condition, msg)
  File "/home/oberstet/pypy-4.0.0-linux64/lib-python/2.7/unittest/case.py", line 424, in assertTrue
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: False is not true

crossbar.router.test.test_authorize.TestPendingAuth.test_wamp_cra_challenge
-------------------------------------------------------------------------------
Ran 141 tests in 1.062s

FAILED (failures=1, successes=140)
make: *** [test] Fehler 1
```

meejah:
I can repeat that with PyPy 2.6.1.
Not sure if it's correct but relaxing the check in the test to ``(six.text_type, str)`` fixes this...

Issue #505, opened 2015-11-03
File Upload Service Python 3 - allow for extra data/dynamic dir placement


singlecheeze:
The resumable.js spec states that extra data may be appended to the query config option:

> query Extra parameters to include in the multipart POST with data. This can be an object or a function. If a function, it will be passed a ResumableFile object (Default: {})

Crossbar already uses this config parameter for json data that hooks up to `on_progress` and `session` as such:

``` javascript
r = new Resumable({
            target: 'upload',
            chunkSize: 1024 * 512, // Deault 1MB 1024 * 1024
            forceChunkSize: true, // https://github.com/23/resumable.js/issues/51
            simultaneousUploads: 4,
            testChunks: true,
            query: {
                on_progress: 'mono.upload.progress',
                session: sess.id
            }
        });
```

What I am proposing is allowing the user some functionality to expand on this further. What I would like to use it for is dynamically assigning a destination directory for the final upload that is a child of the crossbar configuration directory `"directory": "../uploaded"`

So in use a user could do this:

``` javascript
            query: {
                on_progress: 'mono.upload.progress',
                session: sess.id,
                child_dir: 'this_is_a_child_dir'
            }
```

And the result on successful file upload in crossbar would be before the final move occurs:

``` python
if query["child_dir"]:
     """ Maybe check to see if the target child dir is in fact a child 
          if full path of child is already given in the option, 
          this might be a way of adding nesting at n levels as well """
     destination = os.path.join("<<directory from directory config option>>", "<<child_dir>>")
```

Without this option, all files uploaded get put in exactly one directory on the server without the ability to add sub dirs even at one level deep since crossbar is a "Black Box" to a client application.

This is only an example of an implementation, if someone has something better or a way of already accomplishing this (I couldn't find one from the docs) then please share.

The only way to get around this currently is to wait for a `r.on('fileSuccess', function (file, message)` on the client and then make the client responsible to call a method to move the file to it's final destination. Having another mechanism on the server listen to `on_progress` and do something then will not solve this as there is still no way to pass additional data to that method on the server side.

markope:
hm, making subdirs one level deep sounds like a very specific requirement. I feel a solution to this should be more generic in some way. Could you tell me a bit about the actual use case you are after?

singlecheeze:
Use case is just to have the upload data contain a target directory if
specified in the resumable object and to place the reassembled resumable
pieces there on successful upload vs a singular target directory for all
uploads.

[...]

Issue #504, opened 2015-11-03
File Upload Service Python 3 - twisted makedirs


singlecheeze:
@goeddea @hawkowl please ref #481 I think this might be related, same router config and index/script used there.

```text
2015-11-02T20:41:53-0500 [Controller   8066] Entering reactor event loop...
2015-11-02T20:41:53-0500 [Controller   8066] Warning, could not set process title (setproctitle not installed)
2015-11-02T20:41:53-0500 [Controller   8066] Process utilities not available
2015-11-02T20:41:53-0500 [Controller   8066] Joined realm 'crossbar' on node management router
2015-11-02T20:41:53-0500 [Controller   8066] No WAMPlets detected in enviroment.
2015-11-02T20:41:53-0500 [Controller   8066] Starting Router with ID 'wamprouter'...
2015-11-02T20:41:54-0500 [Router       8070] Worker running under CPython-EPollReactor
2015-11-02T20:41:54-0500 [Router       8070] Process utilities not available
2015-11-02T20:41:54-0500 [Controller   8066] Router with ID 'wamprouter' and PID 8070 started
2015-11-02T20:41:54-0500 [Controller   8066] Router 'wamprouter': realm 'realm1' (named 'realm1') started
2015-11-02T20:41:54-0500 [Controller   8066] Router 'wamprouter': role 'role1' (named 'anonymous') started on realm 'realm1'
2015-11-02T20:41:54-0500 [Controller   8066] Router 'wamprouter': role 'role2' (named 'backend') started on realm 'realm1'
2015-11-02T20:41:54-0500 [Controller   8066] Router 'wamprouter': role 'role3' (named 'frontend') started on realm 'realm1'
2015-11-02T20:41:54-0500 [Router       8070] Site starting on 80
2015-11-02T20:41:54-0500 [Controller   8066] Router 'wamprouter': transport 'web' started
2015-11-02T20:43:32-0500 [Router       8070] Unhandled Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/twisted/web/http.py", line 1445, in dataReceived
    finishCallback(data[contentLength:])
  File "/usr/local/lib/python3.5/site-packages/twisted/web/http.py", line 1703, in _finishRequestBody
    self.allContentReceived()
  File "/usr/local/lib/python3.5/site-packages/twisted/web/http.py", line 1766, in allContentReceived
    req.requestReceived(command, path, version)
[...]

PR #521, opened 2015-11-21
file upload fix and features

markope:
Fixes issues #505 and #504. 

Adds a feature to add extra info to the upload post form data. If using resumable.js in the front end you can add a property `chunk_extra` to the `query` property of the resumable Object. This `chunk_extra` property **must** be a JSON string. ( e.g. `JSON.stringify([your JS object])` ) The form field name `chunk_extra` can be configured in the config.

This json string will be added to the on progress event of the uploaded chunks.
Similarly the `finish_extra` stringified js object will be added to the finish event.

Also added two test cases.


markope:
wait somebody else worked on the fileupload since yesterday evening. I'll merge locally.

PR #524, opened 2015-11-24
Support a configurable threadpool in Crossbar's WSGI support

### LINKED ISSUE #236: Make it possible to configure a number of spawned worker for the WSGI container
Serving Django or Flask with crossbar right now is still not competitive with other solutions such as gunicorn, because you usually need to tweak the number of workers to match your server setup.

You can pass along the thread pool when building the WSGIResource (http://twistedmatrix.com/documents/10.1.0/web/howto/web-in-60/wsgi.html), so it would be a nice touch to be able to configure the number of workers used by the pool in config.json.

hawkowl:
Should fix #236 when complete.

### DIFF
--- a/crossbar/worker/router.py
+++ b/crossbar/worker/router.py
@@ -657,7 +657,7 @@ class RouterWorkerSession(NativeWorkerSession):
             # create Twisted Web root resource
             #
             root_config = config['paths']['/']
-            root = self.create_resource(root_config, path_type='root')
+            root = self.create_resource(root_config, nested=False)
 
             # create Twisted Web resources on all non-root paths configured
             #
@@ -728,7 +728,7 @@ class RouterWorkerSession(NativeWorkerSession):
             if path != b"/":
                 resource.putChild(webPath, self.create_resource(paths[path]))
 
-    def create_resource(self, path_config, path_type='nested'):
+    def create_resource(self, path_config, nested=True):
         """
         Creates child resource to be added to the parent.
 
@@ -844,6 +844,9 @@ class RouterWorkerSession(NativeWorkerSession):
             # Create a Twisted Web WSGI resource from the user's WSGI application object
             try:
                 wsgi_resource = WSGIResource(self._reactor, pool, app)
+
+                if not nested:
+                    wsgi_resource = WSGIRootResource(wsgi_resource, {})
             except Exception as e:
                 raise ApplicationError(u"crossbar.error.invalid_configuration", "could not instantiate WSGI resource: {}".format(e))
             else:
@@ -1033,7 +1036,8 @@ class RouterWorkerSession(NativeWorkerSession):
 
         else:
             raise ApplicationError(u"crossbar.error.invalid_configuration",
-                                   "invalid Web path type '{}' in {} config".format(path_config['type'], path_type))
+                                   "invalid Web path type '{}' in {} config".format(path_config['type'],
+                                                                                    'nested' if nested else 'root'))
 
     def stop_router_transport(self, id, details=None):
         """
--- a/crossbar/worker/router.py
+++ b/crossbar/worker/router.py
@@ -1020,8 +1020,9 @@ class RouterWorkerSession(NativeWorkerSession):
             # Create a threadpool for running the WSGI requests in
             pool = ThreadPool(maxthreads=wsgi_options.get("threads", 20),
                               name="crossbar_wsgi_threadpool")
+            self._reactor.addSystemEventTrigger('before', 'shutdown', pool.stop)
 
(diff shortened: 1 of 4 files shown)


hawkowl:
Puts it in its own threadpool... wonder if I need minimums. Also not sure how to really test this, but I've an idea or two.

hawkowl:
Okay, I think this test is good enough.

cc @oberstet, @meejah , please review.

crossbar/worker/test/test_router.py
```diff
@@ -257,3 +267,194 @@ def test_start_router_component_wrong_baseclass(self):
             str(e.exception.args[0]))
 
         self.assertEqual(len(r.get_router_components()), 0)
+
+
+class WSGITests(TestCase):
+
+    skip = WSGI_TESTS
+
```
meejah: This is a nicer way to skip test-suites than the ``if something: ...all the code...`` idiom :)

crossbar/worker/test/test_router.py
```diff
@@ -257,3 +267,194 @@ def test_start_router_component_wrong_baseclass(self):
...
+    """
+    A super dumb WSGI app for testing.
+    """
+    start_response('200 OK', [('Content-Type', 'text/html')])
+    return [b'hello!']
+
+
+# Ugh global state, but it's just for a test...
+count = []
+
```
meejah: This probably means the 'object' configuration-API needs a way to pass in 'other things', in this case ideally we could pass in this "count" list from the test itself...? Or are WSGI apps "always" written this way?

crossbar/worker/router.py
```diff
@@ -1016,9 +833,19 @@ def create_resource(self, path_config):
                 else:
                     app = getattr(mod, obj_name)
 
-            # create a Twisted Web WSGI resource from the user's WSGI application object
+            # Create a threadpool for running the WSGI requests in
+            pool = ThreadPool(maxthreads=path_config.get("maxthreads", 20),
+                              minthreads=path_config.get("minthreads", 5),
+                              name="crossbar_wsgi_threadpool")
```
meejah: Does this alter the default? That is: what was the default before this change? (Not that it probably really matters, but ...)

meejah: Hmm, it looks like ``twisted/internet/base.py``'s ``_initThreadPoool`` uses ``0`` for minthreads.

hawkowl: Ah -- I picked 5 because that's the default of `threadpool.ThreadPool()`. 0 sounds maybe better?

crossbar/worker/test/test_router.py
hawkowl: WSGI apps are a callable that takes an environment and a `start_response` callable.

The other thing to note is that 'object' is actually JSON -- I've just done it as a raw dictionary for simplicity ;)

crossbar/worker/router.py
meejah: Not sure of the semantics either way ;) but 0 is, I think, the equiv. of what you'd have gotten before these changes (if you don't specify minthreads)...

crossbar/worker/test/test_router.py
meejah: Yeah, I understand it's "usually" JSON but allowing a "real" object in there would let Python programmers (and the tests!) pass in "whatever they like". So, like give ``functools.partial(sleep, count)`` as "object" and change ``sleep`` to take ``count`` as an arg.

Either way; the global ``count`` thing is non-ideal but not that terrible ;)

crossbar/worker/router.py
hawkowl: Sounds reasonable to me.

crossbar/worker/test/test_router.py
hawkowl: The objection I have is if that's a method that end-users can do -- if users can't use it, we shouldn't test it at this level.

meejah: Ah, I see: there's actually no way to pass a ``dict`` at all, you're just doing that for the tests. Sorry, I misunderstood that part. So, okay, convinced the global -- or a very different API ;) -- is required.

meejah:
Hmm, locally some of the Tox environments just hang for me on ``WSGITests.test_basic``. For sure: ``py27-twtrunk-abtrunk``, ...

hawkowl:
Maybe the sleep() is too short? That's the only way I can think of testing concurrency, is by actually blocking 20 threads...

meejah:
Ah, sorry: had a crossbar running on 8080, and some tests rely on listening locally on 8080...

Issue #525, opened 2015-11-25
We should support building universal wheels


hawkowl:
Since Crossbar has no binary extensions of its own, we ought to distribute wheels. Right now, we can't do this due to dynamic dependencies (if statements in setup.py), but it is possible to convert these to environment tags (example; wheel itself https://bitbucket.org/pypa/wheel/src/1cb7374c9ea4d5c82992f1d9b24cf7168ca87707/setup.py?at=default#setup.py-41) as defined in PEP-0426.

This'll mean we can publish universal wheels, which are much nicer to install.

hawkowl:
#528 should fix this

meejah:
Only downside: you need a relatively modern pip (version 6+) to successfully install any of the "extras". This can be achieved on most systems like so (tested on Debian stable/jessie):

```
virtualenv foo
./foo/bin/pip install -U pip  # gives you pip 7+
./foo/bin/pip install /path/to/crossbar.whl[msgpack]  # for example
```

oberstet:
There is a free (for open-source projects) CI packaging service here: https://www.appveyor.com/ which supports building Windows wheels: http://python-packaging-user-guide.readthedocs.org/en/latest/appveyor/

oberstet:
Ignore my last comment .. CB does not (right now) have any binary modules on itself.

oberstet:
What I did was of course stupid, and @hawkowl is right .. the dynamic deps thing is what I forgot about ...

oberstet:
"In wheel, the only way to have conditional dependencies (that might only be needed on certain platforms) is to use environment markers as defined by PEP 426."

https://wheel.readthedocs.org/en/latest/#defining-conditional-dependencies

PR #528, opened 2015-11-26
Allow the building of universal wheels

hawkowl:


meejah:
Changes look fine to me (and work on at least Debian jessie); see #525 for discussion.

PR #531, opened 2015-11-26
Don't require a '/' in web config

### LINKED ISSUE #229: '/' should not be mandatory in  Web transport configuration
Starting crossbar with this configuration:
```
         "transports": [
            {
               "type": "web",
               "endpoint": {
                  "type": "tcp",
                  "port": 8080
               },
               "paths": {
                  "ws": {
                     "type": "websocket"
                  },
                  "/notify": {
                     "type": "pusher",
                     "realm": "realm1",
                     "role": "anonymous"
                  }
               }
            }
         ]
```

Raises: 

```
 Exception: missing mandatory path '/' in 'paths' in Web transport configuration
```
But you may want '/' to return a 404, or just want to put nginx in front wich will serve '/' himself and proxy pass the rest, or put varnish in front and do the same. Having to put an artificial value here for crossbar to work doesn't seem right.

hawkowl:
Should fix #229.

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -1193,9 +1193,6 @@ def check_listening_transport_web(transport):
     if not isinstance(paths, dict):
         raise InvalidConfigException("'paths' attribute in Web transport configuration must be dictionary ({} encountered)".format(type(paths)))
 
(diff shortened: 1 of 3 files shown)


meejah:
LGTM +1

PR #547, opened 2015-12-11
Fix Py3 test failures

hawkowl:
I think it has something to do with changing the dir, making a new temp directory, etc etc.


hawkowl:
cc @oberstet the existing problems are fixed, there's just a 3.5 test failure that needs fixing and a PyPy flakiness issue that needs addressing. But as for what this PR went out to do, I think it's good to land.

PR #558, opened 2015-12-15
Fixes missing tracebacks for exceptions during class imports

Paranaix:
The default ctor of `Failure` automatically invokes `sys.exc_info()`, otherwise we would need to unpack `sys.exc_info()` and not only pass the exception object.

Maybe one should check whether this happens in other places as well.


meejah:
Thanks, and yes you're right in an except it's best to just use ``Failure()`` with no args.

Perhaps even better, since we're using the new loggers would be: ``self.log.error("Failed to import class", log_failure=Failure())`` or I think just ``self.log.failure("Failed to import class")`` will work too. (I haven't tested either of those).

PR #559, opened 2015-12-16
Pass ssl= option to ApplicationRunner for CDC

meejah:
This will work with at least the hostname= option


oberstet:
aahh;) right. thanks!

It all comes back to bite sooner or later ("new API unfinished" .. ha)

PR #563, opened 2015-12-22
Disable return of wildcard registrations until they can be fixed

stopthatcow:
Currently if the uncommented code is run it will raise when this RPC is called. This PR seeks to re-institute partial functionality of wamp.registration.list without implementing the listing of wildcard subscriptions.

### DIFF
--- a/crossbar/router/service.py
+++ b/crossbar/router/service.py
@@ -307,7 +307,7 @@ class RouterServiceSession(ApplicationSession):
                 registrations_prefix.append(registration.id)
 
         registrations_wildcard = []
-        # FIXME 
+        # FIXME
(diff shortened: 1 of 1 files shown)


oberstet:
Sorry, no, this should be fixed instead of commented. Also: http://crossbar.io/docs/Contributing-to-the-project/

stopthatcow:
Ok, I was just bringing it inline with the implementation for listing subscriptions (which currently have a FIXME commented section for wildcards) granted the better solution would be to fix the problem I was unsure how to do that without a better understanding of the wildcard matching code. I will gladly sign on to be a contributor!

oberstet:
@stopthatcow I've moved it to https://github.com/crossbario/crossbar/issues/565

Issue #512, opened 2015-11-11
Crossbar no longer rotates log files


taueres:
With version `v0.10.4` Crossbar logging system was using Twisted [`DailyLogFile`](https://twistedmatrix.com/documents/15.2.1/api/twisted.python.logfile.DailyLogFile.html) to manage logging to file-system ([look at the source code](https://github.com/crossbario/crossbar/blob/07f33b4e57ad3821ae4706e32b37f3c159ae4355/crossbar/controller/cli.py#L385)).
This was daily rotating log files which is, IMHO, very useful.

Unfortunately since version `v0.11.0` Crossbar uses [`FileLogObserver`](https://twistedmatrix.com/documents/15.2.1/api/twisted.logger.FileLogObserver.html) which no longer rotates log files automatically ([look at the source code](https://github.com/crossbario/crossbar/blob/0fbdfa6434e107ea8ece54a2d6e6df738c5f4c36/crossbar/_logging.py#L318)).

Crossbar documentation currently says that log rotation is handled by Crossbar itself: [look here the explanation about --logdir option](http://crossbar.io/docs/Automatic-startup-and-restart/).

I think the best solution is to enable again log rotation, clearly this is the intended behavior.

hawkowl:
I think what Twisted used previously (`DailyLogFile`) is now ported, so we could revert to using that again, once 15.5 comes out.

PR #594, opened 2016-01-16
update to twisted 15.5 #512

siscia:
I updated the manifest to use twisted 15.5

Mainly to work for #512


oberstet:
this is already contained in your other PR

PR #595, opened 2016-01-16
Usedaylog

### LINKED ISSUE #512: Crossbar no longer rotates log files

siscia:
I am using the old `DailyLogFile` to fix #512 

I simply run the unit test, didn't yet test if the log really circle daily. 

Thanks to @taueres and @danielemegna


oberstet:
@siscia thanks! before merging, we need a CAA (that is a one-time thing only): http://crossbar.io/docs/Contributing-to-the-project/#send-us-a-contributor-assignment-agreement

siscia:
@oberstet Done, I send the CAA

oberstet:
@siscia thanks!

Issue #598, opened 2016-01-19
Logging docs


oberstet:
rgd https://github.com/crossbario/crossbar/pull/573

@hawkowl Hi, I nearly forgot again that you did already add this;) Could you probably add a small section "best practices with CB logging for CB devs" here https://github.com/crossbario/crossbar/blob/master/DEVELOPERS.md?

Also @meejah Another thing someone was asking me (and I was waving hands): how to do logging "correctly" for app developers using AutobahnPython ApplicationSession via txaio .. would be cool to have a doc page in AutobahnPython on logging too: https://github.com/crossbario/autobahn-python/tree/master/doc

meejah:
Yes, good idea.

I believe the right answer is to use the idioms found in AB/crossbar. Namely, ``log = txaio.make_logger()`` and use it more-or-less like the Twisted new logger: ``self.log.error("{some_arg} is bad", some_arg='foo')``

...but, it should also be possible to interop with old-style twisted logging or Python ``logging`` built-in logging. However, "green field" code should use the structed-logger style like the first paragraph. Also, for error formatting etc there are ``txaio.failure_*()`` functions that should be used on the txaio failure instances you get in the errbacks...

oberstet:
Sounds good! Probably add more reference like logging docs to http://txaio.readthedocs.org/en/latest/, and more usage oriented docs to the AutobahnPython docs?

hawkowl:
also adding a note here to remind myself to document the new `log_category` changes

Paranaix:
Question: Whats the reasoning behind the crossbar logging module being private? 
Whats the best-practice for logging for applications built on top of crossbar. 
Whats the best-practice if more sophisticated logging is required (e.g log security relevant stuff to an extra file)?

meejah:
@Paranaix see http://txaio.readthedocs.org/en/latest/programming-guide.html#logging

oberstet:
@Paranaix The new shiny structured logging is now part of the public API of txaio, and hence usable for all. Docs are in above

FirefighterBlu3:
txaio logging is unable to handle content that has braces in it:

```python
            resp = {
                'realm': principal['realm'],
                'role':  principal['role'],
                'extra': {
                    'roleAdmin':   principal['roleAdmin'],
                    'jpegPhoto':   '<suppressed>',
                    'department':  principal['department'],
                    'displayName': principal['displayName']
                }
            }

[...]

PR #599, opened 2016-01-21
Supporting code for checking logs without asserting the rendered value

hawkowl:
cc @oberstet @meejah 

As a proof of concept, I modified the `test_common` tests to use something that might line up with what both of you mentioned -- including a nicer way of capturing logs.

This puts the log cateories and their general meanings in `_log_categories.py` -- the idea behind this is that this could be pulled out in docs generation, or something.

Thoughts?

### DIFF
--- a/crossbar/_log_categories.py
+++ b/crossbar/_log_categories.py
@@ -38,12 +38,17 @@ log_categories = {
 
     # ARXXX - Adapter, REST Bridge
     "AR100": "REST bridge request recieved.",
+    "AR200": "REST bridge publish succeeded.",
     "AR400": "Malformed request to the REST bridge.",
     "AR405": "Method not accepted by the REST bridge.",
     "AR413": "Request too long.",
     "AR450": "Non-accepted request encoding.",
     "AR451": "Non-decodable request body.",
     "AR452": "Non-accepted content type.",
+    "AR453": "Request body was invalid JSON.",
+    "AR454": "Request body was valid JSON, but not well formed (incorrect type).",
+    "AR455": "Request body was valid JSON, but not well formed (missing key).",
+    "AR456": "REST bridge publish failed.",
 }
 
 
--- a/crossbar/_log_categories.py
+++ b/crossbar/_log_categories.py
@@ -39,8 +39,11 @@ log_categories = {
     # ARXXX - Adapter, REST Bridge
     "AR100": "REST bridge request recieved.",
     "AR400": "Malformed request to the REST bridge.",
-    "AR405": "REST bridge does not support that method.",
-
+    "AR405": "Method not accepted by the REST bridge.",
+    "AR413": "Request too long.",
+    "AR450": "Non-accepted request encoding.",
+    "AR451": "Non-decodable request body.",
+    "AR452": "Non-accepted content type.",
 }
 
 
--- a/crossbar/_log_categories.py
+++ b/crossbar/_log_categories.py
@@ -1,0 +1,47 @@
+#####################################################################################
+#
+#  Copyright (C) Tavendo GmbH
+#
+#  Unless a separate license agreement exists between you and Tavendo GmbH (e.g. you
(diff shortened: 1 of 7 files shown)


oberstet:
@hawkowl I really like this! I think it achieves the "best of both". It makes the structured log information (with the log error cat.) a well-defined "API", and unit test can test that - without the formatted log message standing the the way. Yay=) And the possibility to auto-generate docs is yet another advantage which is quite cool and useful.

crossbar/adapter/rest/common.py
```diff
@@ -161,14 +168,16 @@ def _render_request(self, request):
             content_type_elements = []
 
         if self.decode_as_json:
-            # iff the client sent a content type, it MUST be one of _ALLOWED_CONTENT_TYPES
+            # if the client sent a content type, it MUST be one of _ALLOWED_CONTENT_TYPES
```
oberstet: "iff" (with double f) is math slang for "if and only if" ;)

crossbar/_log_categories.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # ARXXX - Adapter, REST Bridge
+    "AR100": "REST bridge request recieved.",
+    "AR400": "Malformed request to the REST bridge.",
+    "AR405": "Method not accepted by the REST bridge.",
+    "AR413": "Request too long.",
+    "AR450": "Non-accepted request encoding.",
+    "AR451": "Non-decodable request body.",
+    "AR452": "Non-accepted content type.",
+}
```
oberstet: So `log_categories` would essentially contain all errors that can happen anywhere inside CB, right?

crossbar/adapter/rest/common.py
```diff
@@ -96,8 +96,12 @@ def _deny_request(self, request, code, reason, **kwargs):
         """
         Called when client request is denied.
         """
+        if "cb_log_id" not in kwargs.keys():
```
oberstet: Minor thing (naming) .. but maybe `log_category`?

hawkowl: from slack, my reason for picking this was that the `cb_` namespacing means we might not conflict with anything, but that is not a big deal, and `log_category` is easier to type/remember.

crossbar/adapter/rest/common.py
hawkowl: ahh, I might replace it with the words then (since I don't come from a math background, so I thought it was a typo :) )

crossbar/_log_categories.py
hawkowl: Yes -- but I was thinking more logs of any kind -- errors, warnings, successes all have codes. For example, `AR100` is "received a request" -- it's not an error, just an informational code, but it's there. This comes in really handy if you have logs going somewhere, as it's a token you can search on. For example, in Windows, to audit all successful logins, you go to the Log Viewer and make a search for the "successful authentication" log message -- it's really handy.

And yes, all things should have it. When enough do, we can start asserting on requiring one as a test suite, to make sure all our log messages are categorised, and that nothing slips through the cracks.

Currently the category doesn't show in the output -- I am thinking that maybe adding it to the start of every message in "debug mode", and to the start of all error messages in standard modes, would be a good idea? Not sure.

hawkowl: The only other thing which I thought of was making this where the *log templates* are kept, too -- but I am not sure if that makes it too complicated (what generates the log and the format for it are then really far away, code wise), but it was something that went through my mind. Not everything that does is a good idea, but I thought I'd mention it ;)

hawkowl:
@oberstet Hurrah, I'm glad you like the approach. I'll polish up this tomorrow (getting late here :) ) and then we can start using this -- https://github.com/crossbario/crossbar/issues/598 I'll also add usage information in this ticket.

oberstet:
coool;) thanks!

hawkowl:
I think I broke something -- the tests failed, let me just fix it in a followup:)

PR #601, opened 2016-01-23
Further REST bridge logging+testing improvements

hawkowl:


hawkowl:
cc @oberstet this should fix it all up, I think :)

oberstet:
perfect, thanks!

oberstet:
I will start to write log calls using the new `log_category` from now on. This is cool!

oberstet:
@hawkowl One question: both the keys and values in `crossbar/_log_categories.py` are not unicode, but bytes. Is that intentional?

hawkowl:
They're `str`, which is mainly as a source of convenience for us -- because it's a string literal, and Python source files are UTF-8 (and in this case just the ASCII subset), they're un-ambigiously encodable/decodable.

However, if you think it could cause a future issue, I'm happy to change them all to Unicode literals, and make it assert if the log_category item is not Unicode.

oberstet:
Ok, right. I agree, if we'd only use `log_category` within CB code, that would be convenient indeed.

But we likely want to forward the structured log stream to CDC - including `log_category`. Which means, it'll travel over the wire .. needs to be serialized.

So, yes, I am +1 on making both keys and values unicode, and adding an assert.

hawkowl:
Great, I can handle that :) I'll open an issue (since it's late here, and I don't want it to slip through the cracks)

oberstet:
cool! thanks

hawkowl:
@oberstet filed at https://github.com/crossbario/crossbar/issues/602 , add any notes if there's anything I've missed :)

PR #608, opened 2016-01-24
pass fail as simple argument, not as keyword arg

yatskevich:
Otherwise logging itself will fail with:

```
  File "/usr/local/lib/python2.7/site-packages/crossbar/router/session.py", line 477, in _swallow_error_and_abort
    self.log.failure("Internal error (2): {log_failure.value}", log_failure=fail)
  File "/usr/local/lib/python2.7/site-packages/crossbar/_logging.py", line 362, in failure
    return self.logger.failure(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/twisted/logger/_logger.py", line 178, in failure
    self.emit(level, format, log_failure=failure, **kwargs)
exceptions.TypeError: emit() got multiple values for keyword argument 'log_failure'
```

### DIFF
--- a/crossbar/router/session.py
+++ b/crossbar/router/session.py
@@ -513,7 +513,7 @@ class RouterSession(BaseSession):
 
         DO NOT attach to Deferreds that are returned to calling code.
         """
-        self.log.failure("Internal error (2): {log_failure.value}", log_failure=fail)
+        self.log.failure("Internal error (2): {log_failure.value}", fail)
 
         # tell other side we're done
         reply = message.Abort(u"wamp.error.authorization_failed", u"Internal server error")


oberstet:
@hawkowl shouldn't structured log items always be passed as kwargs?

@yatskevich thanks for contributing! if you want code to get merged, we need a CAA (once): http://crossbar.io/docs/Contributing-to-the-project/ - otherwise (as in this case, it's only 1 line), let me know, and I do the change - depending on @hawkowl comments

yatskevich:
@oberstet go ahead and do the change :wink:

hawkowl:
We shouldn't need to pass it, if it's currently being handled -- because `.failure()` will get the failure off the stack. Otherwise, it should be `failure=fail`, which will then get turned into the `log_failure` event key.

hawkowl:
(see the docs here https://twistedmatrix.com/documents/15.5.0/api/twisted.logger.Logger.html#failure )

oberstet:
@hawkowl like this? `self.log.failure("Internal error (2): {log_failure.value}", failure=fail)`

hawkowl:
I believe that should work -- if you are in an errback or except clause, you can omit `failure=fail` completely, as it grabs the one in flight.

oberstet:
https://github.com/crossbario/crossbar/blob/master/crossbar/router/session.py#L516

PR #609, opened 2016-01-25
Support turning on/off coloured logging

### LINKED ISSUE #600: No color with logging for "crossbar version"
A minor thing, but I am wondering why `crossbar version` doesn't do colored output, while `crossbar start` does.

hawkowl:
This fixes #600, and makes the split between format and colouring nicer. It also makes piping to a file better, as it'll automatically disable colouring.

### DIFF
--- a/crossbar/controller/cli.py
+++ b/crossbar/controller/cli.py
@@ -570,7 +570,8 @@ def run_command_start(options, reactor=None):
     # bannerFormat = "{:<18} {:<24}"
     bannerFormat = "    {} {}"
     log.info(bannerFormat.format("Crossbar.io Version:", click.style(crossbar.__version__, fg='yellow', bold=True)))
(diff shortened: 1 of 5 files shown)


hawkowl:
Seems like pypy is broken because of pynacl needing newer PyPy.

crossbar/controller/cli.py
```diff
@@ -859,10 +891,10 @@ def run(prog=None, args=None, reactor=None):
     else:
         options.argv = sys.argv
 
-    # colored logging does not work on Windows, so overwrite it!
+    # coloured logging does not work on Windows, so overwrite it!
     #
```
oberstet: =) sorry, I am just used to the other form

oberstet:
Yeah, PyPy is broken because Travis still doesn't have a current PyPy.

Issue #606, opened 2016-01-23
Least-busy invocation policy


oberstet:
Implement a least-busy invocation policy: a call incoming to the router is forwarded to the callee with the least outstanding invocations.

oberstet:
When there are multiple callees with a least number of outstanding invocations, a tie breaker is needed. This could be

* random
* round-robin

If this is to be exposed, we could have:

* `leastbusy_random`
* `leastbusy_roundrobin`

oberstet:
Regarding mode naming: we probably want to have yet another mode: "quickest reaction time".

Above "least busy" is for maximizing throughput / for situations where the numbers of calls is large. But imagine callees of different speed, a user might want to direct calls to the "quickest callees" first ..

oberstet:
After thinking about a little, I think it makes more sense to stick with a single `leastbusy` policy and leave the tie-breaking mechanism as an implementation detail.

markope:
+1 to your last comment

meejah:
If we tracked the lastest response-time for each observer, this could also provide useful metrics. We could the store observers in some sorted structure (e.g. ``heapq`` sorted by latest-response-time) and just select the front of the list every time.

The resulting behavior would basically just pound the fastest component until it started to get slow ... which I guess is kind of the point? (That's just my hand-wave-y guess though :) )

Keeping the observers sorted might be more pain than it's worth though; O(n) isn't that bad if n is usually only "a few"...

oberstet:
@meejah 

1) Rgd O(n) .. yeah, agreed. Linear search is likely faster for small n. Sharing registrations between 1000s of callees is an unlikely use case. A hundred callees should be possible though.

2) Tracking response times: definitely valuable in general!

3) `heapq`: Yeah, I was also thinking about some priority queue .. things is, if the value sorted by would be the number of outstanding invocations, that obviously changes. 

The Py docs have this to say: https://docs.python.org/2/library/heapq.html#priority-queue-implementation-notes

"If the priority of a task changes, how do you move it to a new position in the heap?"

That's exactly my question;) Where is the answer? Does `heapq` support changing prios without removing elements?

oberstet:
I think both invocation policies would make sense:

*  `leastbusy` => direct invocation to callee with the least number of outstanding invocations
*  `quickestresponse` => direct invocation to callee with "quickest response"

With the latter, the question is: what is _the_ quickest response. Only the last one? Average over last N?

mfojtak:
[...]

PR #613, opened 2016-01-26
Add shared registrations balanced mode

markope:
This is a suggested extra mode for shared registrations.
Shared registrations can be used with mode "balance" in addition to the existing "roundrobin, random, ..." in this pull request.

If a call is requested on the shared URI then crossbar randomly picks one of the registered callees that is not currently busy with an ongoing request. If all registrations are busy then crossbar picks ~~a busy callee randomly~~ the least busy callee by number of pending invocations.

Hope the feature is welcome. I'm happy to discuss...

### DIFF
--- a/crossbar/router/dealer.py
+++ b/crossbar/router/dealer.py
@@ -1,4 +1,4 @@
-##########################################################################
+#####################################################################################
 #
 #  Copyright (C) Tavendo GmbH
 #
@@ -26,7 +26,7 @@
 #  You should have received a copy of the GNU Affero General Public license along
 #  with this program. If not, see <http://www.gnu.org/licenses/agpl-3.0.en.html>.
 #
-##########################################################################
+#####################################################################################
 
 from __future__ import absolute_import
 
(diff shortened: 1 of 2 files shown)


markope:
Also there is an AutobahnPython and a crossbar example pull request coming up with related changes. The example also served as the test.

dustinlacewell:
Why not count how many invocations are pending against each observer and sort them?

markope:
ok, thanks. I improved that now.

crossbar/router/dealer.py
```diff
@@ -386,6 +388,18 @@ def on_authorize_success(authorized):
                     elif registration.extra.invoke == message.Register.INVOKE_RANDOM:
                         callee = registration.observers[random.randint(0, len(registration.observers) - 1)]
 
+                    elif registration.extra.invoke == message.Register.INVOKE_BALANCE:
```
oberstet: This code would be a bottleneck ..

oberstet:
see: https://github.com/crossbario/crossbar/issues/606

markope:
what exactly do you mean by 

> ...would be a bottleneck...

Needs improvement or good enough?
Should I adjust anything else apart from the mode name?

Issue #616, opened 2016-01-29
Don't log deliberate exceptions from user/service code


oberstet:
Raising exceptions from service sessions (`crossbar.router.service.RouterServiceSession`) should NOT be logged. This is similar to running user sessions/components inside a router/container. Such user code explicitly raising should also not log. An exception raised from user code often is used as a valid means to signal call errors.

E.g. consider https://github.com/crossbario/crossbar/blob/master/crossbar/router/service.py#L531

This code may very well raise `wamp.error.history_unavailable` -- and we don't want to clutter the log with:

```console
2016-01-29T14:20:49+0100 [Controller  12617] Router 'worker1': transport 'transport1' started
2016-01-29T14:20:51+0100 [Router      12621] wamp.error.history_unavailable: 
```

meejah:
I'll take a look, but I *think* all you should have to do is override ``onUserError`` to be a no-op (or ignore specific ApplicationError URIs) in ``RouterServiceSession``.

oberstet:
Ah, right. Thanks! I will try that. Essentially, all errs derived of ApplicationError should just go through unlogged, while anything else should be logged .. because it needs fixing, because it should be treated explicitly and use ApplicationError

meejah:
See the PR.

You will of course still get a log if the client-session doesn't handle the exception (that is, the thing doing ``yield call(u"wamp.subscription.get_events", ..)`` needs to handle the exception).

PR #618, opened 2016-01-29
avoid logging ApplicationErrors in the service-session

meejah:
See #616


oberstet:
This didn't work for me (_everything_, including ApplicationErrors was now swallowed). This does work: https://github.com/crossbario/crossbar/commit/334ab439702a0723841d8b4732f6a1bc5db00d86

PR #629, opened 2016-02-01
Fixed obsolete use of 'traceback'

Paranaix:
In case we work with @inlineCallbacks, *traceback* is not able to show the interessting part (where the actual exception is thrown). For consistency we should always use twisteds *Failure* class instead of *traceback*.

EDIT: Fixed another similiar issue resulting in no log message at all. (btw the travis-ci flake8 runner seems broken)


oberstet:
Yep, this should be changed, but differently: `log.error("lala", log_failure=e)`. Moreso: `log.error("lala", log_failure=e, log_category=u'XY123')`.

PR #630, opened 2016-02-01
Fixed crash during node start under Python 2.x

Paranaix:
ComponentConfig fires an assertion otherwise.


oberstet:
Ah, thanks! Fixed in https://github.com/crossbario/crossbar/commit/347695ce911ab3a730571aaf33fdd8eddab86edb

PR #631, opened 2016-02-02
Log wrapping and indentation

Paranaix:
The main motivation behind this PR was to have multiline logs properly indented. The result is something like this:

image

This PR adds a `--logwidth` cli option which can be used to specify wrapping width (required for indentation). It defaults to 0 which leaves the current behaviour unmodified unless desired.

The drawback of this approach is, that it's not further possible to arbitrarly format logs but that formats can only be used to generate a prefix. So something like `[TIME] Log [SUFFIX]` is not possible anymore.

While implementing this I also slightly refactored the logging code to have redundant code in a single place.

### DIFF
--- a/crossbar/_logging.py
+++ b/crossbar/_logging.py
@@ -80,23 +80,23 @@ try:
 except ImportError:
     # No colorama, so just mock it out.
     class _Fore(object):
-        BLACK = ""
-        RED = ""
-        GREEN = ""
-        YELLOW = ""
-        BLUE = ""
-        MAGENTA = ""
-        CYAN = ""
-        WHITE = ""
-        RESET = ""
-        LIGHTBLACK_EX = ""
-        LIGHTRED_EX = ""
-        LIGHTGREEN_EX = ""
-        LIGHTYELLOW_EX = ""
-        LIGHTBLUE_EX = ""
-        LIGHTMAGENTA_EX = ""
-        LIGHTCYAN_EX = ""
-        LIGHTWHITE_EX = ""
+        BLACK = u''
+        RED = u''
+        GREEN = u''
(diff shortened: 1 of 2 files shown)


oberstet:
This does a number of changes I don't agree with

Paranaix:
What kind of changes do you exactly don't agree with? This PR leaves the current behaviour completely untouched. I've also discovered that *click* features a `get_terminal_size` function which we could utilize. In that case we can get rid of the `--logwidth` option and introduce a simple `--logindent` option to toggle between the old and new behaviour.

Regarding the restrictions, as this seems a bit unclear to you: Currently its possible to have format string like: `{time} FOO {text} BAR`. Obviously indenting prevents this kind of format and only things like  `{time} FOO {text}` are possible anymore.

oberstet:
Let's try to ping @hawkowl ;)

My view on this is mainly driven by the current state of "unresolved" other issues related to logging (introduce more feature while the issues are not yet solved):

- are we going to use txaio for logging in CB and deprecated crossbar._logging?
- no "logging how to / idioms" for CB devs (when do I use what idiom? this has all gone through some evolution)
- roll out the "log_category" feature over the whole code base first
- verify that our original vision works: structured logging to a WAMP event stream, receiving the structured log info at the other end
- what about the colored/formatted object logging (was that JSON prettify thing)

All of above is far more important in my view compared to wrapping lines. I am not against that .. it's just IMO we should not start even more open ends before closing a bunch first. And I don't feel like able to analyze the implications of the PR.

Paranaix:
In that case I would however advise you to merge my first commit. It's some basic refactoring making the logging code a bit less duplicated.

Issue #615, opened 2016-01-29
Systematic handling of worker and component lifecycle behavior


oberstet:
For parts of this, see the discussion here https://github.com/crossbario/crossbar/issues/327

PR #634, opened 2016-02-05
Fire on_component_start and on_component_stop properly

meejah:
This fixes (some of the) behavior required for #615 -- specifically in this case firing on_component_start and on_component_stop properly for router workers.

Requires https://github.com/crossbario/autobahn-python/pull/593 to work correctly.


oberstet:
Ah, this is using the new observer thing on Autobahn .. nice;)

Issue #612, opened 2016-01-26
Service readiness notification


oberstet:
When running Crossbar.io under `systemd`, there is often the need to have dependent services start, when Crossbar.io has **fully started**.

However, this is not possible right now, since even when adding a respective dependency to the service unit, `systemd` will start the dependent service immediately when the Crossbar.io **process has started**.

There is a systemd mechanism for this situation call **service readiness notification**. 

Here is a Python example: https://github.com/openstack/oslo.service/blob/master/oslo_service/systemd.py

bb4242:
This functionality would be great to have.  I created #637 to address this; please let me know what you think.  Also, if there is a good place to add documentation about this feature (possibly in [crossbardocs](https://github.com/crossbario/crossbardocs)?), please let me know and I'll be happy to add some.

oberstet:
@bb4242 awesome!! also nice: not exactly a lot of code for a quite useful feature;) Before I can merge, may I ask you to check out http://crossbar.io/docs/Contributing-to-the-project/#send-us-a-contributor-assignment-agreement ? because we would need (but only once) a contributor agreement. yeah, I know .. forms;)

bb4242:
Sure thing, I just submitted a contributor agreement to service@tavendo.de.

oberstet:
@bb4242 yep, received. thanks! closing via merged https://github.com/crossbario/crossbar/pull/637

bb4242:
Great, thanks!

PR #637, opened 2016-02-08
Add systemd notification support.

bb4242:
This allows running crossbar as a systemd service unit that notifies systemd
when it is up and running. This allows other services that depend on crossbar
being up to be started correctly.

Addresses #612

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -495,6 +495,15 @@ class Node(object):
 
         try:
             yield self._startup(self._config)
+
+            # Notify systemd that crossbar is fully up and running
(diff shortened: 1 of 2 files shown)


bb4242:
Here's an example `systemd` unit file:

    [Unit]
    Description=Crossbar WAMP router

    [Service]
    # Set unbuffered python to see output from crossbar immediately (otherwise it will be buffered)
    Environment=PYTHONUNBUFFERED=true
    ExecStart=/usr/local/bin/crossbar start --cbdir /path/to/crossbar/config
    Type=notify

    [Install]
    WantedBy=default.target

PR #641, opened 2016-02-10
Set PYTHONUNBUFFERED in child processes

### LINKED ISSUE #506: Logging should use unbuffered output
When running Crossbar under systemd, log output doesn't appear because of buffering.

Adding `-u` to the `crossbar` script installed works, but
- requires patching the installed script
- not sure about side-effects (we only want unbuffered output for the node controller logging)

``` python
#!/opt/crossbar/bin/pypy -u

# -*- coding: utf-8 -*-
import re
import sys

from crossbar.controller.cli import run

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(run())
```

meejah:
This doesn't set the environment variable in case the user
has configured it already (either via an inherited value or
explicitly in the config).

Should fix #506

### DIFF
--- a/crossbar/controller/process.py
+++ b/crossbar/controller/process.py
@@ -1040,6 +1040,12 @@ def create_process_env(options):
     """
     penv = {}
 
+    # Usually, we want PYTHONUNBUFFERED set in child processes, *but*
+    # if the user explicitly configures their environment we don't
+    # stomp over it. So, a user wanting *buffered* output can set
+    # PYTHONUNBUFFERED to the empty string in their config.
+    saw_unbuff = False
+
     # by default, a worker/guest process inherits
     # complete environment
     inherit_all = True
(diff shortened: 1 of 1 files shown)


oberstet:
But this "only" sets unbuffered mode for workers, not the controller?

meejah:
I don't have a specific test-case for the controller, but looking at the code this only affects native-workers (and guest processes). (``create_process_env``)

oberstet:
I was asking, because if the controller is still running buffered, logging which comes (unbuffered) from workers still don't get written out immediately.

I was myself running into this recently (and not the first time): running CB from our binary package install the `crossbar` run script - and doesn't invoke Python in unbuffered mode. I created a systemd service and was wondering why the heck the logging doesn't appear.

IOW: is it possible to set unbuffered mode on the controller too? That would remove this trap ..

meejah:
Ah, I see. I will look into it -- likely we have to pass options to setup.py to make the "crossbar" script invoke with this option ... "or something".

Issue #642, opened 2016-02-11
OrderedDict.move_to_end is missing in Python2


touilleMan:
```
$  crossbar upgrade
Automatically choosing optimal Twisted reactor
Running on Linux and optimal reactor (epoll) was installed.
Upgrading local configuration file /home/emmanuel/projects/flask-socketio-test/.crossbar/config.json
Upgrading configuration from version 1 to version 2

Error: 'OrderedDict' object has no attribute 'move_to_end'
```

The trouble seems to be here https://github.com/crossbario/crossbar/blob/master/crossbar/common/checkconfig.py#L2779
`move_to_end` has been added in Pythen 3.2 (see https://docs.python.org/3/library/collections.html#collections.OrderedDict.move_to_end)

oberstet:
fixed in trunk

PR #643, opened 2016-02-11
Fix upgrade command for python2.7

touilleMan:
Following issue #642, a fix to get rid of `move_to_end` in the upgrade command

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2775,7 +2775,7 @@ def upgrade_config_file(configfile):
 
         version += 1
 
-    config = OrderedDict([(u'version', version)]+ [i for i in config.items()])
+    config = OrderedDict([(u'version', version)] + [i for i in config.items()])
 
     with open(configfile, 'wb') as outfile:
         data = json.dumps(config,
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2775,7 +2775,7 @@ def upgrade_config_file(configfile):
 
         version += 1
 
-    config = OrderedDict([(u'version', version)] + [i for i in config.items()])
+    config = OrderedDict([(u'version', version)] + list(config.items()))
 
     with open(configfile, 'wb') as outfile:
         data = json.dumps(config,
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2775,8 +2775,7 @@ def upgrade_config_file(configfile):
 
         version += 1
 
-    config[u'version'] = version
-    config.move_to_end(u'version', last=False)
+    config = OrderedDict([(u'version', version)]+ [i for i in config.items()])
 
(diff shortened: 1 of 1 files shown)


oberstet:
Ah, right. Thanks for the fix! May I ask you to follow http://crossbar.io/docs/Contributing-to-the-project/#send-us-a-contributor-assignment-agreement ? We need a contributor agreement before I can merge. This is a one time thing though!

crossbar/common/checkconfig.py
```diff
@@ -2775,8 +2775,7 @@ def upgrade_config_file(configfile):
 
         version += 1
 
-    config[u'version'] = version
-    config.move_to_end(u'version', last=False)
+    config = OrderedDict([(u'version', version)] + [i for i in config.items()])
```
meejah: Can the list-comprehension be simplified to ``list(config.items())``? Not a big deal either way though. :)

Paranaix:
For what even any list comprehension, `items()` already returns a list? Most performant statement would be `OrderedDict([[(u'version', version)].extend(config.iteritems()))`

oberstet:
FWIW, the idea behind this code is: upgrade a config file with minimal impact on the file formatting a user may have done. Hence the need for an ordered dict that can roundtrip to/from json.

But actually, even more neat would be an option to "prettify" a config file. And that would include canonical orderings of attributes. A CB node configuration file essentially contains nested dicts, and the attributes that can appear in those dicts is fixed - with one exception: "paths" in a Web transport (and we could change that so that the URL paths appear as attribute values instead of keys).

So maybe we could have our own, special dict, suited to our needs. Eg a TLS listening config item:

*  `ConfigDict(["key", "certificate", "chain_certificates", "ca_certificates", "ciphers", "dhparam"])

The ctor array would fix the "canonical" (= prettified) attribute order.

Maybe we could even refactor the biig `checkconfig.py` (which historically grew from just a quick check fun, to a large number of check funs) into an active object: `Configuration` which contains both type / value checking and attribute ordering. A (top-level) `Configuration` could in turn contain `Configuration` items down to the smallest items (like a TLS listening config item as above). And then we could add docstrings and _generate_ the docs for CB node configuration.

Alright: now I have been taken away;) I am fine with the patch as is ..

waterbug:
One more FWIW:  pyyaml reads/writes *very* pretty config files. ;)

oberstet:
YAML is already supported (and there is even support for automatic conversion JSON<=>YAML). But that doesn't make it "pretty" auto-magically;)

touilleMan:
@meejah I've updated my PR according to your comment

@Paranaix `iteritems` has been removed from python3, given `items` doesn't return the same thing in python 2 (list of element) and 3 (builtins.dict_items) when must enclosed it into a `list()` call

oberstet:
Thanks @touilleMan : received contributor agreement. Ah, and thanks for the patch of course;)

PR #657, opened 2016-02-23
Protocol: fix memory leak when using cookies

taueres:
The method `WampWebSocketServerProtocol::onDisconnect` was not invoked when needed.
So the protocol set in `CookieStore` was growing up without limit during process lifetime.
This fix removes the protocol reference and overrides the `onClose` method.

### DIFF
--- a/crossbar/router/protocol.py
+++ b/crossbar/router/protocol.py
@@ -163,6 +163,10 @@ class WampWebSocketServerProtocol(websocket.WampWebSocketServerProtocol):
     """
     log = make_logger()
 
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


peelandsee:
hi!
any chance to merge this?

we are experiencing memory leaks too,
this bug prevent us to use cookies in production!

thanks @taueres for reporting and proposing the fix!

meejah:
Thanks for the bump. I had put "needs-investigation" on this because I'd like to understand why onDisconnect wasn't/isn't getting called -- but that doesn't have to stop this getting merged.

taueres:
I've just rebased this PR against master branch

PR #661, opened 2016-03-01
Fix update cookies while reading from file

taueres:
CookieStorage now updates cookie dictionaries rather than overwriting them. This fixes the following bug: sometimes at line 256 the 'created' key was not present.


oberstet:
Thanks @taueres for the CAA! And thanks for the patch of course;)

Issue #578, opened 2016-01-03
Traceback event handler


oberstet:
When an even handler was subscribed that can't consume the event args/kwargs, this throws an error. Check that in general, but in particular with component running embedded:

```
2016-01-04T00:11:08+0100 [Router       5358] Fatal error in component: While firing <function onhello at 0x7ff61ebabb18> subscribed under 5174327938517943. - onhello() takes exactly 1 argument (0 given)
2016-01-04T00:11:08+0100 [Router       5358] Unhandled error in Deferred:
2016-01-04T00:11:08+0100 [Router       5358] 
Traceback (most recent call last):
  File "/home/oberstet/scm/crossbario/crossbar/crossbar/router/router.py", line 170, in process
    self._broker.processPublish(session, msg)
  File "/home/oberstet/scm/crossbario/crossbar/crossbar/router/broker.py", line 337, in processPublish
    txaio.add_callbacks(d, on_authorize_success, on_authorize_error)
  File "/home/oberstet/scm/crossbario/txaio/txaio/tx.py", line 392, in add_callbacks
    future.addCallbacks(callback, errback)
  File "/home/oberstet/python279_1/lib/python2.7/site-packages/twisted/internet/defer.py", line 306, in addCallbacks
    self._runCallbacks()
--- <exception caught here> ---
  File "/home/oberstet/python279_1/lib/python2.7/site-packages/twisted/internet/defer.py", line 588, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/oberstet/scm/crossbario/crossbar/crossbar/router/broker.py", line 313, in on_authorize_success
    for receiver in receivers:
exceptions.RuntimeError: Set changed size during iteration
```

meejah:
Okay, so what's happening here is that the Broker works out the list of subscription receivers and then loops over these, send()ing the publish messages. Upon an error, the connecting is failed.

In the embedded-router case, the ``send()`` is actually synchronous as ``RouterApplicationSession`` is behaving as the ITransport -- and so when the connection fails, it ``detach()``es from the Broker and hence drops all observations, changing the "receivers" set-size.

[...]

PR #665, opened 2016-03-02
Copy set of receivers in case it changes under us

meejah:
See ticket #578: if an embedded component raises an exception
from a subsciption callback, the connection is removed -- causing
the observations to be dropped. This is synchronous as the
ITransport is RouterApplicationSession.

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -257,8 +257,12 @@ class Broker(object):
                             self._event_store.store_event_history(publication, subscription.id)
 
                         # initial list of receivers are all subscribers on a subscription ..
+                        # note: it's wrapped in set() because we really do want
+                        # to copy it in case the size of the set changes while
+                        # we're sending out publishes (e.g. upon disconnect or
+                        # errors)
                         #
-                        receivers = subscription.observers
+                        receivers = set(subscription.observers)
 
                         # filter by "eligible" receivers
                         #


oberstet:
Copying will remove this issue, but will produce more GC pressure. Imagine a set of 100k subscribers to a topic. Each publish will copy the set. Do you see a way to "fix it" without copying?

meejah:
Yes, either of the other approaches mentioned in #578 will work: a ``reactor.callLater(0, ...)`` on either the ``.detach()`` or upstream ``onMessage`` or ``.send()`` -- but then that adds to the "things in reactor queue" instead of "copies of the receiver sets"... :/

meejah:
I guess the "least widespread consequences" would be to simply wrap the ``.detach()`` in a ``reactor.callLater(..)`` (here: https://github.com/crossbario/crossbar/blob/master/crossbar/router/session.py#L133). (This does work, at least for my test-case for the original problem).

meejah:
Another alternative would be to store the observers as a ``list()`` instead of a set since you can iterate a list and (safely) change the size of it while iterating. But, then we have to keep uniqueness etc... :/

oberstet:
@meejah So I close this one, as I merged your other PR https://github.com/crossbario/crossbar/pull/668 - which does not carry the overhead that this PR would have had.

PR #668, opened 2016-03-03
Use a reactor-loop to avoid removing observers until "later"

meejah:
A different way to solve #578 / #665


oberstet:
Cool;) The comments are really good to have there .. this is a tricky one;)

PR #670, opened 2016-03-06
New feature: purge cookie file on startup

taueres:
A new option `purge_on_startup` has been added to the `cookie.store` configuration (default value is `False`).

### DIFF
--- a/crossbar/router/cookiestore.py
+++ b/crossbar/router/cookiestore.py
@@ -237,6 +237,9 @@ class CookieStoreFileBacked(CookieStore):
         # initialize cookie database
         self._init_store()
 
+        if config['store'].get('purge_on_startup', False):
+            self._clean_cookie_file()
+
     def _iter_persisted(self):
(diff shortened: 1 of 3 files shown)


crossbar/router/cookiestore.py
```diff
@@ -292,3 +295,19 @@ def setAuth(self, cbtid, authid, authrole, authmethod):
...
+                    'created': cookie['created'],
+                    'max_age': cookie['max_age'],
+                    'authid': cookie['authid'],
+                    'authrole': cookie['authrole'],
+                    'authmethod': cookie['authmethod']
+                }) + '\n'
+                cookie_file.write(cookie_record)
+
+            cookie_file.flush()
+            os.fsync(cookie_file.fileno())
```
meejah: This probably isn't necessary; should flush on close() which will happen when the ``with`` block closes. But, probably no harm either...

Otherwise, LGTM +1

oberstet:
@taueres Awesome, thanks!!

PR #675, opened 2016-03-14
Skip test if running as root

taueres:
I usually execute Crossbar.io tests via Docker. Since Crossbar.io runs as root the `test_unix_already_listening_cant_delete` test always fail.
I propose to skip it if running as root.

### DIFF
--- a/crossbar/twisted/test/test_endpoint.py
+++ b/crossbar/twisted/test/test_endpoint.py
@@ -142,3 +142,7 @@ class ListeningEndpointTests(TestCase):
         test_unix_already_listening.skip = _
         test_unix_already_listening_cant_delete.skip = _
         del _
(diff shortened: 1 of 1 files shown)


meejah:
LGTM, +1

oberstet:
It does not work on Windows: the patch needs to check for existence of `os.getuid` or sth ..

> Note: In general, CB (like any other daemon) should NOT be run as root. Never. Not on bare OS, not in OS containers.

taueres:
@oberstet Windows platform is already handled at line [139](https://github.com/crossbario/crossbar/pull/675/files#diff-59fb7ce49c5af3233b27f9b069604b9dR139)

oberstet:
@taueres ah, right;)

PR #677, opened 2016-03-16
Migrated conditional requirements to extras_require

agronholm:
This will enable the creation of wheels with conditional requirements.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,16 +30,11 @@
 
 from __future__ import absolute_import
 
-import sys
 import re
 import os
 import platform
+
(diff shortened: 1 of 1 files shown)


oberstet:
I believe the equivalent of this PR in slightly modified form was already merged. So closing this one ...

agronholm:
This PR is a vastly simpler version of the one that was merged. No duplication. I didn't recommend merging this because it still had the problem with the `platform_python_implementation` environment marker not being recognized by older versions of setuptools. The merged PR simply ignores this problem.

Issue #543, opened 2015-12-10
Use pegged dependencies


oberstet:
Here is a suggestion (by @meejah), which should avoid issues like https://github.com/crossbario/crossbar/issues/542 in the future:

> also (and I guess @hawkowl would know best) but I think the usual advice is "open-ended requirements for library like autobahn>=0.10.0, but pegged requirements for programs like autobahn==0.10.2
​> that is, it's ​probably​ best to have crossbar declare exact dependencies, so that autobahn (or txaio) can be upgraded + released w/o affecting crossbar

oberstet:
It makes sense to have that distinction between "libraries" and "programs". And yes, Crossbar.io is a program, in particular when installed from our binary packages (the recommended way).

The only thing to remember: when a user install Crossbar.io into a shared Python environment, and/or wants to host AB Py app components in routers / containers, the AB version pinning will extend to those components as well. But as the user components then will run along CB code, it's probably better to pin anyway.

So I am +1. But @hawkowl will have more experience here I guess.

meejah:
Absolutely I defer to @hawkowl's experience ;) but I suspect the best advice is for people releasing a component to treat it like a library (open-ended requirements). The problems arise when more than 1 thing pins to "a" version (and those versions differ). So, if crossbar.io is pinning, components shouldn't (or, if they must pin they should pin to a crossbar version, not e.g. an autobahn version).

hawkowl:
That sounds good to me :)

oberstet:
Actually, I think we should go further, and not only pin CB's direct dependencies, but all indirect deps as well.

Would it be enough if we'd check in the output of pip freeze from a fresh installation of CB into a fresh virtualenv? Mmh. I guess platform dependent, dynamic imports make that tedious. The requirements file would need to be the superset over all platforms, and have the platform specific stuff guarded in conditionals.

I am wondering about this general question: is there a way to make sure that when I do "pip install X" today and in 2 weeks, that I do get the same package versions of everything? A repeatable install.

waterbug:
I suggest checking out the use of "conda" as a package manager:
http://conda.pydata.org/docs/index.html

It has many advantages over pip, being capable of installing
specified versions of python, set up its own virtualenvs, and
keep all library dependencies isolated from system libraries.

One of conda's features, relevant here, is its capability to
"pin" specific versions of packages in an environment -- see:
https://www.continuum.io/blog/developer/advanced-features-conda-part-1

[...]

PR #678, opened 2016-03-17
Migrated conditional dependencies to PEP 508 format

agronholm:
This uses a much more recent (and better) format for conditional dependencies than the previous PR.
I had to add ez_setup.py because setup.py now requires the latest setuptools to work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,16 +30,14 @@
 
 from __future__ import absolute_import
 
-import sys
 import re
 import os
 import platform
+
+import ez_setup
+ez_setup.use_setuptools('20.3')
+
 from setuptools import setup, find_packages
-
-version = getattr(sys, "version_info", (0,))
-
-CPY = platform.python_implementation() == 'CPython'
-PYPY = platform.python_implementation() == 'PyPy'
 
 LONGSDESC = open('README.rst').read()
 
@@ -55,7 +53,7 @@ else:
     raise RuntimeError("Unable to find version string in {}.".format(VERSIONFILE))
 
 # enforce use of CFFI for LMDB
-if PYPY:
+if platform.python_implementation() == 'PyPy':
     os.environ['LMDB_FORCE_CFFI'] = '1'
 
 # enforce use of bundled libsodium
@@ -95,27 +93,19 @@ install_requires = [
 
     # HTTP/REST bridge (also pulls in TLS packages!)
     'treq>=15.1.0',               # MIT license
-]
-if sys.platform != 'win32':
+
     # setproctitle does not provide wheels (https://github.com/dvarrazzo/py-setproctitle/issues/47) => disable on Windows
-    install_requires.append('setproctitle>=1.1.9')  # BSD license
+    'setproctitle>=1.1.9; sys_platform != "win32"',  # BSD license
+    'pypiwin32>=219; sys_platform == "win32"',       # PSF license
 
-if sys.platform == 'win32':
-    install_requires.append('pypiwin32>=219')       # PSF license
+    'pyinotify>=0.9.6; "linux" in sys_platform',     # MIT license
 
-# FIXME: https://github.com/crossbario/crossbar/issues/581
-if sys.platform.startswith('linux'):
-    install_requires.append('pyinotify>=0.9.6')     # MIT license
-
-# native WebSocket/JSON acceleration - only for CPy (skip for PyPy, since it'll be _slower_ on that!)
-if CPY:
     # wsaccel does not provide wheels (https://github.com/methane/wsaccel/issues/12) => disable on Windows
-    if sys.platform != 'win32':
-        install_requires.append('wsaccel>=0.6.2')   # Apache 2.0
+    'wsaccel>=0.6.2; sys_platform != "win32" and platform_python_implementation == "CPython"',  # Apache 2.0
 
     # ujson is broken on Windows (https://github.com/esnme/ultrajson/issues/184)
-    if sys.platform != 'win32':
-        install_requires.append("ujson>=1.33")      # BSD license
+    'ujson>=1.33; sys_platform != "win32" and platform_python_implementation == "CPython"',     # BSD license
+]
 
 # For Crossbar.io development
 extras_require_dev = [
@@ -123,20 +113,18 @@ extras_require_dev = [
     'colorama>=0.3.3',              # BSD license
     'mock>=1.3.0',                  # BSD license
     'wheel>=0.26.0',                # MIT license
-]
-if sys.platform != 'win32':
+
     # Twisted manhole support
     # pycrypto does not provide wheels => disable on Windows
-    extras_require_dev.append('pycrypto>=2.6.1')   # Public Domain license
+    'pycrypto>=2.6.1; sys_platform != "win32"',  # Public Domain license
+]
 
 # Crossbar.io/PostgreSQL integration
 extras_require_postgres = [
-    'txpostgres>=1.4.0'             # MIT license
+    'txpostgres>=1.4.0',            # MIT license
+    'psycopg2>=2.6.1; platform_python_implementation == "CPython"',  # LGPL license
+    'psycopg2cffi>=2.7.2; platform_python_implementation != "CPython"',  # LGPL license
 ]
-if CPY:
-    extras_require_postgres.append('psycopg2>=2.6.1')       # LGPL license
-else:
-    extras_require_postgres.append('psycopg2cffi>=2.7.2')   # LGPL license
 
 # Crossbar.io/Oracle integration
 extras_require_oracle = [
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 include COPYRIGHT
 include LICENSE
+include ez_setup.py
 recursive-include crossbar/web *
 recursive-include crossbar/templates *
 recursive-include crossbar/adapter/postgres/ddl *
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -1,0 +1,417 @@
+#!/usr/bin/env python
+
+"""
+Setuptools bootstrapping installer.
+
+Run this script to install or upgrade setuptools.
+"""
+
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -30,16 +30,14 @@
...
 
-import sys
 import re
 import os
 import platform
-from setuptools import setup, find_packages
 
-version = getattr(sys, "version_info", (0,))
+import ez_setup
+ez_setup.use_setuptools('20.3')
```
meejah: Is it possible to just do "setup_requires=['setuptools>=20.3']" or so instead of including ez_setup...?

agronholm:
Sigh...looks like pip hasn't vendored a recent enough `packaging` version yet :(

setup.py
agronholm: No. If you do that, the invocation of setup() will fail due to the requirements parse error before setuptools has a chance to update itself.

agronholm:
Scratch that, 8.1.0 should have a good version vendored. Investigating further.

agronholm:
This smells like a number of bugs in pip.

oberstet:
Seems like Travis has pip 6.0.7 =(

https://travis-ci.org/crossbario/crossbar/jobs/116565855#L139

agronholm:
I don't like it but I have to admit that this PR should probably be deferred until after the supporting tools are ready.

oberstet:
@agronholm thanks for the CAA! got it.

>  this PR should probably be deferred until after the supporting tools are ready.

Agreed.

Link: the PR addresses https://github.com/crossbario/crossbar/issues/525

agronholm:
This PR should fix the pip bug I'm seeing: https://github.com/pypa/pip/pull/3307

oberstet:
@agronholm Travis is failing probably because of pip too old. What if you add a `pip install -U pip` here https://github.com/crossbario/crossbar/blob/master/.travis.yml#L5 ?

agronholm:
@oberstet it won't help because even the latest pip will fail when doing `pip install .[crossbar]` because parts of pip still don't support PEP 508 requirements. The PR I linked to should fix that. And even when that happens, upgrading pip manually in the Travis build will only fix the Travis build. People with older pip versions will still have trouble installing the wheels. 

My (previous PR)[https://github.com/crossbario/crossbar/pull/677] _almost_ works, but the wheel installation fails on Travis because the name of the `platform_python_implementation` environment marker used to be called `python_implementation` but in packaging 16.5 (used by setuptools >= 20.2) it was renamed to its proper name. So if I use the new spec compliant name, the wheels will install using recent versions of pip but will introduce breakage for users of older pips (like the one used by Travis). On the other hand, if I use the older name, it will work only with older versions of pip/setuptools. So no matter what I do I lose :(

oberstet:
Oh, well. I see.

I guess we all agree that the PEP508 approach is better than what we do now in setup.py, and that we do want to have wheels.

But we can't have it now because of the issues you explained.

Mmh. So what do we do?

agronholm:
I'll try asking this on distutils-sig and see if they come up with something.

hawkowl:
nooooo don't use ez_setup :(

dstufft has mentioned that if we do it conventionally in the setup.py, and overload it in setup.cfg, then the wheels will operate correctly on older pips, at the cost of duplicating dependencies.

agronholm:
Can you point me to documentation on how to accomplish this? And how does this circumvent the problem with `platform_python_implementation` vs `python_implementation`?

hawkowl:
I am not sure; let me try something out.

agronholm:
According to my research, setuptools 18.5 is the minimum setuptools version that supports the `platform_python_implementation` marker. This is the minimum version needed to do either sdist or bdist_wheel with these markers present. However, pip versions older than that should support them in wheel metadata. So what is the oldest version of pip that crossbar needs to support?

It also seems that PEP 508 support is very incomplete in tooling -- the wheels created from this branch are missing the conditionals entirely. My other PR won't pass checks on Travis because Travis has setuptools older than 18.5. The use of ez_setup would fix that. @hawkowl 's PR is, in my opinion, a hack that should be avoided as the extra maintenance burden is not worth the very modest speed boost gained on installation speed, at least in case of a pure Python project like Crossbar.

oberstet:
Using ez_setup ... nooo.

agronholm:
Can you elaborate on why using ez_setup is bad? It's not like it replaces your current setuptools installation. It just conditionally downloads a newer one to the current directory and imports setuptools from that.

oberstet:
ez_setup is bad, because Python packaging/setup stuff is already way too fucked up, and mixing in yet another thing (ez_setup) IMO will make things worth.

agronholm:
The only purpose of ez_setup is to ensure that a new enough setuptools is used to run setup.py. It's transparent to the user and causes no action to be taken if the already installed setuptools is recent enough.

agronholm:
I'd like to add that most users will never even hit ez_setup because they'll be installing the wheel.

oberstet:
@agronholm I know how ez_setup works, and I don't like it;) A user should have pip installed, and a recent one, and the setup of a package (like Crossbar.io) should NOT fiddle with my setuptools/pip installation while installing the pacakge.

I think it's not worth for now fiddling with old pip and what not just to make a Crossbar.io wheel, which won't cut it anyway: on Linux, there are no binary wheels at all, and CB nowerdays has multiple hard dependencies on binary extensions (NaCl, LMDB, ..).

In other words: for now, you must have a C toolchain to install Crossbar.io from pip. You don't need a toolchain when you install Crossbar.io from our binary packages (deb/rpm, ..). And the latter is the recommended way to install it.

What was your original, concrete itch again?

agronholm:
Pardon me, but from your reply it seems that my intentions were completely misunderstood. The reason for using ez_setup is to ensure the availability of a recent enough setuptools version for the PEP 508 requirement specifiers to be correctly processed when crossbar is installed from source (and not from wheel). For example, if you install directly from git, you would need a recent enough setuptools for the install to succeed if PEP 508 `install_requires` are to be supported. If ez_setup is not used, the user is expected to understand how to deal with the cryptic error message and upgrade their setuptools installation. If install failing on older setuptools versions is acceptable then ez_setup is not needed.

What are the minimum setuptools and pip versions expected from users of Crossbar anyway?

oberstet:
@agronholm Let's just say the minimum versions are the current ones:

```console
(python279_1)oberstet@thinkpad-t430s:~/scm/crossbario/crossbar$ python
Python 2.7.9 (default, Jan 18 2015, 21:40:29) 
[GCC 4.8.2] on linux2
>>> import pip
>>> pip.__version__
'8.1.1'
>>> import setuptools
>>> setuptools.__version__
'20.6.7'
```

I will check if I can make Travis update setuptools/pip before installing CB. If so, I am gonna remove the `OLD_SETUPTOOLS` stuff in `setup.py` and make above versions the officially supported minimal ones.

oberstet:
@agronholm Another note: soonish, we will **pin** all Crossbar.io dependencies to exact point releases - and that will include setuptools.

Crossbar.io is supposed to be installed either into a virtualenv of its own - or installed via binary packages which bundle everything including Python and the kitchen sink anyway.

oberstet:
@agronholm Another note: soonish, we will **pin** all Crossbar.io dependencies to exact point releases - and that will include setuptools.

Crossbar.io is supposed to be installed either into a virtualenv of its own - or installed via binary packages which bundle everything including Python and the kitchen sink anyway.

oberstet:
see here https://github.com/crossbario/crossbar/issues/543 - we will pin all direct as well as indirect dependencies

agronholm:
Just so you're aware: pinning setuptools won't help with the PEP 508 requirements problem because setuptools parses the install requirements *before* it tries to process the setup dependencies. Stupid design, yes, but that's the reality we live in.

Pinning dependencies exactly in Crossbar will cause me (and possibly other users) a *major* headache. My Asphalt framework has a testing dependency on Crossbar and if it suddenly starts pinning dependencies to exact versions, I will be unable to upgrade mine.

oberstet:
Yes, it won't help for setuptools itself. But I figured that our plans for pinning would be of interest to you in this context. But having Crossbar.io as a Python level dependency is wrong. Cb should be treated as a blackbox, much like Nginx. Imagine I would find time and reasons some day, and I'd rewrite CB in C++ or Rust. There is no way you can have a Python level dep to a Rust written thing ..

Could you describe your Asphalt framework, and how/why it depends on CB at the Python level?

agronholm:
Because I do integration testing with it in the asphalt-wamp subproject. See this function here: https://github.com/asphalt-framework/asphalt-wamp/blob/master/asphalt/wamp/utils.py#L40
I launch an ad-hoc Crossbar instance and run all integration tests through it. You're right in that if it wasn't written in Python it'd be much more of a PITA to do this.

meejah:
@agronholm For my integration testing things I'm doing something very similar (although with py.test, and fixtures) and simply have the "crossbar" fixture rely on a "virtualenv" fixture, so that the "crossbar instance under test" is in its own virtualenv.

For sure slightly more pain than just a spawn in the same ("parent"?) virtualenv, but also nice separation and does actually treat the "crossbar" process more as a black-box (i.e. the same fixture would work if it was calling a compiled program; just the 'install' part becomes something besides 'make a new virtualenv'). This has an additional advantage: the 'parent' testing framework can (and does) use Autobahn libraries, but it can still test e.g. a released version of Autobahn while using a "master" or branch version itself...

agronholm:
Ok @meejah, any chance you could share those fixtures?

oberstet:
@agronholm This functional/integration test infrastructure for Crossbar.io itself is not open-source.

meejah:
@agronholm Those particular fixtures are in a private repository :( but they're pretty similar to the code you pointed to -- just instead of ``sys.executable`` it's ``/path/to/venv/bin/crossbar``

oberstet:
I have a PR that seems to work on Travis too: https://github.com/crossbario/crossbar/pull/707

Closing this one ..

PR #680, opened 2016-03-22
Realm templates

oberstet:
This implements realm templates. Realm templates have their name in the node configuration be a regular expression. When the authentication has assigned a realm and that realm is not running, the automatic realm activation will match the realm against all patterns, and upon a match, start the realm from the configuration template under the actual name.

Works with the example here https://github.com/crossbario/crossbarexamples/blob/master/realmtemplates/.crossbar/config.json#L34

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -479,16 +479,19 @@ class Node(object):
             self.log.info("Using default node shutdown triggers {}".format(self._node_shutdown_triggers))
 
         # router and factory that creates router sessions
-        #
         self._router_factory = RouterFactory(self)
         self._router_session_factory = RouterSessionFactory(self._router_factory)
 
         rlm_config = {
+            # local node management realm (by default, "crossbar")
             'name': self._realm
         }
         rlm = RouterRealm(None, rlm_config)
 
-        # create a new router for the realm
+        # create a new router for the node management realm
+        #
+        # Note: the router factory by default only has a single role "trusted",
+        # stored in self._router_factory._routers[self._realm]._roles
         router = self._router_factory.start_realm(rlm)
 
         # add a router/realm service session
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -40,7 +40,7 @@ from twisted.internet.defer import inlineCallbacks, Deferred
 from twisted.internet.ssl import optionsForClientTLS
 
 from autobahn.util import utcnow
-from autobahn.wamp.types import CallDetails, CallOptions, ComponentConfig
+from autobahn.wamp.types import CallDetails, ComponentConfig
 from autobahn.wamp.exception import ApplicationError
 from autobahn.twisted.wamp import ApplicationRunner
 
@@ -199,6 +199,9 @@ class Node(object):
 
         # map fro router worker IDs to
         self._realm_templates = {}
+
+        # fake call details we use to call into the local node management API
+        self._call_details = CallDetails(caller=0)
 
         # for node elements started under specific IDs, and where
         # the node configuration does not specify an ID, use a generic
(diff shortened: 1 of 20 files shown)


meejah:
Interesting approach with the regular-expressions! I tried this branch with my auto-realms test-case and I still get some odd failures (i.e. doesn't work at all) that I haven't figured out..some assertion from inside maybeDeferred() when calling ``.hello()`` on the pending auths.

The way I was thinking of having auto_realm work is by accepting either True/False for "auto_realm=" in the config, **or** accepting a whole realm config (minus the name) there -- and in that case that would be the realm-config for any auto-started realms. Not sure which is less confusing ;) I guess it depends on your reg-ex fu ;)

I'm pretty sure you'll have the race-condition where if you create two "auto-started" realms at "the same" time such that the auto-start from the first connection isn't finished when you get the second connection you'll be trying to auto-start the same-named realm twice (which will fail, and be wrong)..

I was also stumbling on how to get configuration to the workers (e.g. adding a "--auto-realm" option smelled bad) -- perhaps we want/need a more-general way to pass configuration down to the different worker types? (It doesn't look like this branch does that at all, right? That is, there's no "auto_realm" option etc?)

oberstet:
The idea with the regex stuff also is to allow to define catching, named groups in the regex, which a parsing out pieces from eg the realm being activated, and then the carved out piece can be used in the template, like eg in URIs.

Yeah, there will be race conditions I guess. The basic case seems to work fine .. the first client connecting to a realm matching will make the node controller create the realm (and the roles). And only then the client is connected. It's transparent even for the very first client - it only takes slightly longer.

Now, above PR is about **realm templates** and **auto-activation of realms**. What's probably missing is **role templates** and **auto-activation of roles**.

What's also missing is a way to automatically start workers, like containers with components, whenever a realm is started dynamically. Since up until now, we could manually define what containers and components to start, as the set of realms started was fixed too. But since the latter isn't any longer the case, we need a way to **dynamically start workers**, at least containers and guests too.

I am still thinking of how to do the latter best. Currently, I am leaning towards having a new attribute in a container/guest configuration item:

```json
{
    "type": "container",
    "template": "realm",
    "realm": "^com.crossbario.cdc.mrealm-(?P<mrealm>[a-z][a-z0-9_-]{2,8})$"
}
``` 

When a matching realm is (dynamically) started, a new container worker is started (with the "realm" parameter replaced by the actual value). The container worker could run a component that connects back to a router under the respective realm. In result, whenever the realm is started, the backend component is started (in a separate container worker) too.

In a way, automatic realm templating and activation together with container/guest worker auto-start can provide a simple way of dynamically activating app services ... at least that was the idea here.

oberstet:
I've moved the example and attached complete log output here https://github.com/crossbario/crossbarexamples/tree/master/templates/realm-templates#how-to-try

oberstet:
A key to the code is: while the router has the need for a realm not running to be started, it doesn't start it directly. It asks the node controller to activate the realm, which in turn calls into the router to actually start the realm. Why that? Because that way, we still have one code path through which realms are started in a node - via the central code in the node controller that reads a config and orchestrates the startup of a realm by calling into the management API exposed by the router workers.

meejah:
At least one way to stop the race-condition is to keep a list of "realms that are starting" (with their Deferreds) so that on the second "please start X" you can look it up in that list and return the existing Deferred...

oberstet:
@meejah ah, thats a good idea! this minimizes the race condition. as soon as the node controller is asked to activate a realm, it creates and stores a deferred that fires when it has finished doing so. When another request to activate the realm comes in concurrently, it is just handed out the existing deferred. As long as the node controller isn't interrupted between receiving the request to activate and the storage of the deferred, that should work.

waterbug:
Yeah, I suspect that "interrupted" scenario is equivalent to the
"Two Generals' Problem" (of two-phase commit fame):
https://en.wikipedia.org/wiki/Two_Generals'_Problem
so, IOW, this approach is probably the best you can do.  ;)

meejah:
An idiom like:

```
d = self._starting[realm_name] = defer.succeed(None)
# ...
d.addCallback(lambda _: self._start_realm(realm_name))
# ...
return d
```
...should ensure that you're storing "the" Deferred before you even issue the request to start the realm. Might still be a window for the generals to mis-communicate that I'm missing ;)

oberstet:
This last commit adds support for dynamically starting containers when realms are started (dynamically). Here is an example https://github.com/crossbario/crossbarexamples/tree/master/templates/worker-templates

crossbar/controller/node.py
```diff
@@ -531,231 +550,267 @@ def start(self, cdc_mode=False):
...
-                        yield self._controller.call('crossbar.node.{}.worker.{}.start_container_component'.format(self._node_id, worker_id), component_id, component, options=call_options)
-                        self.log.info("{worker}: component '{component_id}' started",
-                                      worker=worker_logname, component_id=component_id)
+            yield self._controller.call('crossbar.node.{}.worker.{}.start_router_transport'.format(self._node_id, worker_id), transport_id, transport)
+            self.log.info("{}: transport '{}' started".format(worker_logname, transport_id))
 
-                    # after 2 seconds, consider all the application components running
-                    self._reactor.callLater(2, component_stop_sub.unsubscribe)
+    @inlineCallbacks
+    def _run_container_config(self, worker, worker_id, worker_logname, worker_options):
```
meejah: refactoring these to methods is really nice :)

oberstet:
@meejah a handful of unit tests fail https://travis-ci.org/crossbario/crossbar/jobs/118307039#L475 - any idea?

meejah:
Hmm, probably because it's logging those errors "properly" now, we have to tell trial they're expected; I'll investigate...

PR #684, opened 2016-03-23
Make universal wheels work

hawkowl:
merged forward + fixed version of #678 by @agronholm

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -153,7 +153,7 @@ else:
         'pyinotify>=' + pyinotify_version
     ]
 
-    extras_require['sys_platform != "win32" and platform_python_implementation == "CPython"'] = [
+    extras_require[':sys_platform != "win32" and platform_python_implementation == "CPython"'] = [
         'wsaccel>=' + wsaccel_version, 'ujson>=' + ujson_version
     ]
 
--- a/setup.py
+++ b/setup.py
@@ -30,16 +30,22 @@
 
 from __future__ import absolute_import
 
-import sys
 import re
 import os
+import sys
 import platform
-from setuptools import setup, find_packages
 
-version = getattr(sys, "version_info", (0,))
+from setuptools import setup, find_packages, __version__ as setuptools_ver
 
-CPY = platform.python_implementation() == 'CPython'
-PYPY = platform.python_implementation() == 'PyPy'
+if setuptools_ver < "20.0":
+    OLD_SETUPTOOLS = True
+
+    if "bdist_wheel" in sys.argv:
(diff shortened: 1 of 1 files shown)


hawkowl:
I think that this should work. I've done some tests and it *appears* to function fine on older setuptools + newer ones, without *too* much duplication.

agronholm:
Um, so you're saying wheels created with this also install on older pip versions? Because that was my only concern with my other PR.

agronholm:
Are you sure you've seen my alternative PR? https://github.com/crossbario/crossbar/pull/677

hawkowl:
Old pips won't download wheels from PyPI by default -- so, it'll break if you manually install a wheel, but that's not really something you should do...

agronholm:
How old pip versions are we talking about?

hawkowl:
```
Python 3.4.3+ (default, Oct 14 2015, 16:03:50) 
[GCC 5.2.1 20151010] on linux
>>> import pip
>>> pip.__version__
'1.5.6'
>>> import setuptools
>>> setuptools.__version__
'18.4'
>>> import wheel
>>> wheel.__version__
'0.26.0'
```

agronholm:
I just tried pip 1.5.4 and it downloads wheels by default. Ubuntu Trusty ships with setuptools 0.7.2 too.

hawkowl:
I installed those two specific versions, and it appears to work fine (on cpy, it installs wsaccel, and pypy it does not).

agronholm:
I will take a better look at this once I'm properly rested.

setup.py
```diff
@@ -95,27 +106,56 @@
...
-    # setproctitle does not provide wheels (https://github.com/dvarrazzo/py-setproctitle/issues/47) => disable on Windows
-    install_requires.append('setproctitle>=1.1.9')  # BSD license
 
-if sys.platform == 'win32':
-    install_requires.append('pypiwin32>=219')       # PSF license
+setproctitle_version = "1.1.9"  # BSD license
+pypiwin32_version = "219"  # PSF license
+pyinotify_version = "0.9.6"  # MIT license
+wsaccel_version = "0.6.2"  # Apache 2.0
+ujson_version = "1.33"  # BSD license
```
agronholm: Why are the versions separate from the requirement lines? They weren't so in the original nor in my PRs and I don't see why they should be.

hawkowl: So that when I do a version bump, I don't accidentally require two different versions depending on if your setuptools version :P

agronholm: Good point.

PR #707, opened 2016-04-08
update pip, setuptools, wheel on Travis before running tests

oberstet:


agronholm:
It's your project and I have no say over this but my PR (https://github.com/crossbario/crossbar/pull/677) had a substantially smaller and simpler setup.py and upgrading pip+setuptools like you did in this PR would have worked there just as well.

oberstet:
@agronholm I do appreciate your input and work! It's really not about not wanting to merge your code. It's just in this particular case where we need to collect some loose ends, and iterate a bit, because: this PR isn't finished .. the goal is to get rid of package versions in setup.py altogether and move to requirements.txt and pinned dependencies (direct and indirect)

agronholm:
I am then correct in understanding that pip installing crossbar is going to be deprecated? If so, it makes sense to completely get rid of setup.py. I have done that on a couple occasions. Or do you simply want to read the requirements to setup.py from an external file?

oberstet:
@agronholm: IOW, right now, your PR is essentially that same as this, but without the pip upgrade for Travis (and removal of DB integration, to get rid of stuff) - which you mention you could have added as well. It's just: **I** was under the (false) impression that wouldn't work .. but it does. So the OLD_SETUPTOOLS workaround implemented by @hawkowl (which was, not ultra pretty I admit) isn't required anymore. Because, I only care about Travis - not old pip/setuptools around the world ..

---

"pip install crossbar" must continue to work, but ideally, I'd like to have all package versions in (and only in) requirements.txt.

Not the least to be able to do this hash checking thing @meejah mentioned https://github.com/crossbario/crossbar/issues/543#issuecomment-203725353

agronholm:
If `pip install crossbar` must continue to work, then the dependencies must be available to the setup() call somehow. I'm having a bit of trouble imagining how the conditional dependencies would work in such an arrangement.

oberstet:
The setup.py continues to serve as a holder of package metadata. when the setup.py runs from direct execution or from pip, it will - depending on platform - load one of multiple requirements.txt files. It "parses" the file by simply splitting at newlines and whitespace stripping. The resulting list of strings is used in the setup() function.

By doing so, exactly the set of package fitting for platform and pinned to version is installed.

I think this could work ...

oberstet:
Well, this or using the platform or python flavor conditions, and get away with just a single requirements.txt

agronholm:
So you'd have separate requirement files for win32-cpython, win32-pypy, darwin-cpython, darwin-pypy, linux-cpython, linux-pypy (and what about freebsd et al) ? What about the conditional extras?

agronholm:
The only way to do this in a single requirements.txt is to use PEP 508 specifiers.

agronholm:
I just don't quite grasp what value the requirements.txt adds here if you're loading it in setup.py anyway.

oberstet:
@agronholm the point is, that when I do `pip install crossbar==x.y.z`, what you get currently is a moving target, because of 2 reasons:

1. dependencies (direct and indirect) are not pinned to exact versions
2. you have to hope that PyPI wasn't compromised, because lack of signatures

What I'd like to accomplish is: given above command, you should _always_ get exactly the same _bits_ installed. (if you just do `pip install crossbar`, you would get the "latest" of course .. a moving target - but still, within a CB release, everything pinned).

Let's call it: **repeatable, verified builds/installation**

I am still experimenting with this (see above). I am not sure if the current PyPI and Python infra is enough to accomplish that ..

oberstet:
eg here is what you get when hashes fail:

```console
(cpy2711_1) oberstet@corei7ub1310:~/scm/crossbario/crossbar$ pip install --ignore-installed --no-cache-dir --require-hashes -r requirements2.txt
Collecting click==6.6 (from -r requirements2.txt (line 1))
  Downloading click-6.6.tar.gz (283kB)
    100% |████████████████████████████████| 286kB 2.1MB/s 
THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    click==6.6 from https://pypi.python.org/packages/source/c/click/click-6.6.tar.gz#md5=d0b09582123605220ad6977175f3e51d (from -r requirements2.txt (line 1)):
        Expected sha256 cc6a19da8ebff6e7074f731447ef7e112bd23adf3de5c597cf9989f2fd8defe8
             Got        cc6a19da8ebff6e7074f731447ef7e112bd23adf3de5c597cf9989f2fd8defe9
```

agronholm:
Yes, I understand what your goals are. Hash checking makes sense, but how will you specify conditional dependencies in a requirements.txt?

meejah:
"dstufft" says on #python that it should be possible to put conditional-deps in a requirements.txt (but he confirms it's *not* possible to do hash-checking in the ``requires=`` lists in setup.py)

agronholm:
Well, that was basically what I was asking about :) So PEP 508 it is.

agronholm:
Whatever loads the requirements.txt in setup.py will then have to erase the hashes.

meejah:
@agronholm yup, it sounds like the only path forward for "want hash checks" is via a requirements file :/

I have also done hacks in other projects that read the requirements.txt in setup.py so there's "just one" place to list the deps, but that trick presumably won't work for setup.py's that are to become wheels...?

Pragmatically, it might be best for now to have the "requirements.txt" with pinned versions and hashes as the "expert" way to install while leaving setup.py workable with wheels etc and so that releases via PyPI continue to work...at least then internal CB release process can use the requirements.txt for repeatability etc when building binaries.

meejah:
@agronholm yes, you beat me to it ;) if we loaded requirements.txt in setup.py we'd have to parse out the ``--hash=sha256:...`` bits.

oberstet:
@agronholm yes, conditional deps via pep508 specifiers: https://github.com/oberstet/crossbar/blob/new_pip_setuptools/requirements-in.txt#L28

This works with `pip install -r requirements-in.txt` **and** when used from `setup.py` (with a little tweaking): https://github.com/oberstet/crossbar/blob/new_pip_setuptools/setup.py#L55

So, @meejah , that is what Donald says. And yep, I was wondering how to make hashes work with that too next -- but he also says it doesn't work anyway - so I stop looking =(

And then: all this hashing stuff is a step forward for sure, but it's not the real deal IMO.

I don't want to check hashes (against known goods I get .. from where?), but check proper cryptographic **signatures** that have been created by **upstream maintainers** themselves. Essentially, similar to what eg Debian deb/apt-get tooling with GPG signed binary packages provides.

Well, its simply not there I guess in the Python world.

> Pragmatically, it might be best for now to have the "requirements.txt" with pinned versions and hashes as the "expert" way to install

Yes

> while leaving setup.py workable with wheels etc and so that releases via PyPI continue to work

The setup.py should use pinned versions (with pep508 conditionals) too. But it can't do hashes.

So we need 2 requirements.txt files .. with and without hashes. Or we parse out the hashes. And then pip doesn't have a public API for it's requirememts parsing. Python packaging is fun! ;)

Relevant resources (new for me at least):

* https://caremad.io/2013/07/setup-vs-requirement/
* http://www.peterbe.com/plog/bestest-python-dependencies

meejah:
Rgd "real" crypto-signatures: yes, they're there and *possible* but not really "part of" the ecosystem entirely. So, for example, you can download an sdist of txtorcon (**and/or** a wheel) which has a proper detached GPG signature. ``pip`` certainly doesn't include a way to check these, however, so you have to download the signature and verify "by hand" -- but at least PyPI provides a way to distribute the signature alongside.

The workflow I use to produce these is documented in https://github.com/meejah/txtorcon/blob/master/docs/release-checklist.rst

To upload them, I use ``twine`` (https://github.com/pypa/twine) as this allows you to *separately* build the tarball and wheel, test them, and *then* upload (whereas the setuptools method basically builds, signs and uploads in one step -- so if it was wrong, you have to delete, increment version and try again etc). Even without the signatures, I think using twine for uploads is way better. (And *with* signatures, it lets you produce the signatures on a different machine and use gpg or whatever yourself instead of hoping setuptools is calling it correctly).

All the hash-checking gets you is to confirm that you're getting the same dependency as the maintainer saw -- i.e. for crossbar, it would let installers know that they're getting the very same dependency that you saw (and did "whatever" to confirm) as you intended. Certainly not a replacement for a proper cryptographic signature from the upstream provider. Still, it means that e.g. you as crossbar maintainer can at least "pass along" whatever due-diligence you did when confirming dependencies.

You can read about where "pip" is at with verifying signatures here: https://github.com/pypa/pip/issues/1035 (the main issue/discussion seems to be "how do we know which keys are allowed/supposed-to sign which packages). And really the state-of-the-art here is https://theupdateframework.github.io/ -- which grew out of Tor's "thandy" updating discussions.

oberstet:
@meejah Thanks a lot! What you write above is interesting and there are gems in it that are completely new to me. The release process of `txtorcon` seems sophisticated.

> FWIW, after more fiddling and reading (and resulting confusion), I've decided to take a step back and first solve an easier one https://github.com/crossbario/txaio/issues/63, learn there, and then come back to this issue here. Unrelated: I will also now force myself into the proper "issue, feature branch, PR" workflow;) No more direct pushing into master. I am lazy, yeah, but now its time to get some discipline.

PR #708, opened 2016-04-08
Split node keys into .pub and .priv files

meejah:
Also refactor the loading etc. a bit so that we can more
easily read and write the two different files, and add
logic to confirm they're consistent (i.e. that the public
key really corresponds to the private one).

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -34,6 +34,7 @@ import os
 import traceback
 import socket
 import getpass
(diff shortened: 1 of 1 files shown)


oberstet:
I would merge, when flake8 is happy, as I turned the switch on the repo to require all CI to pass before being able to merge;)

meejah:
No worries, I like having checks turned on -- even if it means more "fix flake8" commits ;)

Issue #742, opened 2016-04-28
Add recipe deploying CB with Puppet


oberstet:
Natale has posted hints and bits for deploying Crossbar.io with [Puppet](https://puppet.com/product/how-it-works) on the mailing list [here](https://groups.google.com/forum/#!topic/crossbario/E1_dwxUnQxQ).

This would be a nice addition to the docs for users deploying their stuff via Puppet.

Suggested title: "Deploying Crossbar.io with Puppet".

blues-man:
Hi,
I put those files on a git repo in order to retrieve them easily, manage changes etc
https://github.com/blues-man/crossbar-puppet

and then better starting writing a module :)

oberstet:
If you want this to be included in Crossbar.io docs, could you please do a PR? There is no reason for having a bunch of files in a separate repo, plus the license is unclear. 

> Also, you should not include our GPG public key verbatim in the recipe, but add an instruction in Puppet that retrieves the key from key servers based on key fingerprint.

blues-man:
Sure, I'll do with those changes

PR #765, opened 2016-05-01
Add Deploying section and Puppet recipe #742

blues-man:
We would add a specific section for Deploying Crossbar with most
used configuration management tools as Puppet, Ansible, Salt etc
filling it with a Puppet recipe

### DIFF
--- a/docs/pages/installation/Deploying-Crossbar.io-with-Puppet.md
+++ b/docs/pages/installation/Deploying-Crossbar.io-with-Puppet.md
@@ -1,0 +1,38 @@
+[Documentation](.) > [Installation](Installation) > [Deploying Crossbar.io](Deploying Crossbar.io) > Deploying Crossbar.io with Puppet
+
+# Install Crossbar.io from Puppetforge 
+
+With Puppet you can provision Crossbar.io WAMP Router on your systems with standalone agent, Puppet Server or Puppet Enteprise.
+
+We provide a module on [Puppetforge](https://forge.puppet.com/bluesman/crossbar).
+
+For an overview of all feature supported and more technical details, see the [GitHub repository](https://github.com/blues-man/crossbar-puppet).
+
+
+## Usage
+
+Installation:
+
+```console
+puppet module install bluesman-crossbar
+```
+
+To run:
+
+```console
+puppet apply -e 'include crossbar'
+```
+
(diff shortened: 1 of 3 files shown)


docs/pages/installation/Deploying-Crossbar.io-with-Puppet.md
```diff
@@ -0,0 +1,38 @@
...
+For an overview of all feature supported and more technical details, see the [GitHub repository](https://github.com/blues-man/crossbar-puppet).
+
+
+## Usage
+
+Installation:
+
+```console
+puppet module install bluesman-crossbar
+```
```
oberstet: No, we need this under our control

oberstet:
Thanks! I will have to try that first myself. Also, it is probably best to wait a little, since we are gonna restructure the installation docs and make it simple. That means, the recommended way of installing CB will be more prominent - this is via

1. first, our own binary packages
2. then, Docker
3. then, anything else is "advanced"

docs/pages/installation/Deploying-Crossbar.io-with-Puppet.md
blues-man: ah ok, fine

blues-man:
Ok sounds good, in the while I just update the module that installs crossbar supporting both CentOS 7.x and Ubuntu 14.04  official packages and repo :)

oberstet:
I had a quick look into your repo. It seems slightly more than trivial;) And since I don't use Puppet, and don't speak Ruby, this is a problem for me .. because what we say in the official docs MUST be proven, supported etc - we can't put stuff there we don't understand and hence cannot support.

Now, how about this: we add a page on the docs here for "community recipes" .. with a section for Puppet, which basically says "yeah, possible" .. please see here for a community repo that has it. And link to your repo.

oberstet:
@blues-man this https://github.com/blues-man/crossbar-puppet/blob/master/manifests/repo.pp#L15 is highly appreciated. It gives users confidence that they are using exactly the bits we publish .. thanks for that!

blues-man:
Eheh yes always better to query directly the key server :)

I didn't know where to exactly put Puppet recipes, but for this case and others, community recipes looks pretty fine!

oberstet:
@goeddea pls see above .. some "community recipe page" or similar .. where we can point at @blues-man 's repo for CB deployment under Puppet.

PR #773, opened 2016-05-03
Remove incorrect cert path checking

w1z2g3:
If we want to check the cert path, we have to check os.path.join(cbdir, cert).

Please see crossbartwisted/endpoint.py:

    cert_fname = os.path.abspath(os.path.join(cbdir, config['certificate']))

I cannot find a clean way to get it done correctly yet. So propose to remove the check for now.
I think checking cert file exist is optional, and currently the wrong checking is a bug.

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -746,19 +746,9 @@ def check_connecting_endpoint_tls(tls):
         if k not in ['ca_certificates', 'hostname', 'certificate', 'key']:
             raise InvalidConfigException("encountered unknown attribute '{}' in connecting endpoint TLS configuration".format(k))
 
-    for k in ['certificate', 'key']:
-        if k in tls and not os.path.exists(tls[k]):
-            raise InvalidConfigException(
-                "File '{}' for '{}' in TLS configuration "
-                "not found".format(tls[k], k)
-            )
-
     if 'ca_certificates' in tls:
         if not isinstance(tls['ca_certificates'], Sequence):
             raise InvalidConfigException("'ca_certificates' must be a list")
-        for fname in tls['ca_certificates']:
-            if not os.path.exists(fname):
-                raise InvalidConfigException("'ca_certificates' contains non-existant path '{}'".format(fname))
 
     for req_k in ['hostname']:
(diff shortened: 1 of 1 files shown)


oberstet:
I agree, there is "some issue" here. So I think we first need to decide if we want to check existence of cert files during checking the config. If we want to do that, we need to do it correctly (respecting node directory `cbdir`). If not, then we remove the check for file existence (as your PR does).

> Note: for contributions to be merged, be would need a CAA http://crossbar.io/docs/Contributing-to-the-project/#send-us-a-contributor-assignment-agreement

meejah:
Additionally: I can see usefulness of allowing full-paths (i.e. outside of crossbar's ``cbdir``) -- e.g. to "system" locations like ``/etc/ssl/certs`` so perhaps what we really want here is to *add* support for checking relative paths? That is so that you can put a relative **or** absolute path in the config.

I think no matter what, we might as well check the paths -- or do you want to just pass along (possibly bogus) paths into the Twisted APIs and let it fail on startup, if Twisted decides that's the correct course?

w1z2g3:
Hi @meejah :

>>> os.path.join('/cbdir', '/abs/cert')
'/abs/cert'

So we do support absolute path.

Without knowing the --cbdir, if we do want to do some check, we can do:

  if os.path.isabs(cert) and not os.path.exist(cert): rainse Exception

meejah:
Actually, okay I agree with @w1z2g3 -- we should just remove the checks, as if they're relative paths they'll succeed anyway (because later, when we're really running, we've ``chdir``'d to ``cbdir``). Currently, nothing in checkconfig has access to ``cbdir`` so we'd have to pass it all the way down from ... CLI level.

oberstet:
@w1z2g3 Thanks for the CAA! Could you please rebase your PR to master - then I merge.

w1z2g3:
A little trick to update a pull request. Not sure whether I need close this pull request and open a new one.

For now, I just rebased the patch in the same branch. Please let me know whether this pull request is good now.

meejah:
@w1z2g3 Updating "in place" like that is perfectly fine. Thanks! :)

PR #777, opened 2016-05-04
Invoke worker processes via -m

meejah:
This allows us to correctly honour "whatever it is that Python does" to allow .pyc or .pyo files to be used (e.g. if .py files don't exist).

I propose this as an alternative to achieve the same goal as #638

### DIFF
--- a/crossbar/controller/process.py
+++ b/crossbar/controller/process.py
@@ -47,7 +47,6 @@ from autobahn.util import utcnow, utcstr
 from autobahn.wamp.exception import ApplicationError
 from autobahn.wamp.types import PublishOptions, RegisterOptions
 
-import crossbar
 from crossbar.common import checkconfig
 from crossbar.twisted.processutil import WorkerProcessEndpoint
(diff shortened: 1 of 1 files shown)


yegorich:
@meejah seems to work. Thanks for the fix.

meejah:
Great, thanks for the feedback :)

Obviously, I still have some build issues to fix first...

oberstet:
This is of course much more elegantly solving the #638 - Nice, nice ..

yegorich:
@oberstet, @meejah when are you planning to make a new release?

I have one more issue: in https://github.com/crossbario/crossbar/blob/master/crossbar/controller/process.py you try to import shutilwhich. This should not be present in recent Python 3 installations (>=3.3). I'd suggest something like

try
from shutil import which
except
from shutilwhich import which
except
 Alarm Alerm Exit

This should be also addressed in setup.py/requirenments.txt. What do you think?

oberstet:
merged via https://github.com/crossbario/crossbar/pull/788

PR #782, opened 2016-05-07
Improve environment variable parsing and sync document

w1z2g3:


meejah:
LGTM +1

crossbar/common/checkconfig.py
```diff
@@ -258,7 +258,7 @@ def _readenv(var, msg):
         raise InvalidConfigException("{} - environment variable name '{}' does not match pattern '{}'".format(msg, var, _ENV_VAR_PAT_STR))
 
 
-_ENVPAT = re.compile(u"^\$\{(.+)\}$")
+_ENVPAT = re.compile(u"^\$\{([a-zA-Z_][a-zA-Z0-9_]*)\}$")
```
oberstet: There is already a pattern for that: https://github.com/crossbario/crossbar/blob/master/crossbar/common/checkconfig.py#L242 - maybe there was a reason to have 2 functions `_readenv` and `maybe_from_env`, or maybe not. If not, we could consolidate these 2 functions.

Issue #786, opened 2016-05-10
REST bridge: request checking not watertight


oberstet:
Eg I ran into: `options` is contained in the request dict, but it is `null` (not a dictionary). This leads to a CB internal error (because `null` can't be accessed like a dict).

The REST bridge should absolutely do its best at scrubbing and sanitizing ..

hawkowl:
I also believe I fixed this in #799.

Issue #785, opened 2016-05-10
REST bridge: wrong MIME type for 400


oberstet:
When using the **caller** part of the REST bridge, and when the called WAMP procedure produces an error, the bridge (correctly) returns a JSON thing with the WAMP error. It uses status code 400 (rather than 202 for a successful call). However, it does not set the MIME type as for 202: it is `text/html`.

Now, in retrospective, I'm not sure if it is a good idea to have a different status code (400) for WAMP app level errors anyway.

So either we change the behavior of the REST bridge to always return 202 (application/json) for any WAMP response and only return 400 on bad requests and 500 on internal errors OR the bridge should return the proper MIME type.

oberstet:
As far as I see, we do use these HTTP response status codes: 200, 202, 400, 500

Maybe we can do with 3 instead?

* 200: WAMP processed (can either be a WAMP level success or a WAMP level error!)
* 400: bogus HTTP request (the client did wrong)
* 500: internal error (we screwed up .. there is a bug in CB)

meejah:
Maybe not appropriate for this ticket, but related in a sense is a similar situation with the "transport" versus "wamp session" especially when connecting -- if we make a connection to crossbar successfully (at the transport level) but there's a WAMP-level error (e.g. "realm not found" when joining), we get an ``onLeave`` notification (i.e. WAMP level) of the error, but the transport closes cleanly.

So, to be "philosophically similar" in the ReST bridge, it would point to "always return 202" if the mechanics are correct (e.g. a valid HTTP request etc) but if there was a WAMP-layer error (e.g. not authorized or something) that comes out in the JSON --  and use 400/500 as described above.

The counter-point might be that it would make WAMP-layer errors more-obvious to clients if they were reflected in HTTP-level error-codes -- but that might be incorrectly blurring the line between transport (HTTP in this case) and WAMP...

TL;DR == I agree  your comment above makes ReST bridge more-consistent with rest of crossbar's behavior :)

hawkowl:
This appears to be fixed in #799.

PR #799, opened 2016-05-19
don't actually log these failures

hawkowl:


hawkowl:
cc @oberstet, good to go when the statuses are green

oberstet:
Cool, this is improving things! Here my review.

Smallish:

1. Maybe use `_set_common_headers` throughout (well, where it may apply) instead of having the respective code "duplicated" (one place to maintain is better)
2. [This](https://github.com/hawkowl/crossbar/blob/a30e857069bfd45614fc356517d38d18d70c5f16/crossbar/adapter/rest/caller.py#L57) is more robust than [this](https://github.com/hawkowl/crossbar/blob/a30e857069bfd45614fc356517d38d18d70c5f16/crossbar/adapter/rest/publisher.py#L56) - please use the former for both

Important:

1. Please test that in all errors thrown from [this](https://github.com/crossbario/autobahn-python/blob/master/examples/twisted/wamp/rpc/errors/backend.py) example app code, the caller endpoint returns a 200 with an `{"error": .., }` including the respective `args` and `kwargs` with error payload
2. In the docs for the HTTP bridge: could we have a complete description of the expected behavior of the bridge rgd the "HTTP level" .. in particular status codes and such. The bridge is part of the public API of CB, and it isn't spec'ed anywhere. See also: http://crossbar.io/docs/Compatibility-Policy/ (where the HTTP bridge isn't mentioned, but it should be)

hawkowl:
Thanks @oberstet, I'll fix these up and reopen it + CC you when it's ready :)

hawkowl:
Reopening so that Travis will do its thing...

hawkowl:
@oberstet I've fixed small points 1 (which should fix https://github.com/crossbario/crossbar/issues/785 ? ) and 2 (so I believe this should fix https://github.com/crossbario/crossbar/issues/786 too? I need to write a test-case for that), and important point 1. Important point 2 I will do in a followup PR and file a new issue, since this diff is getting pretty big.

PR #800, opened 2016-05-19
More towncrier changelog

meejah:
This adds a ``make news`` target that runs ``towncrier`` and cat's together the resulting `NEWS.md` file and a header into `docs/pages/ChangeLog.md`

I also inspected the logs from v0.13.2 until now and made up some newfragments for (I hope!) all the changes, so 0.14.0 can grow a changelog :)

So, @oberstet should just be able to run "make news" and "git commit" it right before tagging a new release...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,11 +25,11 @@ clean:
 	find . \( -name "*__pycache__" -type d \) -prune -exec rm -rf {} +
 
 news: towncrier.ini crossbar/newsfragments/*.*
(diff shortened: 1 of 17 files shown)


meejah:
Oh, a minor nit that will need another towncrier patch: the file is always called ``NEWS.rst`` despite the fact that in this case it's really markdown to match the rest of the docs (so input/output filename option needed for towncrier).

oberstet:
@meejah coool! thanks a lot .. also for digging through commits and making up news fragments after the fact;)

PR #812, opened 2016-05-26
Release 0 14 0

oberstet:


meejah:
@oberstet: One thing: if we follow the "recommended" workflow for ``towncrier`` we should delete all the notes in ``crossbar/newsfragments/*`` along with a release. By default towncrier should have staged these when it ran.

So, then the *next* release it should just *modify* ``NEWS.md`` and ``ChangeLog.md`` with the new ones...That is, "crossbar/newsfragments/*" is all the "new, but unreleased" fixes and NEWS/ChangeLog are all the past releases...I think.

Issue #825, opened 2016-06-30
dynamic auth followed by cookie auth are out of sync for realm assignment


FirefighterBlu3:
if a client originally authenticates like so:

realm: "foo" (not a configured realm), and the authenticator reassigns to realm: "bar", when the client browser reloads/reconnects, the `RouterSession.onHello()` method is passed the realm "foo" from the client and the router will attempt to find the realm "foo" in its known realms (`router/session.py:569`) and issue an Internal error (`router/session.py:656`) instead of using the authentication assigned realm "bar".

the cookie authenticator needs to be able to recover the realm that the authenticator originally assigned to the client

PR #826, opened 2016-06-30
When a dynamic authentication is used and the authenticator reassigns…

FirefighterBlu3:
… the client realm, and the client reconnects - the reassigned realm is now recovered from the cookie storage. This PR is for issue #825


goeddea:
CAA received and added.

oberstet:
right, good catch!

FirefighterBlu3:
my pleasure!

PR #832, opened 2016-07-13
add allow_null_origin to websocket options

meejah:
default is true, to match behavior before `allowNullOrigin` was added in Autobahn.


oberstet:
So this does depend on an AB PR, right? Ah, and flake8 is complaining ..

oberstet:
@meejah Should we disable this flake thing?

"crossbar/router/protocol.py:106:5: E731 do not assign a lambda expression, use a def"

meejah:
@oberstet yeah, I think it's fine with the lambda; intention more-clear than a `def per_message_deflate(): pass` anyway...

PR #834, opened 2016-07-14
Unitransport

oberstet:


crossbar/router/unisocket.py
```diff
@@ -0,0 +1,150 @@
...
+        self._proto = None
+        self._data = b''
+
+    def dataReceived(self, data):
+
+        if self._proto:
+            # we already determined the actual protocol to speak. just forward received data
+            self._proto.dataReceived(data)
+        else:
+            if data[0] == b'\x7F':
```
hawkowl: This will need to be `data[0:1]` for Python 3 (since indexing bytes on Py3 returns an integer, not a byte, but slicing returns bytes)

oberstet: Ah, good spot! Corrected ..

Issue #844, opened 2016-07-26
Cannot set 'shutdown' controller attribute


aquemy:
Hi,

FYI: Crossbar.io Version: 0.15.0, Python 2.7

I wanted to modify the ```shutdown``` controller attribute and I understood it should be a list (by the way, it is not documented [here](http://crossbar.io/docs/Controller-Configuration/)).

However, with or without ~~you~~ a list, I got the following:

```console
raise InvalidConfigException("invalid type {} for 'shutdown' in node controller options (must be a list)".format(type(options['shutdown_mode'])
```

I am using Python 2.7 and the test [here](https://github.com/crossbario/crossbar/blob/f4fa8a00b15d683280f36cfae6af2e9cad5e887a/crossbar/common/checkconfig.py#L2641) returns false when I do a simple manual test:

```python
>>> type(options['shutdown'])
<type 'list'>
>>> type(options['shutdown']) == Sequence
False
```

I suppose this is a problem of conversion earlier in the configuration check process but I am not really familiar with it.

On top of that ```['shutdown_mode']``` should be ```shutdown``` [here](https://github.com/crossbario/crossbar/blob/f4fa8a00b15d683280f36cfae6af2e9cad5e887a/crossbar/common/checkconfig.py#L2642).

Regards,

meejah:
Yes, the check should be using `isinstance` not `==`

oberstet:
Fixed via https://github.com/crossbario/crossbar/pull/845

PR #845, opened 2016-07-27
Issue844 shutdown

meejah:
This fixes the checks for the 'shutdown' option and adds some documentation (see #844)

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2586,8 +2586,8 @@ def check_controller_options(options):
             raise InvalidConfigException("'title' in 'options' in controller configuration must be a string ({} encountered)".format(type(title)))
 
     if 'shutdown' in options:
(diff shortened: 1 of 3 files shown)


meejah:
Hmm, not sure what's up with Travis -- works with `tox` locally but looks like it can't find coverage-data on travis...

oberstet:
This is certainly a bug, and the PR looks good to me! Why does Travis fail .. no clue .. but the repo is configured to only allow merging when green =(

meejah:
Okay, I'm looking into what's different with travis versus just running "tox" locally...

meejah:
Looks like something with the new coverage 4.2 release (also, I learned that coverage has a C-extension?!)

For now, this pins to `<4.2` if you merge it .. I will figure out what's different though.

meejah:
okay, we don't have to pin version, must pass `--append` to `coverage combine`

PR #862, opened 2016-08-26
"authextra" in dynamic authenticators

### LINKED ISSUE #853: Dynamic authentication: authextra argument will not be forwarded.
When using dynamic authentication the `authextra` argument of `def join(self, realm, authmethods=None, authid=None, authrole=None, authextra=None):` will not be forwarded from the client to the authenticator when using an other authentication method than `WAMP-cryptosign`.

I have discussed this issue with @oberstet on the Autobahn Google Group in this [Post](https://groups.google.com/forum/#!topic/autobahnws/6f9nehxn_6c).

Fixing the bug is very straightforward and requires adding the following code line to every `PendingAuth`-Subclass in the "dynamic authentication" part of the overridden `def hello(self, realm, details)` (except cryptosign.py, which is already forwarding the `authextra` argument):

    self._session_details[u'authextra'] = details.authextra

The Bug is located in the Authentication Subclasses (PendingAuth Subclasses) [here](https://github.com/crossbario/crossbar/tree/master/crossbar/router/auth). This line can be found in cryptosign.py at line 170 and is missing in every other subclass.

I have cloned the crossbar sources and fixed it myself to see if this is fixing the bug and yes, it is! When this line is added, the references are build up and the `authextra` argument is being forwarded into the dynamic authenticator, allowing clients to transmit additional information when requesting authentication.

meejah:
This should fix #853 for all `dynamic` authenticators by passing along `authmethod` and `authextra` in the `details=` kwarg to the dynamic authentication call (in the same way `cryptosign` already does).

Also adds unit-tests for all these.

### DIFF
--- a/crossbar/newsfragments/853.bugfix
+++ b/crossbar/newsfragments/853.bugfix
@@ -1,0 +1,1 @@
+properly pass 'authextra' and 'authmethod' keys to all dynamic authenticators
--- a/crossbar/router/auth/anonymous.py
+++ b/crossbar/router/auth/anonymous.py
@@ -85,6 +85,8 @@ class PendingAuthAnonymous(PendingAuth):
             if error:
                 return error
 
+            self._session_details[u'authmethod'] = self._authmethod  # from AUTHMETHOD, via base
+            self._session_details[u'authextra'] = details.authextra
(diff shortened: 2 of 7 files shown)


oberstet:
LGTM, thanks!

Issue #610, opened 2016-01-25
Long-poll broken on Python 3 (probably Twisted Web)


oberstet:
Long-poll fails on Python 3 with this response: `POST http://127.0.0.1:8080/lp/open 500 (Internal Server Error)`. This is coming from Twisted.

To reproduce, go to https://github.com/crossbario/crossbarexamples/tree/master/longpoll, start crossbar, open http://localhost:8080 in browser and click "Open Session"

hawkowl:
This seems to be fixed when using this version:

```
     __  __  __  __  __  __      __     __
    /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/

 Crossbar.io        : 16.10.dev1
   Autobahn         : 0.16.0 (with JSON, MessagePack, CBOR, UBJSON)
   Twisted          : 16.3.0-SelectReactor
   LMDB             : 0.89/lmdb-0.9.18
   Python           : 3.5.2/CPython
 OS                 : Darwin-15.4.0-x86_64-i386-64bit
 Machine            : x86_64
```

joachimneu:
I still observe the issue using this version:
```
     __  __  __  __  __  __      __     __
    /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/
                                        
 Crossbar.io        : 16.10.dev1
   Autobahn         : 0.16.0 (with JSON, MessagePack, CBOR, UBJSON)
   Twisted          : 16.3.0-EPollReactor
   LMDB             : 0.89/lmdb-0.9.18
   Python           : 3.5.1/CPython
 OS                 : Linux-4.5.1-1-ARCH-x86_64-with-arch
 Machine            : x86_64
```

meejah:
I can confirm this is still a problem with Python 3.5 on Debian with crossbar + autobahn master -- some  issues with bytes vs str, I have a patch coming.

joachimneu:
It works now in my case. Thank you so much! :)

PR #864, opened 2016-09-01
#610 long poll python3

meejah:
There's two pieces here: one fixes a bunch of logging things, and the other fixes some incorrect `bytes` vs `str` handling. See #610 

For logging:

 - use late string-interpolation always
 - debug-log tracebacks for errors
 - use the autobahn trick to lazily-hexlify debug octet logging

For longpoll specifically:

 - always return bytes for responses etc
 - simplify some of the failure messages (e.g. remove exception messages; they're in debug/info logs now so there's no good reason to return them to the client)

I tested against the longpoll example on both Python 3.5 and 2.7 (under Debian).

### DIFF
--- a/crossbar/router/longpoll.py
+++ b/crossbar/router/longpoll.py
@@ -43,7 +43,7 @@ from twisted.web.resource import Resource, NoResource
 from twisted.web import http
 from twisted.web.server import NOT_DONE_YET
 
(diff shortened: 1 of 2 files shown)


oberstet:
Perfect. I totally forgot we have the `_LazyHexFormatter` ;)

PR #865, opened 2016-09-02
convert all log calls using .format() to use kwargs

meejah:
After looking at #864 I ran a search for all uses of `.format` with a `.log.*` call and re-worked them all to not use `.format`.

There's a really nasty failure-case if you *just* pass a string-to-log to the logging calls: if that string happens to have `{}`'s in it, then the logger will try to interpolate it and most likely fail. I find this happens "somewhat often" when printing exceptions. The safest way to log a straight string is like:

```
self.log.debug("{msg}", msg="a string that can contain { and } safely")
```


oberstet:
I see. This is indeed nasty;) The straight string logging .. thanks for going all over the code!

Issue #817, opened 2016-06-01
Authorizer actions are not documented


ecorm:
The list of possible values for the `action` argument of the dynamic authorizer are not documented. In the meantime, if someone familiar with the Crossbar codebase can post a link to the relevant source file, that would be great.

The possible non-boolean return values should also be documented.

oberstet:
register/subscribe actions:

```json
{
    "allow": true
}
```

call/publish actions:

```json
{
    "allow": true,
    "disclose": true
}
```

> Note: we'll also have an attribute `"cache": true` in the future for all actions.

oberstet:
See: https://github.com/crossbario/crossbar/blob/master/crossbar/router/role.py#L286

oberstet:
For backwards compat, you can also return a bool: https://github.com/crossbario/crossbar/blob/master/crossbar/router/router.py#L274

PR #867, opened 2016-09-02
Dynamic auth improvements

meejah:
While looking at #817 and documenting dynamic authorization, I made some improvements. The last commits completely fails authorization if the authorizor returns incorrect information, while with just the second-last commit the call will still go through but with default authorization (I think it's safest to do the behavior with all commits in this branch).

Note this is on top of the documentation PR (#866) so you could *just* merge this one and close that one ...


oberstet:
NIce, nice. This make things more robust. Merging.

Issue #870, opened 2016-09-06
Missing machine id on mac


markope:
```
2016-09-06T23:33:55+0200 [Controller  84758] Traceback (most recent call last):
2016-09-06T23:33:55+0200 [Controller  84758]   File "/usr/local/bin/crossbar", line 11, in <module>
2016-09-06T23:33:55+0200 [Controller  84758]     load_entry_point('crossbar', 'console_scripts', 'crossbar')()
2016-09-06T23:33:55+0200 [Controller  84758]   File "/Users/marko/git/crossbar/crossbar/controller/cli.py", line 1023, in run
2016-09-06T23:33:55+0200 [Controller  84758]     options.func(options, reactor=reactor)
2016-09-06T23:33:55+0200 [Controller  84758]   File "/Users/marko/git/crossbar/crossbar/controller/cli.py", line 598, in run_command_start
2016-09-06T23:33:55+0200 [Controller  84758]     pubkey = node.maybe_generate_key(options.cbdir)
2016-09-06T23:33:55+0200 [Controller  84758]   File "/Users/marko/git/crossbar/crossbar/controller/node.py", line 279, in maybe_generate_key
2016-09-06T23:33:55+0200 [Controller  84758]     raise Exception("Corrupt node private key file {} - {} tag not found".format(privkey_path, tag))
2016-09-06T23:33:55+0200 [Controller  84758] Exception: Corrupt node private key file /Users/marko/git/REone/re_crossbar/.crossbar/key.priv - machine-id tag not
```

The issue is this line: 
https://github.com/crossbario/crossbar/blame/master/crossbar/controller/node.py#L191

When the machine id is not present there should be a key "machine-id" with value None.

oberstet:
@markope could pls you post the complete output of the commands

* `which ioreg`
* `ioreg -rd1 -c IOPlatformExpertDevice` 

here? I lack a OSX box ..

Getting a unique machineID on OSX: http://stackoverflow.com/questions/933460/unique-hardware-id-in-mac-os-x

markope:
```console
Markos-MacBook-Pro:RA marko$ which ioreg
/usr/sbin/ioreg
Markos-MacBook-Pro:RA marko$ ioreg -rd1 -c IOPlatformExpertDevice
+-o MacBookPro11,4  <class IOPlatformExpertDevice, id 0x100000112, registered, matched, active, busy 0 (209390 ms), retain 42>
    {
      "compatible" = <"MacBookPro11,4">
[...]

PR #879, opened 2016-09-24
[SUPER WIP] MQTT

hawkowl:
This implements MQTT using Crossbar.io as the message broker.


oberstet:
Awesome!! The code is really looking great, clean and well factored. I'll now go over in detail ..

crossbar/controller/node.py
```diff
@@ -167,7 +167,8 @@ def _machine_id():
         with open('/var/lib/dbus/machine-id', 'r') as f:
             return f.read().strip()
     except:
-        return None
+        # OS X? Something else? Get a hostname, at least.
```
oberstet: Yeah, something else;) Please see https://github.com/crossbario/crossbar/issues/870

crossbar/common/checkconfig.py
```diff
@@ -1544,6 +1544,29 @@ def check_listening_transport_web(transport, with_endpoint=True):
 _WEB_PATH_PATH = re.compile(_WEB_PATH_PAT_STR)
 
 
+def check_listening_transport_mqtt(transport):
```
oberstet: This is great .. so CB will be able to talk MQTT over Unix domain sockets;)

Default ports for MQTT are: "TCP/IP port 1883 is reserved with IANA for use with MQTT. TCP/IP port 8883 is also registered, for using MQTT over SSL." - But since we don't supply default ports for any of the other transports as well, it makes sense to require the user to explicitly specify the port to be used for MQTT as well. IOW: the MQTT ports should be simply documented in the CB docs, and that's it.

I am wondering if we could expand the "universal transport" also to allow running everything on 1 port (WebSocket/WAMP, RawSocket/WAMP, Web and MQTT - like on TLS/443) - of course that means MQTT clients need to connect on a non-standard port, but MQTT clients should allow that?

What is the first octet on an incoming MQTT connection?

hawkowl:
Hmm. The first octet is \x10. That is, 4 bit int to say packet 1, and 4 bits of flags all being 0. Anything else as the first bits is a protocol violation.

I have no idea if most clients let you freely pick the port, but twisted-mqtt does, at least; I've not investigated the others. I think it'd be standard enough that we could utilise unisocket, although maybe we should have the others as a fallback?

- Amber

crossbar/common/checkconfig.py
oberstet: [Seems](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718028) like the first octet is `0x10`, right?

* `0x10` is a LF in ASCII, and I think this cannot be a valid first octet for any HTTP request? At least practically ..
* `0x7F' is first octet of RawSocket

So we probably could make it work (I mean, being able to run MQTT on a standalone transport port *and* being able to run it under universal transport switching).

The thing with that is: one could then use 1 TLS certificate for everything .. which is nice.

oberstet:
> I think it'd be standard enough that we could utilise unisocket, although maybe we should have the others as a fallback?

Yes, both. That is, ability to run MQTT on standard port (1883/8883) on 
a standalone transport _and_ (optionally) to stick MQTT on unisocket ..

---

https://github.com/wolfSSL/wolfMQTT/

for example also seems to allow to specify the port to connect to

https://github.com/wolfSSL/wolfMQTT/blob/master/examples/mqttexample.c#L116

I am considering

https://www.wolfssl.com/wolfSSL/Products-wolfmqtt.html

as a benchmark client to test agaist .. from reading, it looks good

http://www.hivemq.com/blog/mqtt-client-library-enyclopedia-wolfmqtt
http://www.prweb.com/releases/2016/02/prweb13221670.htm
https://community.arm.com/docs/DOC-11314

It claims to be able to go down to tiny devices, and it is commercially 
supported.

oberstet:
@hawkowl it isn't finished, but could you make CI pass and I merge nevertheless (it is non-invasive, no problems? I have broken down the things still needed here: https://github.com/crossbario/crossbar/issues/881

> Note: this doesn't yet address the "hard part" .. session persistence and QoS .. I am working on this ..

hawkowl:
@oberstet the tests should be green now (when they finish running) :) -- all the changes ought to be non-invasive, purely additive, so should be good to merge!

oberstet:
@hawkowl cool! there are some strange errs in some envs left .. if you can make it green, I'll merge! I have also have written something up about session freeze/resume .. will post that in an issue.

hawkowl:
@oberstet I've fixed the build issues, and patched up how our coverage is reported :)

oberstet:
Awesome!! This is a big thing!

> Signed commit .. hey, I also want that nice green box;)

hawkowl:
@oberstet wrt the signed commit, you can see how to do it at https://help.github.com/articles/signing-commits-using-gpg/ :)

hawkowl:
@oberstet wrt the signed commit, you can see how to do it at https://help.github.com/articles/signing-commits-using-gpg/ :)

Also, which issue did you write about session freezing in?

PR #906, opened 2016-10-18
Fixed typo in Path Service docs

normanjaeckel:


normanjaeckel:
Sorry but I do not know what the change in line 59 is. GitHub made this when I was using the editing UI.

Issue #889, opened 2016-09-25
MQTT: acknowledged delivery


oberstet:
WAMP and Crossbar.io has (optional) acknowledged publications. WAMP currently lacks acknowledged delivery of events (eg there is a WAMP EVENT message, but there is not response message to the former).

In it's simplest incarnation, we would need a WAMP message

    [EVENT_PROCESSED, EVENT.Publication|id]

that acknowledges the reception of an `EVENT` that was to be delivered with acknowledged delivery.

This requires work on WAMP and AutobahnPython first.

hawkowl:
I am guessing this would go into https://raw.githubusercontent.com/wamp-proto/wamp-proto/master/rfc/text/basic/bp_pubsub_publishing_and_events.md ? Or would this be an 'advanced profile' item?

oberstet:
This should be part of the WAMP Advanced Profile.

hawkowl:
Right. Are we also going to have exactly-once QoS in WAMP, too? Since this would do QoS1 either way, but not 2.

oberstet:
Yes, this is only about QoS 1 (in MQTT terms).

> QoS 2 is a whole different kind of beast .. need to see where this goes.

PR #908, opened 2016-10-22
More MQTT work

### LINKED ISSUE #890: MQTT: universal transport multiplexing
The IANA registered default ports for MQTT are:
- port 1883 for non-TLS MQTT
- port 8883 for MQTT over TLS

Since we don't set default ports for any of transports defined in CB node configurations, we should require an explicit port in MQTT transport configs as well but at least document the ports that should be used with MQTT.

That being said, users might run MQTT on different ports. One obvious candidate is running secure MQTT (over TLS) on the standard HTTP/TLS port 443.

Fruther, we should expand the "universal transport" also to allow running everything on 1 port (WebSocket/WAMP, RawSocket/WAMP, Web and MQTT - like on TLS/443) - of course that means MQTT clients need to connect on a non-standard port, but at least some MQTT clients allow that. One could then use 1 TLS certificate for everything .. which is nice.

The very first octet in a new MQTT client connection that is sent in the MQTT CONNECT message to the server is `0x10` (see [here](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718028)). And `0x10` is a `LF` in ASCII, and I think this cannot be a valid first octet for any HTTP request? At least practically .. but we should recheck that.

So the multiplexing in [unisocket](https://github.com/crossbario/crossbar/blob/master/crossbar/router/unisocket.py) could be expanded accordingly.

hawkowl:
Fixes #890 , and expands the test coverage + implementation to support QoS 2 publishes on the protocol (but not the WAMP) layer.


hawkowl:
cc @oberstet :)

hawkowl:
@oberstet there are a few other fixes for stuff in here, did you want me to cherry pick them out on other branches? i've managed to get them in fixed commits if you want to see the fixes individually :)

oberstet:
No, this implements QoS at the adapter level. The adapter should only be a shim, with no substantial own semantics. 

Sidenote: it also adds a new dep only to get serial on osx.

hawkowl:
@oberstet I guess we should tackle https://github.com/crossbario/crossbar/issues/889 then, though? I'm not really sure what you want the semantics to be -- MQTT's is between each hop, but I think you want an end to end QoS?

oberstet:
No, the semantics is fine (pubsub cannot have end-to-end QoS by definition), but the code is in the wrong place. It should be implemented inside the broker core of CB. The MQTT adapter should essentially only adapt to the wire, with no own intelligence whatsoever. This won't happen tomorrow (since it requires WAMP spec work), and hence, the MQTT support will stay at QoS 0 for now.

oberstet:
A lot of wasted work (please ask me if unsure before): maybe some of the PR can be rescued? Essentially, anything but any "smartness" in the adapter itself.

hawkowl:
@oberstet It's actually just a handful of lines that has to change in the adapter to delegate to WAMP (the tests are mostly valid and test MQTT compliance, I just have to remove the checks for the existing internal state).

There's some places where I'm not sure about how we actually can map MQTT to WAMP in this case. MQTT requires resending unacknowledged publish messages under QoS1/2 if the connection is lost, but we don't have that disconnection in WAMP yet (maybe I'll look at how we can do that today).

oberstet:
As we discussed elsewhere, those QoS levels only make sense in the context of session resumption, as TCP takes care of relibility during a session. And right now in WAMP, we don't have session resumption. Pls see here for a longer treatment https://github.com/wamp-proto/wamp-proto/blob/master/rfc/session-state-machine.md

hawkowl:
Yeah, I think we have to decouple those before QoS in WAMP can be handled, yeah... but I'm wondering if the QoS1/2 is even sort of WAMP-y as it seems the thing to do is use event history? Or would a WAMP router keep the in-flight needing-ack WAMP events that it's already sent before the connection was lost but never got an ack to, and then resend *just* those upon session resumption, and leave it up to the client to use event history to get things after the WAMP router knew of its disconnection?

oberstet:
> but I'm wondering if the QoS1/2 is even sort of WAMP-y as it seems the thing to do is use event history?

Yeah! My personal take on this is: 99% of use cases are better off with event history and NOT paying the price for requiring the router to remember which event it dispatched to which subscriber already (in the presence of lost clients).

This was (and still is) my original thinking when we started WAMP.

Only change in my view now is: I do think we can have it both! 

On the other hand: I am wondering how MQTT can live up to it's "exactly once" delivery QoS without employing a full blown [two-phase commit protocol](https://en.wikipedia.org/wiki/Two-phase_commit_protocol) message exchange.

Eg say the router has delivered an event, the subscribed has responded with ACK for reception, but then the network goes down before the router receives the ACK _and_ the client crashes. When the client comes up again, it needs to have a log of the events it already processed and resend the ACKs the router didn't get (otherwise the router would resend the event).

Let's first tackle:

* "normal" acknowledged delivery (which we also need) https://github.com/crossbario/crossbar/issues/889
* "session freeze/resume"

and then go after QoS step-by-step.

hawkowl:
>  I am wondering how MQTT can live up to it's "exactly once" delivery QoS without employing a full blown two-phase commit protocol message exchange.

It kind of does! It has a publish -> received -> release -> complete sort of thing -- the release/complete steps mean that the client also has confirmation that the server got their ack (meaning they won't get a republish in the future, and that any publish with the same content/ID is actually a new one).

> When the client comes up again, it needs to have a log of the events it already processed and resend the ACKs the router didn't get (otherwise the router would resend the event).

MQTT does have a 'dup' flag for this purpose -- if it's a QoS 1 or 2 message, and it gets a dup=1 Publish, it can compare it to what it believes it has got. If, on reconnection, the client gets a dup=1 but it thought it sent the ack, it can know that the ack was not received and redo it.

So, just to confirm, on the WAMP side -- do we resend things we *think* they might have got (ie. non-ack'd messages sent before connection) and then leave it up to event history for the intervening times (which the MQTT adapter can call to do what MQTT clients expect), or do we just leave it up to the client to interrogate the event history for things that the server would retry under MQTT?

This is where WAMP and MQTT might diverge -- if it's the former then the adapter just sends on the router's retry, if it's the latter then we might need to have some smarts in the adapter, as the MQTT spec has some guarantees for messages that the server sent and never got ACK'd, to be compliant. In either case, the adapter will have to remember when it sent the last message to the client to fufil MQTT's subscribe-queue-whilst-client-is-offline spec requirements, just calling the WAMP event history API and then doing in-order publishes, to make it drop-in compared to other MQTT servers.

PR #922, opened 2016-11-15
Prevent automatic Twisted reactor loading

### LINKED ISSUE #920: Running on *BSD or MacOSX, but cannot install kqueue Twisted reactor
Full error: https://gist.github.com/jscheel/26aa1fb1d1db9bfa0a76794e73bab5ec
Version output: https://gist.github.com/jscheel/84c13f24340cf2a637a2b00c59b7f90e

IRC conversation
```
jscheel: also, I’m getting this error when installing / running: “Running on *BSD or MacOSX, but cannot install kqueue Twisted reactor”
[11:55am] jscheel: coming from this: raise error.ReactorAlreadyInstalledError("reactor already installed")
[11:55am] jscheel: any idea what might be causing that?
[11:56am] jscheel: I installed crossbar with pip3 on el capitan
[11:57am] oberstet: jscheel: do you get that message when starting crossbar (in the startup log)?
[11:57am] jscheel: oberstet: yes
[11:58am] oberstet: could you paste the output from "crossbar version" and the beginning of the log somewhere?
[11:58am] jscheel: oberstet: full error: https://gist.github.com/jscheel/26aa1fb1d1db9bfa0a76794e73bab5ec
[11:58am] jscheel: oberstet: version: https://gist.github.com/jscheel/84c13f24340cf2a637a2b00c59b7f90e
[12:00pm] oberstet: this isn't fatal or breaking things, but bad (needs fix) - it means, you dont get max performance
[12:00pm] oberstet: background: something is importing a Twisted class which in turn imports reactor, and once we get to importing "our optimal" one, we can't, because it was already imported
```

hawkowl:
Makes KQueue on OSX/BSD and IOCP on Windows work now, fixes #920

### DIFF
--- a/crossbar/worker/router.py
+++ b/crossbar/worker/router.py
@@ -90,8 +90,7 @@ from crossbar.twisted.site import createHSTSRequestFactory
 
 from crossbar.twisted.resource import JsonResource, \
     Resource404, \
-    RedirectResource, \
-    ReverseProxyResource
+    RedirectResource
 
 from crossbar.adapter.mqtt.wamp import WampMQTTServerFactory
 
@@ -1062,6 +1061,10 @@ class RouterWorkerSession(NativeWorkerSession):
         # Reverse proxy resource
         #
         elif path_config['type'] == 'reverseproxy':
+
+            # Import late because t.w.proxy imports the reactor
+            from twisted.web.proxy import ReverseProxyResource
+
             host = path_config['host']
             port = int(path_config.get('port', 80))
             path = path_config.get('path', '').encode('ascii', 'ignore')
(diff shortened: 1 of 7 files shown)


crossbar/worker/router.py
```diff
@@ -90,8 +90,7 @@
 
 from crossbar.twisted.resource import JsonResource, \
     Resource404, \
-    RedirectResource, \
```
oberstet: what about this?

oberstet:
Thanks! I think just a minor glitch .. see comment.

> This global reactor thing is so annoying. Only second to the GIL;)

crossbar/worker/router.py
hawkowl: `crossbar.twisted.resource` doesn't import the global reactor (I got rid of the re-export which did and moved it into the only usage of it down on L1065).

hawkowl:
I think I addressed the glitch, will merge so next release has KQueue on OSX and IOCP on Windows again...

PR #924, opened 2016-11-21
More MQTT

hawkowl:
This removes all the QoS smartness from the adapter, and implements the hooks so that the WAMP layer can implement the WAMP QoS responses once they're there. We still send the QoS messages regardless because the spec requires that they are sent, even if it's not relating to any packet that we have ever sent, we have to send back the next step in the QoS. The tests just test the wire now (ie. just the messages sent) and a few of the QoS ones are skipped -- they're compliance tests though, so we'll need to make them pass at some point, once WAMP/CB has QoS2 and message resending in it.


oberstet:
=) Really awesome work. Nice, nice .. so we can have MQTT co-listening with Web, WebSocket and RawSocket all on one (presumably TLS) port! That is way cool.

Issue #873, opened 2016-09-16
cookie name on transport not according to config


markope:
when specifying an alternate cookie name in the crossbar config. (i.e. something else than "cbtid") then this cookie is set properly with the specified name in the http header of my autobahnJS session. But the transport of the session still shows the cookie with default name "cbtid". 

(see line 262 of protocol.py maybe )

PR #926, opened 2016-11-21
Fix some issues to do with cookie authentication

### LINKED ISSUE #895: Missing `authextra` block in saved cookies
Scenario:
- Crossbar 0.15 on FreeBSD 10.1
- `cookie` + `ticket` dynamic authentication by RPC Procedure on PHP Thruway
- AutobahnJS 0.10 client with `cookie` + `ticket` auth enabled

On the initial authentication (`ticket`) authenticator returns to client an `authextra` object with some extras that are needed on the client side. All fine here.

On the second attempt and on the auth is going by `cookie` and can't be completed on the client side, because it lacks that `authextra` block returned from the Crossbar.

When I look into the `cookies.dat` file, I see that the `authextra` block is simply not being stored there and therefore is not being sent to the client.

The easy workaround is to disable the `cookie` authentication, but is there a possibility to save that `authextra` object?

Many thanks

hawkowl:
Fixes #895 and #873

### DIFF
--- a/crossbar/newsfragments/873.bugfix
+++ b/crossbar/newsfragments/873.bugfix
@@ -1,0 +1,1 @@
+When using cookie authentication, the debug logs will now mention the correct cookie name if it is not the default.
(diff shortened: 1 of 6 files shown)


chiefmc:
Looks good! When shall we expect this to be released?

PR #927, opened 2016-11-21
Make `python{2,3} -m crossbar` work

hawkowl:
It uses the same setuptools plugin as Crossbar, so we don't need to test it extensively. This change will:

a) make it easier to pass `-u` for unbuffered, as well as any other Python environments
b) make it easier to use python2 or python3 crossbar, in case you have multiple Crossbars installed (mainly developers here)
c) bring Crossbar in line with other Python utilities that are moving to `-m` as an option (and probably later the default) -- e.g. pip, Twisted, pytest, et al

Maybe references or helps fix https://github.com/crossbario/crossbar/issues/506 ?

### DIFF
--- a/crossbar/__main__.py
+++ b/crossbar/__main__.py
@@ -1,9 +1,9 @@
 #####################################################################################
 #
-#  Copyright (C) Tavendo GmbH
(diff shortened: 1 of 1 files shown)


oberstet:
Very useful. Neat.

PR #931, opened 2016-11-29
Last Will/Testament

hawkowl:
Impl of https://github.com/wamp-proto/wamp-proto/pull/268


hawkowl:
cc @oberstet here's a first stab -- I ran into some problems implementing it where the rest of the WAMP meta API is, as unless you have caller disclosure turned on, the WAMP meta API doesn't know who the caller is. Since it's session based anyway, it seemed like going in there was reasonable.

oberstet:
@hawkowl rgd the problems implementing "wamp.session.add_testament"/"wamp.session.flush_testaments" in the same place as the rest of the meta API (service.py): yes, I see what you mean. We can:

* enforce caller disclosure on all "wamp.*" meta API procs.
* get at the `session_id` of the caller from `details`, and then use `self._router._session_id_to_session` to get the session instance

hawkowl:
`details` doesn't have a `session_id`... but the regular caller disclosure way works, which is how I've implemented it here. Much cleaner. I think it's good to go?

@oberstet

oberstet:
Looks good! Exactly what I had in mind with moving the code to service.py and only enforce disclose in broker/dealer.

> Disclosing on wamp/crossbar _events_ isn't strictly necessary, but it doen't hurt either. It will be "constant" anyway (as the publisher is CB itself). Maybe also use a fixed value of 0 for pub_id. But these are minor things .. merged.

PR #937, opened 2016-12-01
MQTT fixups

meejah:
When making the MQTT example for the Autobahn repo I had to make the following fixes so it would work. 

@hawkowl should probably review in case I got bigger-picture stuff wrong :)


hawkowl:
Yeah, a lot of this is going to need rewriting/deleting anyway, I'll do it on my branch instead...

hawkowl:
(the session handling is all junk and will become part of the proper session resumption in WAMP, not in the adapter)

PR #945, opened 2016-12-17
A way to configure a 'functional' component

meejah:
Here is a (working) sketch of how to take a "functional"-style component and configure it into crossbar.

Basically, it's just configuring methods to call for the 4 listeners: `connect, disconnect, join, leave`.

This sketch only allows for single callbacks -- it might be nicer to allow a list of callbacks, too. To use this, you'd add something like this to a ``components`` list in crossbar config:

```
                {
                    "type": "function",
                    "realm": "crossbardemo",
                    "role": "anonymous",
                    "callbacks": {
                        "join": "func_component.on_join",
                        "leave": "func_component.on_leave",
                        "connect": "func_component.on_connect",
                        "disconnect": "func_component.on_disconnect"
                    }
                }
```

(So it might make sense to allow, e.g., `"join": ["func_component.on_join", "some.other.module.joined_cb"]`)

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2019,7 +2019,7 @@ def check_router_component(component):
         raise InvalidConfigException("missing mandatory attribute 'type' in component")
 
     ctype = component['type']
-    if ctype not in ['wamplet', 'class']:
+    if ctype not in ['wamplet', 'class', 'function']:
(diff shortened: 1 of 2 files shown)


oberstet:
Yeah, allowing lists for handlers would allow to reap the benefits of using such a functional style. Maybe even have a variant where a module name is sufficient when using standard names for function therein (like "on_join", "on_leave", ..) ?

PR #950, opened 2017-01-11
allow cross origin requests to StaticResource

ftes:
Like the JSON value service, the Static Web Service should also allow cross origin requests.


oberstet:
Thanks for contributing! Before merging, we need a CAA: https://github.com/crossbario/crossbar/blob/master/CONTRIBUTING.md#send-us-a-contributor-assignment-agreement

oberstet:
@ftes Thanks for the CAA (got it) and your contribution!

PR #981, opened 2017-03-03
validate MQTT options, assign realm

meejah:
So I'm not actually sure if this is a real bug that should be fixed, or just my misunderstanding.

markos on #autobahn figured out that there wasn't a realm when using `mosquitto_pub` to publish to the MQTT listener. There is a comment that the realm will be automatically assigned, but I don't see where/how that happens.

I also don't know if the config *must* contain `realm` (as I made mandatory in this PR) or the `role` (which is not mandatory, but I think might be unused).


meejah:
This now (see last 3 commits) also fixes some bugs in getting round-trip MQTT messages. This works with `mosquito_pub` and `mosquito_sub` version 1.4.11 (must pass `-V mqttv311`) for me on Debian.

markos:
I just tested it and I can confirm that it works, good job! :+1:

agronholm:
The `self.log.debug(log_category="MQ400")` line should be `self.log.debug(log_category="MQ404")`.

Issue #885, opened 2016-09-25
MQTT: payload mapping


oberstet:
I think this should be configurable on realms and per URI (patterns), similar to [this](https://github.com/crossbario/crossbarexamples/blob/master/event-history/.crossbar/config.json#L6)

```
mqtt_payload_format == opaque | json | cbor | ..
```

With `mqtt_payload_format==opaque', payload from MQTT publisher is published using payload transparency to WAMP side (we need that exposed .. dig through the code [here](https://github.com/crossbario/crossbarexamples/tree/master/encryption/cryptobox)).

When a WAMP publisher publishes to a topic configured for opaque payload on MQTT, it'll be just forwarded "as is" to the MQTT client. If a WAMP publisher does a regular publish to such a topic, the router will log a warning and throw away the event (since it doesn't know how to serialize).

With `mqtt_payload_format==json`, the MQTT client MUST publish payload serialize in JSON with content

``` json
{
   "args": [],
   "kwargs": {}
}
```

If it does publish with nonconforming payload, deny the publish (QoS1/2) or silently throw away the event.

The other direction is straigt forward: WAMP publisher publishes event, event gets serialized to JSON like above for MQTT.

hawkowl:
I think this is the next big important thing... will have to work on exposing it tomorrow. Unsure how this will work, as it'll need, like, two functions passed to subscribe -- one which takes args/kwargs (and is also used by cryptobox), and one which takes non-cryptobox transparent payloads. What do you think @oberstet?

hawkowl:
it's either that or two subscriptions, one which takes args/kwargs and one which accepts transparent payloads... but that seems kinda silly to me.

oberstet:
Leaving cryptobox out for a moment, here is a sketch / start for discussion:

1. MQTT client publishes event to topic T:

* if the router has T configured to be `mqtt_payload_format == opaque`, then a WAMP event with `args=[<opaque_payload>]` and `kwargs=None` is dispatched by the broker (where `<opaque_payload>` is a binary string - means we finally have to implement transparent base64 encoding for WAMP/JSON too;) The WAMP spec talks about this)
* if the router has T configured to be `mqtt_payload_format == json|cbor|..`, then the broker deserializes the raw payload according to the configured serializer, and will expect a dict with `args` and `kwargs`, and use these to dispatch a WAMP event
[...]

PR #983, opened 2017-03-06
bridge WAMP <->MQTT PUBs as well

markos:
MQTT is b64 encoded, WAMP is just UTF-8, this allows WAMP PUBs to be received to MQTT SUBs

To be used along with: 

https://github.com/crossbario/crossbar/pull/981

### DIFF
--- a/crossbar/adapter/mqtt/wamp.py
+++ b/crossbar/adapter/mqtt/wamp.py
@@ -116,10 +116,21 @@ def wamp_payload_transform(payload_format, event):
     Given an Event sent by a WAMP client, transform it into a payload.
     """
     if payload_format == "opaque":
-        if event.kwargs.get(u"mqtt_message"):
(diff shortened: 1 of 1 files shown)


oberstet:
No, this isn't correct .. we need mapping logic, since WAMP can be arbitrary payload (both non-text, and both multi-position and multi-kw)

oberstet:
@markos pls see the issue here https://github.com/crossbario/crossbar/issues/885

markos:
Ok, understood, then I need to explicitly set the payload format? If that is the case, shouldn't the default be something that "just works" by default? Or am I misunderstanding how it's supposed to work?

markos:
@oberstet I think I understand the problem, my fix was only for the most trivial of cases (WAMP string -> MQTT string), but you need a more complete mapping that covers all other cases as well.

PR #985, opened 2017-03-11
allow configuring via a Twisted endpoint-string

meejah:
This adds a third way to configure endpoints (on top of "tcp" and "unix") which is to directly pass any valid Twisted endpoint-string (i.e. anything `serverFromString` would accept).

The use-case here would be to take advantage of any extra endpoints types that got added via the plugin API (or maybe you find it easier to specify tcp/unix endpoints via the Twisted endpoint string).

For example, you could use the `onion:` endpoint type available if you've installed `txtorcon`or the `i2p:` endpoint if you've installed `txi2p` to provide WAMP/Crossbar services via two different anonymity networks. You could use the `le:` or `lets:` endpoints provided by `txacme`. Presumably there are other examples.

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -924,6 +924,35 @@ def check_connecting_endpoint_unix(endpoint):
         check_endpoint_timeout(endpoint['timeout'])
 
 
+def check_connecting_endpoint_twisted(endpoint):
+    """
+    :param endpoint: The Twisted connecting endpoint to check.
+    :type endpoint: dict
+    """
+    for k in endpoint:
+        if k not in ['type', 'client_string', 'timeout']:
+            raise InvalidConfigException(
+                "encountered unknown attribute '{}' in connecting endpoint".format(k)
+            )
+
+    if 'client_string' not in endpoint:
+        raise InvalidConfigException(
+            "missing mandatory attribute 'client_string' in Twisted endpoint "
+            "item\n\n{}".format(pformat(endpoint))
+        )
+
+    client_string = endpoint['client_string']
+    if not isinstance(client_string, six.text_type):
+        raise InvalidConfigException(
+            "'client_string' attribute in Twisted endpoint must be "
+            "str ({} encountered)".format(type(path)))
+    # can we make Twisted tell us if client_string parses? or just
+    # save it until we actually run clientFromString()?
+
+    if 'timeout' in endpoint:
+        check_endpoint_timeout(endpoint['timeout'])
+
+
 def check_listening_endpoint(endpoint):
     """
     Check a listening endpoint configuration.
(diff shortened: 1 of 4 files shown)


oberstet:
Neat. With this little code;) Did you actually try txtorcon as transport?

> btw: I've praised txtorcon in the interview I did for https://www.podcastinit.com/ last week ;) first among my Py picks .. 2nd: anything from Brian Warner - who will soon be on that podcast too anyway I learned .. ha.

meejah:
Yeah, it works with `onion:8080` or similar if you've got txtorcon installed .. launches Tor etc during "crossbar start" :)

meejah:
Also, checking out podcastinit -- I think it's been in my list for a while, but never actually listened to it ;)

oberstet:
`onion:8080`: coool;)

podcastinit: yeah, wasn't aware too. there are quite some interesting interviews, eg fijal/pypy or armin ronacher. I hope I did a decent job in the interview too .. should be online soonish.

tbh, I am a total Tor dummy .. I only have a rough idea about it. I would need a little working example (in cb examples) and a short README to get started;)

What does it buy me in the end (running CB over Tor) in terms of added privacy or such? I mean, a CB client (usually) will have to authenticate to CB to join a realm and get a role assigned. So it can't be anonymity of the client w.r.t. CB. I guess I just don't know enough ..

oberstet:
What about the connecting side? In CB/checkconfig, above PR adds a new `check_listening_endpoint_twisted`, but not a `check_connecting_endpoint_twisted`. And what about AB? Probably we need it there first .. ApplicationRunner should be able to connect to `onion:8080` ..

meejah:
I have some new-api branches I'm cleaning up that answer the client-side question. Those actually let you pass an `IStreamClientEndpoint` instance but maybe they should (also, or instead) let you do a similar thing to this PR (i.e. pass a string to give to `clientFromString`).

And yes probably adding the client-side of this to crossbar config would be good too. I'll do that as part of this PR.

As far as "what do I get": what an onion-service you is "network location anonymity"; so even though a client still authenticates (so you learn their identity) the client doesn't know "where on the network" the server is (similarly, the server doesn't know "where on the network" the client is, even if they do learn *who* they are via auth). Also onion domains are self-authenticating without relying on Certificate Authorities: they're a hash of the public key, so you can be sure you're connecting to a service run by someone who controls the private-key for the service.

meejah:
BTW, the client-side connection-string would look like `tor:mbxurzjuznuve5dh.onion:8080` or similar.

oberstet:
@meejah should we add `txtorcon` to CBs deps? It wouldn't add any additional reqs above what CB already has, and since it is pure Python ..

There is also a need for docs;) And maybe an example in CB-examples ..

meejah:
If we were going to add `txtorcon` as a dep. it might then be worth doing a little code so we'd add an "official" type for it, e.g. `"type": "tor"` as a 4th option to configure endpoints.

It'd definitely be a "niche" user who wants Tor support, but also interesting. And yes an example would be good in either case :)

oberstet:
Ah yes. Actually, Tor was the only reason I got interested in this patch in the first place.

A specific Tor endpoint would be better plus removing the generic Twisted endpoint stuff again.

The latter leaks a Twisted thing into CBs public API ... I should have remembered that we want to avoid that and get rid of the leaking we already have (mainly py components).

Actually, neither tx nor py should leak into our public api. This leaves us the freedom of reimpl. CB in another language.

oberstet:
So now we have this change (generic tx endpoints leaking), web should leave it undocumented / discourage use .. or remove it again. Actually, the word "Twisted" should not appear in CB user docs ...

PR #986, opened 2017-03-11
functional components for containers

meejah:
This is very similar to an earlier PR enabling "functional-style" components. That PR just enabled them for "router component"; this one does the same for components running in a container.


oberstet:
Ah, I see .. only the config checking was missing, the actual instantiation already was there for both router and container components, as it is shared in `_appsession_loader`. So cool.

Issue #656, opened 2016-02-21
Implement authid/authrole based black-/whitelisting for pubsub


esso23:
I was browsing the specification last week and I stumbled upon black/white-listing using authroles/authid. In our application we have a router and many clients connecting to it and each client exposes a common interface. This interface is usable by administrator client for controlling and monitoring. The problem is, we don't want RPCs from this interface to be accessible by regular (or front-end) clients. Black/white listing using session IDs is not applicable in this case as the clients connecting do not know about other clients (in fact that is the core principle of routed rpc/pubsub so I cannot imagine any use case for this feature).

So my question is, are there any plans on implementing this feature in the near future?

Thanks
Esso

oberstet:
Yes, we will have that feature soon.

> Note that bw-listing is about pubsub, not rpc. If you want to restrict rpc, that can be done by authenticating an admin UI frontend under a different role than regular UI frontends.

Paranaix:
This can already be achieved by now. However you might have to write some utility functions on your own to make this productive.

Just include some identifier representing the principal to which multiple peers belong (e.g an userid) within the uri.  E.g `com.example.notification_for_user.42`. Setup a custom authorizer, extract the userid from the uri and map wamp_session_id -> userid. If they match, allow the action (e.g subscribe or publish) if not prohibit it.

izderadicka:
Current status is bit problematic - we can use this option in autobahn python in publish - but it's ignored by crossbar - and sends events to all subscribers  - could at least CB return error when publishing with this option?  This is potential security issue -  no error is given so client  believes that events are published only to limited subscribers.

markope:
+1 waiting for feature

oberstet:
this has been implemented long ago now:

* `exclude_authid`
* `exclude_authrole`
* `eligible_authid`
* `eligible_authrole`

PR #989, opened 2017-03-15
656 black white listing

meejah:
Implement #656

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -149,6 +149,69 @@ class Broker(object):
         else:
             raise Exception("session with ID {} not attached".format(session._session_id))
 
+    def _filter_publish_receivers(self, receivers, publish):
+        """
+        Internal helper.
+
+        Does all filtering on a candidate set of Publish receivers,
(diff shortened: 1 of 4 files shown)


oberstet:
First, splitting out the code into a filter function, out of the already long code path for publishing: +1

Then, the code looks good as in "it has the desired behavior".

My main issue (which we could postpone): I would expect it to produce a lot of GC pressure.

Because it uses list comprehensions, instead of set functions like `set(eligible) & receivers`.

Eg assume the final, even filtered receiver list is 10k long.

> Sidenote: looking at the code (right now, before the PR): at some point we seem to have lost yielding every N publishes to the final receivers - it is a tight, unlimited loop over all receivers - which is bad if you again take the "large receiver" number scenario. but this is definitely another issue ..

oberstet:
So:

* either merge as is plus file an issue "optimized for large receivers" (which would then need set operators plus maybe batching the receiver publication loop) OR
* do the set operator change on this PR code and leave the batching for another issue.

I am fine with both ..

oberstet:
great, this should produce less GC pressure

PR #990, opened 2017-03-16
do publishes in batches

meejah:
Followup to #989 

This should do publishes in groups of 100 before yielding to the reactor.

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -381,13 +381,22 @@ class Broker(object):
                                                     publisher_authid=publisher_authid,
                                                     publisher_authrole=publisher_authrole,
                                                     topic=topic)
-                            for receiver in receivers:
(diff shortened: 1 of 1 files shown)


oberstet:
yeah, exactly. that's what I meant. and yes, would be good to have it configurable via a router worker option ..

oberstet:
to keep the terminology consistent with eg the WebSocket ping-pong/keep-alive stuff, it should probably be named "chunking" rather than "batching".

* "batching": aggregating multiple individual things into one
* "chunking": disaggregating (splitting) a large thing into multiple smaller ones

http://txaio.readthedocs.io/en/latest/api.html#txaio.make_batched_timer

Maybe a new router worker option: `event_dispatching_chunk_size`

meejah:
Yeah, `chunking` is a better word especially since we've already used batching (differently!)

PR #993, opened 2017-03-17
filter on union of eligible, not individual + docs

meejah:
fixes a logic error in eligble_authid and eligible_authrole filtering

Also re-works the documentation, links to the example and explains the new options

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -171,13 +171,17 @@ class Broker(object):
 
         # if "eligible_authid" we only accept receivers that have the correct authid
         if publish.eligible_authid:
+            eligible = set()
             for aid in publish.eligible_authid:
-                receivers = receivers & self._router._authid_to_sessions.get(aid, set())
+                eligible.update(self._router._authid_to_sessions.get(aid, set()))
+            receivers = receivers & eligible
 
         # if "eligible_authrole" we only accept receivers that have the correct authrole
         if publish.eligible_authrole:
+            eligible = set()
             for ar in publish.eligible_authrole:
-                receivers = receivers & self._router._authrole_to_sessions.get(ar, set())
+                eligible.update(self._router._authrole_to_sessions.get(ar, set()))
+            receivers = receivers & eligible
 
         # remove "excluded" receivers
         #
--- a/docs/pages/programming-guide/pubsub/Subscriber-Black-and-Whitelisting.md
+++ b/docs/pages/programming-guide/pubsub/Subscriber-Black-and-Whitelisting.md
@@ -3,9 +3,11 @@ toc: [Documentation, Programming Guide, Subscriber Black and Whitelisting]
 
 # Subscriber Black- and Whitelisting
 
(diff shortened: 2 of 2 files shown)


crossbar/router/broker.py
```diff
@@ -171,13 +171,17 @@ def _filter_publish_receivers(self, receivers, publish):
 
         # if "eligible_authid" we only accept receivers that have the correct authid
         if publish.eligible_authid:
+            eligible = set()
             for aid in publish.eligible_authid:
-                receivers = receivers & self._router._authid_to_sessions.get(aid, set())
```
markope: I believe this is not correct. You can not iteratively intersect with the sessions of the elements of the eligible_authid list. something like this should work I believe
```python
        if publish.eligible_authid:
            eligible = set()
            for aid in publish.eligible_authid:
                eligible = eligible | self._router._authid_to_sessions.get(aid, set())
            receivers = receivers & eligible
```

(same below for authrole)

oberstet: I'm afraid @markope is right here. eligible needs to intersect the receivers with the unionset of all sets of session IDs mapped from authids/authroles.

Logically, this can be done in a loop as above. This definitely produces temporary sets over and over (within the loop body). Whether the following is producing less GC pressure (and hence is faster), I'm not sure, but here is a functional alternative:

```python
set.union(*[set([1,2,3]), set([2,3,4]), set([3,4,5])])
```

oberstet: Or as complete code:

```python
if publish.eligible_authid:
    receivers &= \
        set.union(*[self._router._authid_to_sessions.get(aid, set()) for aid in publish.eligible_authid])
```

oberstet: I guess, above code at least has a better chance of being optimized, as `set.union` explicitly can work on a list of set-valued arguments. As said, I don't know if say PyPy will be able to optimize away temporary objects.

The time complexity of computing the union set of a list/set of sets should be O(len(s1) + len(s2) + ... len(sn)) ..

oberstet: FWIW, I do think the wording in the WAMP spec is in line with above semantics: https://github.com/wamp-proto/wamp-proto/blob/master/rfc/text/advanced/ap_pubsub_subscriber_blackwhite_listing.md#feature-definition

But maybe the text can be improved to make it more obvious ..

meejah: I believe my (new) code *is* correcting the "iterative" problem @markope identifies?

That is: I'm iteratively union-ing all the authroles/authid and only *then* intersecting against the possible receivers.

meejah: (as in the "red" code that's currently already merged on master is the wrong stuff)

meejah: BTW, I have no idea if the list-comprehension version vs. iteratively "extend()"-ing a set is faster...;)

markope: oh, ok you're right @meejah. Sorry for generating such confusion here.

oberstet:
LGTM

> I got confused too;) but it's good now I'd say

PR #994, opened 2017-03-18
Chunked publishes

meejah:
followup to #656, replaces #990

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2248,12 +2248,33 @@ def check_container_components(components):
 
 def check_router_realm(realm):
     """
-    Checks the configuration for a router realm entry, which can be *either* a dynamic authorizer or static permissions.
+    Checks the configuration for a router realm entry, which can be
+    *either* a dynamic authorizer or static permissions.
     """
     # router/router.py and router/role.py
 
     for role in realm.get('roles', []):
         check_router_realm_role(role)
+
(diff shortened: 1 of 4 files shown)


oberstet:
nearly;)

The config should not be forwarded on the command line (to the forker worker process), but like this:

1. https://github.com/crossbario/crossbar/blob/master/crossbar/worker/router.py#L353 get config dict (and therein, there is the new chunk option)
2. the realm object https://github.com/crossbario/crossbar/blob/master/crossbar/worker/router.py#L384 carries the chunk option to ..
3. router_factory.start_realm https://github.com/crossbario/crossbar/blob/master/crossbar/worker/router.py#L389
4. this arrives here: https://github.com/crossbario/crossbar/blob/master/crossbar/router/router.py#L426 [note: the mqtt_payload_format there is also wrong .. that needs to be carried within the options)
5. the router (for a specific realm) carries broker/dealer, and also the options https://github.com/crossbario/crossbar/blob/master/crossbar/router/router.py#L92
6. and finally, the broker can access 5. then to do its thing

oberstet:
so this is mostly about how to structure things to keep stuff in a coherent design ..

meejah:
I was hoping the command-line option would be wrong :)

So since `start_router_realm` is getting the "realm" config, this is a per-realm option? (not per-router as I have in here?)

meejah:
Okay, now getting the options via `start_router_realm`.

PR #1003, opened 2017-03-27
fix unicode

meejah:
Latest autobahn got rid of a "if py2" check that turned str into unicode, which breaks a bunch of crossbar calls when it's running in python2.


oberstet:
Yeah, I tighten the type checking to fix https://github.com/crossbario/autobahn-python/issues/455

PR #1015, opened 2017-03-31
guard _router being None

meejah:
I think it would be possible to hit this other ways (e.g. client-TLS cert failing to verify) but I didn't check that.

### DIFF
--- a/crossbar/router/session.py
+++ b/crossbar/router/session.py
@@ -764,14 +764,18 @@ class RouterSession(BaseSession):
 
     def onLeave(self, details):
 
-        # todo: move me into detatch when session resumption happens
-        for msg in self._testaments[u"detatched"]:
-            self._router.process(self, msg)
+        # _router can be None when, e.g., authentication fails hard
+        # (e.g. the client aborts the connection during auth challenge
+        # because they hit a syntax error)
+        if self._router is not None:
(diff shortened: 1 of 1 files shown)


oberstet:
So this is a defensive move, or did you actually run into this?

meejah:
I ran into it: if the onChallenge throws e.g. a syntax error or similar then you're in a state where _router is None, but onLeave still gets called (it's still None because a Hello never went out to make it non-None). Maybe this isn't the right solution and I should file a bug for that case?

oberstet:
Oh, I see. I think the change here is then perfectly fine, because it is straight forward, and less complex than getting control over all edges cases of session lifecycle. merging ..

meejah:
The reason I thought TLS might be affected too is because (at least in Twisted) you get `onConnect` but then might get disconnected nearly immediately with certificate verification errors. That is, with an invalid cert you'll still get onConnect always...

oberstet:
Actually, I might have run into this with https://github.com/crossbario/crossbar-examples/tree/master/authentication/tls (invalid cert triggers a traceback)

I was going over many of the examples yesterday before release, but it is daunting now. I only tried so an so many. Hence, yeah, absolutely, better be defensive than sorry;)

Issue #1011, opened 2017-03-30
404 on http://crossbar.io/docs/Using-Schemas/


Yamakaky:
404 Not Found

    Code: NoSuchKey
    Message: The specified key does not exist.
    Key: docs/Using-Schemas/index.html
    RequestId: FDD4A1A8F47AA6F3
    HostId: I5mKmQgs3+qO6gkV6qoqspX1GVFmSqjwqJg+pwEuecVkqQqBSrnT3QdXOPeBjxAKqHv+ViyH31A=

Yamakaky:
Same for http://crossbar.io/docs/Logging-in-Crossbar.io/

oberstet:
The Schema stuff should be removed for now ..

goeddea:
The above links are not contained in the docs anymore.

PR #1033, opened 2017-04-08
Fixed broken doc links for Issue 1011

vetsin:
https://github.com/crossbario/crossbar/issues/1011


goeddea:
Thanks, but the links work as-is (and are now deployed in working order).

PR #1047, opened 2017-04-19
Fix for python 2.7

DZabavchik:
Fails to start on Python 2.7 (PyPy or CPython) if there are containers/components in config. Fixed.

### DIFF
--- a/crossbar/controller/node.py
+++ b/crossbar/controller/node.py
@@ -757,7 +757,7 @@ class Node(object):
                             self._reactor.stop()
                         except twisted.internet.error.ReactorNotRunning:
                             pass
-                    topic = 'crossbar.worker.{}.container.on_component_stop'.format(worker_id)
+                    topic = u'crossbar.worker.{}.container.on_component_stop'.format(worker_id)
                     component_stop_sub = yield self._controller.subscribe(component_exited, topic)
 
                     # start connections (such as PostgreSQL database connection pools)


oberstet:
Uups. Thanks! Hopefully, these are some of the last fallouts from the recent complete type tightening and disabling of all type coercion code in AutobahnPython.

DZabavchik:
Was just trying out 0.17.X, because we're still on 0.15.0.  Besides this, 0.17 also appears to retain zombie registrations when client sessions disconnect. Which is a much greater issue.  I was not able to find the cause.  But it happens when client sessions disconnect in unclean manner (cable pulled, NIC disabled on client).

oberstet:
@DZabavchik Bad, I fully agree. Well. Ah, other thing, have you tried the Docker images? Because these are all based on Python 3 and PyPy 3, where above issue not occur, and I wonder if you also wouldn't see the disconnect issues.

DZabavchik:
(sorry for using wrong thread for this discussion)

I tried latest release and current development on Python 3.6, and was getting `wamp.error.procedure_already_exists`.

Client A connects and registers `a.proc`
Client A connection is lost (i.e. cable unplugged) - session is destroyed on the router due to ping timeout (expected all procedures registered by client A are removed) 
... Minutes/hours later ...
Client A connects again, tries to register `a.proc` again - `wamp.error.procedure_already_exists` is returned
(no other clients connected or registered `a.proc` URI)

Attempts to `call('a.proc')` from other client "succeed" (wamp.error.no_such_procedure is not returned), but never receive any response (invocations accumulate).

Issue #1048, opened 2017-04-22
Relax node key tag checking / improve creator tag


oberstet:
Relax the node key file extended tag checks, or improve the creator determination - this is from the 2nd start of a container. The key was successfully created and the node started fine on the first run, but on the 2nd run, it fails, because the generated key missed the creator tag.

```
(cpy360_1) oberstet@thinkpad-t430s:~/test/crossbar$ docker run --user $UID -v ${PWD}:/node -p 8080:8080 --name crossbar --rm -it crossbario/crossbar
2017-04-22T12:25:20+0000 [Controller      1] Traceback (most recent call last):
2017-04-22T12:25:20+0000 [Controller      1]   File "/usr/local/bin/crossbar", line 11, in <module>
2017-04-22T12:25:20+0000 [Controller      1]     sys.exit(run())
2017-04-22T12:25:20+0000 [Controller      1]   File "/usr/local/lib/python3.6/site-packages/crossbar/controller/cli.py", line 1109, in run
2017-04-22T12:25:20+0000 [Controller      1]     options.func(options, reactor=reactor)
2017-04-22T12:25:20+0000 [Controller      1]   File "/usr/local/lib/python3.6/site-packages/crossbar/controller/cli.py", line 665, in run_command_start
2017-04-22T12:25:20+0000 [Controller      1]     pubkey = node.maybe_generate_key(options.cbdir)
2017-04-22T12:25:20+0000 [Controller      1]   File "/usr/local/lib/python3.6/site-packages/crossbar/controller/node.py", line 299, in maybe_generate_key
2017-04-22T12:25:20+0000 [Controller      1]     raise Exception("Corrupt node private key file {} - {} tag not found".format(privkey_path, tag))
2017-04-22T12:25:20+0000 [Controller      1] Exception: Corrupt node private key file /node/.crossbar/key.priv - creator tag not found
```

and

```
(cpy360_1) oberstet@thinkpad-t430s:~/test/crossbar$ cat .crossbar/key.pub 
Crossbar.io node public key

created-at: 2017-04-22T12:25:12.001Z
machine-id: 92489fc3d3e4
public-key-ed25519: d57d617d45841751c5a914858d4679090af79764959a150176dc9f2e3ffdcc99
(cpy360_1) oberstet@thinkpad-t430s:~/test/crossbar$ cat .crossbar/key.priv 
Crossbar.io node private key - KEEP THIS SAFE!

created-at: 2017-04-22T12:25:12.001Z
[...]

PR #1050, opened 2017-04-26
ticket #1048

meejah:
here's one way to solve #1048. Or did you also want to make "creator" not-required?


oberstet:
Setting it "unknown" is perfectly fine

Issue #944, opened 2016-12-16
Subscribing with a topic of "" and prefix matching should subscribe to all topics


RogMicrotest:
From my question at [stackoverflow](http://stackoverflow.com/questions/41167416/can-i-subscribe-to-all-topics-on-a-crossbar-io-broker-with-a-single-subscription):

I'm trying to monitor all topics published on a crossbar.io broker. My attempt so far is to subscribe to topic="" with matching policy of 'prefix' but that doesn't seem to work, whereas topic="a" with matching policy of 'prefix' does see all events for topics starting with 'a'.

A proprietary broker does allow subscribing to all topics in this way, so I would expect crossbar.io to be the same?

oberstet:
This should work .. needs investigation

RogMicrotest:
It would be nice if it did work, however on re-reading the 'spec', it does state that a URI cannot be an empty string and hence using that as argument to the SUBSCRIBE is not valid.  On the other hand, it would be nice if it worked...

oberstet:
I have finally come around and analyze this - took me almost 2h. Turns out to be a race-like condition:

1) CB produces WAMP/CB meta events .. and these include events when a subscription is created [here](https://github.com/crossbario/crossbar/blob/master/crossbar/router/broker.py#L492)
2) when a client prefix subscribes to `""`, this _also_ includes WAMP/CB meta events, and these events are dispatched to all eligible subscribers - including the one that currently subscribes
3. however, these events (destined for that currently subscribing client) are dispatched _before_ the clients subscribe request returns! and hence, the client will receive an EVENT for a yet unknown subscription ID - this blows up [here](https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/protocol.py#L651) upon the client will drop the connection immediately

---

A similar situation (a real race) _might_ happen even with normal subscribes, but actually I think because of the way Twisted or async programming works (single-threaded run-to-completion), and given our current code, this can't happen.

oberstet:
Here is a log trace (with the blowing up in AB changed to logging at warn level):

```console
2017-04-18T16:11:57+0200 [Container   31912] backend joined: SessionDetails(realm=<realm1>, session=1939439143236387, authid=<FY74-MTRN-W573-VNAR-Y9Y3-FURW>, authrole=<backend>, authmethod=anonymous, authprovider=static, authextra=None, resumed=None, resumable=None, resume_token=None)
[...]

PR #1051, opened 2017-04-26
Ticket944 meta events

meejah:
Addresses #944 

This works around the race-y "sending an Event before you've finished subscribing" by doing all the wamp meta-event publishes in "the next trip around the reactor loop" `callLater(0, ..)`. Thus, the ack for subscribing etc goes out before the meta-event.

Note that this also means you'll immediately get a meta-event for your own thing. That is, if you're subscribing to `wamp.subscription.on_subscribe` (whether exact-match, wildcard or prefix) you'll immediately get an "on_subscribe" (because your session just subscribed).

I made a short example in the Autobahn repo; should I push that, too?

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -157,13 +157,15 @@ class Broker(object):
 
                 # publish WAMP meta events
                 #
-                if self._router._realm:
-                    service_session = self._router._realm.session
-                    if service_session and not subscription.uri.startswith(u'wamp.'):
+                if self._router._realm and self._router._realm.session:
+
+                    def _publish():
+                        service_session = self._router._realm.session
                         if was_subscribed:
                             service_session.publish(u'wamp.subscription.on_unsubscribe', session._session_id, subscription.id)
                         if was_deleted:
                             service_session.publish(u'wamp.subscription.on_delete', session._session_id, subscription.id)
+                    self._reactor.callLater(0, _publish)
 
             del self._session_to_subscriptions[session]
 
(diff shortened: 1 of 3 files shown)


oberstet:
Another approach would be: use subscriber exclusion on the meta events published to skip the currently subscribing client. The WAMP spec does not talk about whether such "self-triggered" meta events are received or not by the triggering session ..

meejah:
@oberstet per your comment, you mean: if a subscriber doesn't want the "immediate meta-event" for their own session, they should exclude themselves when subscribing? (Or do you mean: the include/exclude processing isn't happening in this case and needs to be fixed?)

meejah:
Okay the last commit uses `exclude=` from the router meta-events so the subscribing session doesn't get its own meta-event immediately.

meejah:
Note: personally, I think it makes more sense to let *all* the events through -- that is, remove the last commit I just made -- because otherwise there's no way to get *actually-all* of the events and that seems the most-consistent with the semantics of matching on `""`

oberstet:
self._reactor.callLater is not needed if the meta-event subscribing client is excluded itself ..

meejah:
Hmm, right. I did just do a quick test and that does work (i.e. I just called `_publish()` instead of the callLater). I guess this would keep some pressure off the timer queue in Twisted...

Although it seems most-correct to give the client **all** the messages (i.e. including their own) it's also hard to imagine *why* you'd need that...shall I change this to: a) not use callLater and b) keep using exclude?

oberstet:
Keeping exclude and removing calllater, yes exactly. I do get your arguments, I just think the former is simple / straightforward, and we should take this Q to WAMP proto cause the beh. is missing in the spec. If we then decide the beh. you descr. is more appropr., we can add it then. In general, every callLater is always making me think twice, exactly because of what you refer to with the pressure. Sorry, I am on the road, cant paint the whole picture as I see it.Am 09.06.2017 7:11 nachm. schrieb meejah <notifications@github.com>:Hmm, right. I did just do a quick test and that does work (i.e. I just called _publish() instead of the callLater). I guess this would keep some pressure off the timer queue in Twisted...
Although it seems most-correct to give the client all the messages (i.e. including their own) it's also hard to imagine why you'd need that...shall I change this to: a) not use callLater and b) keep using exclude?

oberstet:
superseded by https://github.com/crossbario/crossbar/pull/1103

Issue #1000, opened 2017-03-23
HTTP/2 support in Web transports


oberstet:
Crossbar.io should already support HTTP/2 for Web transports by the virtue of recent Twisted supporting HTTP/2.

* Test that this is indeed the case.
* Document that

> Note that we don't need actual unit tests or such here, as this is Twisted business.

oberstet:
Twisted Web claims support for HTTP/2 since 16.3 (https://twistedmatrix.com/trac/ticket/7460).

Support for HTTP/2 is enable automatically when H2 is installed (from `from twisted.web.http2 import H2Connection`):

```
>>> from twisted.web import http
>>> http.H2_ENABLED
False
>>> 
```

So everything that is needed for HTTP/2 support would be adding a new dependency (`pip install h2`).

oberstet:
Here is more info also about testing this stuff https://pawelmhm.github.io/python/twisted/http2/python3/2016/07/30/twisted-http2.html

meejah:
Hmm, this might not be quite ready for prime-time? With the changes from above branch, I get this when a connection is timed-out:

```
Timing out client: IPv4Address(TCP, '127.0.0.1', 37605)
2017-04-26T18:30:19-0600 [Router      21355] Unhandled Error
Traceback (most recent call last):
  File "/home/mike/work-tavendo/src/crossbar/crossbar/worker/process.py", line 287, in <module>
    run()
  File "/home/mike/work-tavendo/src/crossbar/crossbar/worker/process.py", line 275, in run
    reactor.run()
  File "/home/mike/work-tavendo/venv/local/lib/python2.7/site-packages/twisted/internet/base.py", line 1243, in run
    self.mainLoop()
  File "/home/mike/work-tavendo/venv/local/lib/python2.7/site-packages/twisted/internet/base.py", line 1252, in mainLoop
    self.runUntilCurrent()
--- <exception caught here> ---
  File "/home/mike/work-tavendo/venv/local/lib/python2.7/site-packages/twisted/internet/base.py", line 878, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/home/mike/work-tavendo/venv/local/lib/python2.7/site-packages/twisted/protocols/policies.py", line 735, in __timedOut
    self.timeoutConnection()
  File "/home/mike/work-tavendo/venv/local/lib/python2.7/site-packages/twisted/web/http.py", line 2009, in timeoutConnection
    self.loseConnection()
[...]

PR #1052, opened 2017-04-27
Ticket1000 h2 support

meejah:
Using this, I am able to get Firefox to do http2 connections to a crossbar instances Web pages. You have to use HTTPS. See #1000

It seems .. odd that we have to pass "h2" explicitly to `CertificateOptions`. Also, the latest h2 (3.0.1) doesn't work -- I *think* it's just "too new for Twisted" but I didn't look very deeply.

### DIFF
--- a/crossbar/twisted/endpoint.py
+++ b/crossbar/twisted/endpoint.py
@@ -213,6 +213,12 @@ def _create_tls_server_context(config, cbdir, log):
 
     # create a TLS context factory
     # see: https://twistedmatrix.com/documents/current/api/twisted.internet.ssl.CertificateOptions.html
+    # XXX can we get CertificateOptions to do this for us?
+    from twisted.web import http
+    protocols = [b'http/1.1']
+    if http.H2_ENABLED:
+        protocols = [b'h2', b'http/1.1']
+
     ctx = CertificateOptions(
         privateKey=key,
         certificate=cert,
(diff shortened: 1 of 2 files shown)


Lukasa:
The newest h2 is too new for Twisted. Support for the 3.0 release comes in the next Twisted release though, and that Twisted release *does not* support the h2 2.0 release series (they have some mutually incompatible APIs). Just an FYI.

crossbar/twisted/endpoint.py
```diff
@@ -213,6 +213,12 @@ def _create_tls_server_context(config, cbdir, log):
 
     # create a TLS context factory
     # see: https://twistedmatrix.com/documents/current/api/twisted.internet.ssl.CertificateOptions.html
+    # XXX can we get CertificateOptions to do this for us?
+    from twisted.web import http
+    protocols = [b'http/1.1']
+    if http.H2_ENABLED:
+        protocols = [b'h2', b'http/1.1']
```
Lukasa: This shouldn't be necessary: CertificateOptions checks based on the factory that creates the protocol, see `IProtocolNegotiationFactory` and [this code](https://github.com/twisted/twisted/blob/trunk/src/twisted/web/server.py#L787-L797). If you're using a custom protocol factory, you may want to implement `IProtocolNegotiationFactory` instead.

meejah: Okay, good! I was hoping that was wrong, but couldn't get things to work otherwise. I'll look into the factory interface. (And yes, we're doing a bunch of custom stuff in here so that's probably it).

Thanks for the feedback :)

Issue #974, opened 2017-02-27
Configurable timeout for Web connections


oberstet:
/twisted/web/http.py:        log.msg("Timing out client: %s" % str(self.transport.getPeer()))

oberstet:
This affects transports of type `web` and `unisocket`, not `websocket`. It also affects the HTTP bridge.

oberstet:
We are using `twisted.web.server.Site` (see [here](https://github.com/crossbario/crossbar/blob/master/crossbar/worker/router.py#L903))

PR #1060, opened 2017-04-28
configurable web timeouts

meejah:
add 'client_timeout' option, pass it to Site and make HTTPChannel less chatty

see #974


oberstet:
Awesome: two birds one stone;)

Issue #430, opened 2015-09-02
Dynamic authorizers should get details


oberstet:
Say you want everyone who happens to know or guess the number `0634534534` to be allowed to subscribe to `chat.channel.0634534534`, but disallow everyone from doing a prefix-based subscribe on `chat.channel.`.

In principle, this is a perfect example of why we have dynamic authorizers:

``` python
@wamp.register('com.example.authorize')
def custom_authorize(session, uri, action, options):
   # your custom authorization logic to determine whether client
   # session should be allowed to perform action on uri
   if (action == 'publish' or (action == 'subscribe' and options.get('match', 'exact') == 'exact'))
        and uri.startswith(u'chat.channel.'):
      try:
          int(uri.split('.')[2])
          return True # allow action
      except ValueError:
          return False
   else:
      # deny action
      return False
```

There is only 1 catch: the `options` are currently NOT forwarded to the authorizer (the 4th argument is missing).

matthewtorgerson:
Is there currently any plan to implement this feature?  Not being able to tell if a subscription request is using exact or prefix matching makes the dynamic authorizer a little limited.

meejah:
This has been implemented in #1086

PR #1086, opened 2017-05-19
Ticket430 dynamic auth method

meejah:
This adds a new arg to authorizer methods (see #430) and also handles backwards-compatibility so at least existing authorizer methods will still work (except with an additional round-trip to the authorizer after detecting the error). Depends also on associated Autobahn PR: https://github.com/crossbario/autobahn-python/pull/840

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -301,7 +301,7 @@ class Broker(object):
 
             # authorize PUBLISH action
             #
-            d = self._router.authorize(session, publish.topic, u'publish')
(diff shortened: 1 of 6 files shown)


meejah:
All of the "abtrunk" builders pass; this can't build fully cleanly here until there's an autobahn-python release with the corresponding `.marshal_options` changes (or, obviously, simultaneous releases).

oberstet:
superseded by https://github.com/crossbario/crossbar/pull/1102

Issue #280, opened 2015-03-26
Pattern-based subscriptions and their authorisation configuration


deceze:
I'm not entirely sure whether this is a bug or just lacking documentation: I can't get publications to work with pattern-based authorisation. Meaning, I am trying to publish to topics such as:

``` python
self.publish('foo.bar.baz.%s.something-happened' % baz.id)
```

With subscribers listening to:

``` python
self.subscribe(handler, 'foo.bar.baz..something-happened', SubscribeOptions('wildcard'))
```

This works fine, however I can't figure out whether the router supports limiting permissions for wildcard topics. E.g.:

``` yaml
- { uri: foo.bar.baz..something-happened, publish: true }
- { uri: foo.bar.baz..something-else-happened, subscribe: true }
```

This doesn't seem to work. `foo.bar.baz.*.something-happened` doesn't seem to work either. All publications are equally being rejected with `session not authorized to publish to ..`.

oberstet:
Yes, wildcards in permissions configuration isn't implemented currently: https://github.com/crossbario/crossbar/blob/master/crossbar/router/role.py#L101

deceze:
But it's coming, right? :o)

While I'm at it, may I ask for some clarification? Is the wildcard syntax explicitly looking for `..` as the placeholder? Or is this just a convention and some sort of longest-substring matching is applied? What if I was separating my event names like `foo-bar-baz` instead and wanted a `foo--baz` wildcard? Why not an explicit `foo.*.baz` wildcard syntax?

oberstet:
WAMP specifies only 2 special characters for URIs: `.` and `#`. The latter is strictly for router internal use and must not appear within URIs. The `.` separates an URI into components. And URI components must not be empty - unless it is not an URI but a URI wildcard pattern.

Wildcard matching is based on matching components. And since empty components is disallowed in normal URIs, we use empty URI components to denote a wildcard component.

We could also use a dedicated, special wildcard symbol like `*` - but this then would need to be reserved (like `.` and `#`).

If you would like to see this, then we would first need to have that in the WAMP spec .. hence we needed an issue on the WAMP repo ..

oberstet:
And yes, we want permissions to allow using URI patterns as well.

deceze:
[...]

PR #1092, opened 2017-06-02
wildcard perms

meejah:
See #280 -- this is a stab at enabling wildcard permissions in static auth configurations.

"exact" permissions are always preferred. If a "wildcard" matches, it is preferred over a "prefix".


meejah:
Ah, tests on non-master configs are failing because this depends on https://github.com/crossbario/autobahn-python/pull/841 which isn't in 17.5.1

oberstet:
ok, only the pinned ones fail - but this is expected. so merging

Issue #299, opened 2015-04-12
Call timeouts and canceling of calls


oberstet:
1. timeout (call|options.timeout > 0)
2. interrupt invocations (INTERRUPT message)
3. active cancel (CANCEL message)

oberstet:
connected:

* #1382
* #1377

noisyboiler:
Hiya @oberstet 
Can we have an update on this issue please as it has been open for some time.
I'm specifically interested in whether Crossbar now implements the [call timeouts](https://wamp-proto.org/_static/gen/wamp_latest.html#call-timeouts).
Thanks

oberstet:
@meejah do you happen to know where we are on this? I seem to remember that you worked on call timeouts in the past? not sure .. maybe I'm mixing it up

oberstet:
also see https://github.com/crossbario/crossbar/pull/1694

meejah:
We have both timeouts and (active) cancelling now, so I'm closing this. If there are bugs in the implementations probably new tickets are better anyway.

PR #1111, opened 2017-06-22
Implement call cancelation #299

mbonneau:


oberstet:
@mbonneau thanks a lot Matt, looks perfect!

PR #1116, opened 2017-06-24
INTERRUPT invocation when caller goes away (if supported)

mbonneau:


oberstet:
Awesome! Highly welcome.

PR #1121, opened 2017-07-01
Change snap confinement to classic

om26er:
I was running the Django [example](https://github.com/crossbario/crossbar-examples/tree/master/django/realtimemonitor) and found out that crossbar was not able to start because it did not find django in the environment. The reason for that is with "strict" confinement, the snap is only able to look for paths within its own container, which is bad for tools like crossbar.

The fix was to use the classic [confinement](https://snapcraft.io/docs/reference/confinement), which allows the package to be able to read paths outside it space.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -8,7 +8,7 @@ description: |
   on your app's features.
 
 grade: devel
-confinement: strict
+confinement: classic
 
 apps:
   crossbar:


oberstet:
> the snap is only able to look for paths within its own container

This is exactly what we want.

The Django app thing should be run in its own snap, talking to CB over (local) network, possibly Unix domain socket.

Now, if the app snap contains static Web assets (HTML, JS, .. plain files), and these are to be hosted by the Web server built into CB, then of course `strict` confinement is a problem.

How can CB access the static Web assets left by the app snap?

Is that what `content` interfaces are for? https://snapcraft.io/docs/reference/interfaces

om26er:
> This is exactly what we want.

Good then, I was coming from the view of running the current django example with the snap, so thought that was problematic, now it makes sense.

> Now, if the app snap contains static Web assets (HTML, JS, .. plain files), and these are to be hosted by the Web server built into CB, then of course strict confinement is a problem.
> 
> How can CB access the static Web assets left by the app snap?
> 
> Is that what content interfaces are for? https://snapcraft.io/docs/reference/interfaces

We can share content between snaps using that interface, though there are some things that we still need to clear.

The content interface is generally based on the consumer/provider mechanism e.g. a framework being the provider and the "app" being the consumer of that framework, this is pretty simple case.

Our case is a bit complex than that because the consumer is actually going to be crossbar snap, even though the Django app snap is going to "connect" itself to crossbar.

I'll close this PR now.

oberstet:
My current understanding of how interfaces could work for us:

1. CB is a provider of (usually) one interface: a WAMP router listening endpoint app snaps can connect to)
2. CB may be a consumer of 0 or more content interfaces

So for 2., the app snap would need to declare that it provides a content interface.

And CB would need to declare that it uses the content interfaces of the app snap(s).

However, as both providing and using interfaces is declared in snapcraft.yaml, I still don't see how this could work, as the dependency is now circular.

And further, the CB snap would already be installed, and can't refer to any specific (user) app snap that would provide a content interface.

Mmh.

Issue #1063, opened 2017-05-02
Use h2 3.0 after next Twisted release


oberstet:
Please see comments here: https://github.com/crossbario/crossbar/pull/1052

oberstet:
We still have h2<3.0 in min requirements (https://github.com/crossbario/crossbar/blob/master/requirements-min.txt#L5)

Is that still needed after above was merged?

meejah:
It looks from the logs like Twisted 17.5.0+ should support h2 3.0+

oberstet:
fixed, we now use `h2>=3.0,<4.0`

https://github.com/crossbario/crossbar/blob/master/requirements-min.txt#L8

PR #1125, opened 2017-07-06
Python 3: fall back to native shutil.which

yegorich:
Try to import shutilwhich module, if it fails we're most likely
running Python 3.3+, so we can import the native module.

### DIFF
--- a/crossbar/controller/process.py
+++ b/crossbar/controller/process.py
@@ -34,9 +34,12 @@ import os
 import sys
 import signal
 from datetime import datetime
-# backport of shutil.which
-import shutilwhich  # noqa
-import shutil
+try:
+    # try backport of shutil.which first
+    from shutilwhich import which
+except ImportError:
+    # try Python 3.3+ native module
+    from shutil import which
 
 from twisted.internet.error import ReactorNotRunning
 from twisted.internet.defer import Deferred, DeferredList, inlineCallbacks, returnValue
@@ -869,7 +872,7 @@ class NodeControllerSession(NativeProcessSession):
         else:
             # try to detect the fully qualified path for the guest
             # executable by doing a "which" on the configured executable name
-            exe = shutil.which(config['executable'])
+            exe = which(config['executable'])
             if exe is not None and check_executable(exe):
                 self.log.info("Using guest worker executable '{exe}' (executable path detected from environment)",
                               exe=exe)


lelit:
Don't know the policy wrt Py2/3 compatibility in crossbar, but wouldn't it be better swapping the order, trying the stdlib import first?

oberstet:
this is already in the upstream library: https://github.com/mbr/shutilwhich/blob/master/shutilwhich/__init__.py

yegorich:
My intention was to get rid of shutilwhich package dependency for Python 3.3+. This package is not available in Debian at all and in Buildroot we've enabled it for Python2 only.

crossbar relies on enum34 and ipaddress, that are not needed in recent Python versions either. What about declaring them in requirenments.txt as follows:

enum34>=1.1.6; python_version < '3.4'
ipaddress>=1.0.18; python_version < '3.3'

Unrelated question: is this correct, that only h2 older 3.0 can be used?

oberstet:
The dependencies of Crossbar.io are fixed (version pinned and hashed), and this works great for us. I don't care about Debian packaging: we don't support that.

Buildroot is another thing (we do want to support that) .. how is Crossbar.io built for that? Because we used pinned/hashed deps, I wonder how that works in that environment ..

Regarding h2, ah yeah, could catch, pls see https://github.com/crossbario/crossbar/issues/1063

yegorich:
Frankly speaking crossbar support in BR is still a WIP. Every time I try to bump crossbar version I have to add a bunch of new packages :-) Getting rid of shutilwhich was long on my TODO list.

As for pinned and hashed versions, I think the syntax lets you combine hashes with Python version dependency. I've followed this [example](https://stackoverflow.com/questions/29222269/is-there-a-way-to-have-a-conditional-requirements-txt-file-for-my-python-applica/35614580#35614580). Is there any reason against specifying python_version < '3.4'?

In BR we have either Python 2 or 3 environment and we try to reduce unneeded dependencies (enum34, ipaddress and shutilwhich are available only for Python 2 environment).

oberstet:
@yegorich could you please file an issue like "buildroot support", and dump all the itches you have there? we do want to be good citizens to buildroot, can't promise, but I am willing to look into this and possible change stuff we do!

yegorich:
@oberstet OK. Will do more tests and report the results to a special ticket. Thanks.

meejah:
Personally this change looks harmless to me.

When installing Debian-packaged Python libraries, ``setup.py`` isn't used at all (that is, dependencies are tracked by Debian's packaging). So the pinned dependencies shouldn't be an issue, IIUC.

PR #1127, opened 2017-07-06
Fix handling of authextra for ticket + cookie case

meejah:
Further fixes to authextra handling when a ticket-authenticated session re-authenticates via a set cookie (previously, the authextra was not sent to the second session).

This could use a unit-test; does have a CTS integration test.


oberstet:
merged via https://github.com/crossbario/crossbar/pull/1144

PR #1128, opened 2017-07-06
Issue1105 concurrency and errors

### LINKED ISSUE #1105: Callee error responses do not free up concurrency
Still looking into this and can post a small repro application & configuration if necessary. But I set concurrency to 16 on my callee (happens to be round robin registered but was the only callee at the time), and after 16 occurrences of autobahn.wamp.exception.ApplicationError being raised on the callee, the router registered it as having reached capacity, despite the calls being completed and making their way back to the caller.

It looks like `callee_extra.concurrency_current` just gets decremented here:
https://github.com/crossbario/crossbar/blob/8076324355bf9e01e74b6e8c00b8a1f6fba86c7e/crossbar/router/dealer.py#L701

And without being familiar with the code at all, it seems like it would need to be decremented elsewhere for errors and the like, perhaps in here.
https://github.com/crossbario/crossbar/blob/8076324355bf9e01e74b6e8c00b8a1f6fba86c7e/crossbar/router/dealer.py#L725

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46239978-callee-error-responses-do-not-free-up-concurrency?utm_campaign=plugin&utm_content=tracker%2F462544&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F462544&utm_medium=issues&utm_source=github).
</bountysource-plugin>

meejah:
This should fix #1105

### DIFF
--- a/crossbar/router/dealer.py
+++ b/crossbar/router/dealer.py
@@ -842,6 +842,12 @@ class Dealer(object):
             #
             invocation_request = self._invocations[error.request]
 
+            # if concurrency is enabled on this, an error counts as
+            # "an answer" so we decrement.
+            callee_extra = invocation_request.registration.observers_extra.get(session, None)
+            if callee_extra:
+                callee_extra.concurrency_current -= 1
+
             if error.payload is None:
                 # validate normal args/kwargs payload
                 try:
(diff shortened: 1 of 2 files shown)


oberstet:
Ah, right. Good catch.

PR #1142, opened 2017-07-21
Outstanding invoke tracking

meejah:
The IDs of "Call" messages are only supposed to be unique per-session -- this fixes "outstanding call" invoke tracking to be the 2-tuple `(session_id, request)` to fix this behavior.


meejah:
I didn't file a separate bug for this, but the bug is actually fairly easy to trigger since the IDs in question increment from 1 per-session, so if you have a bunch of sessions that all start around the same time it'd be pretty easy to trigger the bug.

PR #1148, opened 2017-07-26
Tor support

meejah:
This adds `onion` (for listening) and `tor` (for connecting) endpoint types. Also documentation.

I added `txtorcon` as a dependency (and upgraded plus pinned all the dependencies).


meejah:
Note: this leaves "running a Tor daemon" as an exercise to the reader; we could later add something that runs a private-to-crossbar Tor instance if this makes sense. I expect most people interested in running an Onion service can at least follow the Tor Project Debian/Ubuntu installation instructions (which will then work with this easily).

oberstet:
Awesome! The type name `onion` is cool for me. Also cool that we both support listening and connecting transports! Nice.

PR #1152, opened 2017-08-02
Remove misc files from data_files

doctaweeks:
Using data_files causes LICENSE,etc to be installed in prefix which could clobber another file. Using only MANIFEST will include it in the tarball without this risk.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,6 @@ setup(
         ]},
     packages=find_packages(),
     include_package_data=True,
-    data_files=[('.', ['COPYRIGHT', 'LICENSE', 'LICENSE-FOR-API'])],
     zip_safe=False,
 
     # http://pypi.python.org/pypi?%3Aaction=list_classifiers


oberstet:
> Using only MANIFEST will include it in the tarball without this risk.

The AGPL requires that a user gets a copy of the license text. By including the files in `data_files`, we make sure the user has a copy of the license installed next to the software.

doctaweeks:
This change still meets the AGPL requirement. Please review your use of data_files to ensure it does not install directly into sys.prefix. (It sounds like this patch in addition to using package_data is appropriate for your goal.)

oberstet:
The license files must be included even when packaged as a wheel, or any other non-source package (tarball). We cannot use package_data, because the files referenced cannot reside outside the package directory, which they do in our repo. But we certainly don't recommend installing Crossbar.io into the host system. A virtualenv or Docker is way better of course.

Yamakaky:
I agree, it's pretty annoying to have these files auto-created

PR #1162, opened 2017-08-14
Refactor workers

meejah:
Depends upon https://github.com/crossbario/autobahn-python/pull/880

This re-factors three things:

 - use `prefix=` AB feature to get rid of `self.PROCS` and lists of procedures
 - move a bunch of router's transport/endpoint things to be bare helper methods
 - refine concept of `entrypoints=` discoverable worker types

### DIFF
--- a/crossbar/common/process.py
+++ b/crossbar/common/process.py
@@ -34,7 +34,7 @@ import gc
 
 from datetime import datetime
 
-from twisted.internet.defer import DeferredList, inlineCallbacks, returnValue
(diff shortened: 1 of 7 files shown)


meejah:
(None of this will pass CI until we can depend on an Autobahn release with https://github.com/crossbario/autobahn-python/pull/880)

oberstet:
This looks good! I will push an AB release ..

oberstet:
Ok, AB is released, and this PR is rebased to master here: https://github.com/crossbario/crossbar/pull/1164

PR #1175, opened 2017-09-02
 Node config startup hooks

meejah:
This re-factors some of the "start from config" logic to make it possible for subclasses of `Node` to have a chance to do startup-things for any custom native worker types they may have.


oberstet:
awesome! great job. lots of (necessary) refactoring.

PR #1184, opened 2017-09-05
snap: read version from central place

om26er:
The snap version is now read from crossbar/__init__.py


oberstet:
nice! and kudos to your unix foo

PR #1189, opened 2017-09-15
Container exit option

meejah:
This allows configuring of the exit behavior of containers (introduced in b66528217acdefaec152b2ca166029a1ec48db6a).

Probably the default should actually be `shutdown-on-last-worker-exit` to match old behavior (and Fabric nodes can override the default somehow).

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -2689,7 +2689,15 @@ def check_router_options(options):
 
 
 def check_container_options(options):
-    check_native_worker_options(options)
+    check_native_worker_options(options, extra_options=['shutdown'])
+    valid_modes = [u'shutdown-manual', u'shutdown-on-last-component-stopped']
+    if 'shutdown' in options:
+        if options['shutdown'] not in valid_modes:
+            raise InvalidConfigException(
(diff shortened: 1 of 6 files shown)


oberstet:
> Probably the default should actually be shutdown-on-last-worker-exit to match old behavior (and Fabric nodes can override the default somehow).

Yeah, that would actually maintain backward compatibility in behavior, which is good. We can override in CF, yep, that will work.

oberstet:
So I merged all your other work to my branch, and merged that PR to here - but now there are conflicts with this one =(

meejah:
I think the best we could do is make a note in the tracing options that there's a tiny chance of that happening if you have both tracing and chunking turned on, *and* the last chunk happens to all vanish at the "same" time.

We could also probably guard against this a bit more by special-casing it so that if the last chunk will be "small" (less than 1/4 the chunk-size?) then we just include it in the second-last batch...

I'll look at the conflicts and re-push this.

meejah:
Conflicts-wise I just skipped my `correlation -> correlation_id` patch since that was just there to get unit-tests to pass (and since other stuff got merged, don't need that fix anymore).

PR #1191, opened 2017-09-15
use deepcopy

oberstet:


oberstet:
mmh. actually, this also seems wrong: https://travis-ci.org/crossbario/crossbar/jobs/275906598#L1016 .. or maybe the fake router? dunno ..

oberstet:
I could however run this large trace https://gist.githubusercontent.com/oberstet/a1f1e6e2d9add56f3429a26c26a83e80/raw/16ea93712997863b8c8686cdcf1fa2f796aeb32f/gistfile1.txt - in any case, there is _something_ ;)

crossbar/router/broker.py
```diff
@@ -554,9 +556,7 @@ def on_authorize_success(authorization):
                             # all the event messages are the same except for the last one, which
                             # needs to have the "is_last" flag set if we're doing a trace
                             if self._router.is_traced:
-                                last_msg = message.Event.parse(msg.marshal())
-                                # XXX why do I have to copy all this stuff myself? not in marshal()
-                                # on purpose?
+                                last_msg = copy.deepcopy(msg)
                                 last_msg.correlation_id = msg.correlation_id
```
meejah: Hmm, does this really work? If so, seems like probably a bug in `.marshal()`? (That is, marshal shouldn't return anything that can magically mutate later IMO).

oberstet: actually, both should work I'd say. I don't understand why `message.Event.parse(msg.marshal())` triggers a problem - and hence, I'm not sure about this either.

meejah: The parse(marshal()) is erroring because: the ctor takes `args` as a list or a tuple, but parse insists on a list -- don't see this other places because it goes through the serializer first which turns the tuple into a list.

PR #1198, opened 2017-09-23
crossbar keygen

### LINKED ISSUE #1197: A 'crossbar keygen' command
...to make it easier to generate keys for 'crytosign' or 'cryptobox' use within Crossbar

meejah:
fixes #1197

### DIFF
--- a/crossbar/controller/cli.py
+++ b/crossbar/controller/cli.py
@@ -28,7 +28,7 @@
 #
 #####################################################################################
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import argparse
 import click
@@ -806,6 +806,22 @@ def run_command_upgrade(options, **kwargs):
         sys.exit(0)
 
 
+def run_command_keygen(options, **kwargs):
+    """
+    Subcommand "crossbar keygen".
+    """
+
+    try:
+        from autobahn.wamp.cryptobox import KeyRing
+    except ImportError:
+        print("You should install 'autobahn[encryption]'")
+        sys.exit(1)
+
+    priv, pub = KeyRing().generate_key()
+    print('  private: {}'.format(priv))
+    print('   public: {}'.format(pub))
+
+
 def run(prog=None, args=None, reactor=None):
     """
     Entry point of Crossbar.io CLI.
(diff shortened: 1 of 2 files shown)


oberstet:
mmh, not sure, I would have thought we could reuse the machinery we already have for generating an initial node key pair from an explicit `crossbar keygen` command and/or have the same being simply done automatically upon `crossbar init`, like with this outcome:

```
(cpy362_3) oberstet@thinkpad-t430s:~/test2$ cat .crossbar/key.pub 
Crossbar.io node public key

creator: oberstet@thinkpad-t430s
created-at: 2017-09-23T17:18:27.717Z
machine-id: 7a3357418649414dbd536c31c6bd0ee8
public-key-ed25519: 8065d87ff5e5d3a89f869acd0ec8c42c0fc236a0a7690177df20f1933fcd164a
(cpy362_3) oberstet@thinkpad-t430s:~/test2$ cat .crossbar/key.priv 
Crossbar.io node private key - KEEP THIS SAFE!

creator: oberstet@thinkpad-t430s
created-at: 2017-09-23T17:18:27.717Z
machine-id: 7a3357418649414dbd536c31c6bd0ee8
public-key-ed25519: 8065d87ff5e5d3a89f869acd0ec8c42c0fc236a0a7690177df20f1933fcd164a
private-key-ed25519: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
(cpy362_3) oberstet@thinkpad-t430s:~/test2$ 
``` 

not sure I get the intent of the PR here ..

meejah:
For auth keys, the internal node-keys may be used, yes.

But for end-2-end crypto uses there's no easy way to generate appropriate keys -- I added the signing keys to the command for completeness.

meejah:
Also, yeah I should probably be using crossbar's ``Key.generate`` to make the keypairs...

meejah:
Maybe it would be better to turn this more-explicitly into a "generate keys for e2e crypto" command ..? e.g. maybe include the URI as well so it can spit out keys for originator + responder and dump a code-snippet for how to construct each side's ``autobahn.wamp.cryptobox.Key`` and add it to ``KeyRing``?

meejah:
Okay, I've stripped this down to just do keygen for the "cryptobox" use-case (i.e. keypairs for URIs using `autobahn.wamp.cryptobox.KeyRing`). Does this make sense? (Motivation is: don't want "everyone" to have to discover running ``.generate_key()`` and hence write a tiny program).

Issue #1202, opened 2017-10-18
Crossbar v17.9.2 Memory Leak with Anonymous Authentication


rrueth:
Hello,

Output of `crossbar version`:
```
     __  __  __  __  __  __      __     __
    /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/

 Crossbar.io        : 17.9.2 (Crossbar.io COMMUNITY)
   Autobahn         : 17.9.3 (with JSON, MessagePack, CBOR, UBJSON)
   Twisted          : 17.9.0-EPollReactor
   LMDB             : 0.93/lmdb-0.9.18
   Python           : 3.5.2/CPython
 OS                 : Linux-4.4.0-1035-aws-x86_64-with-Ubuntu-16.04-xenial
 Machine            : x86_64
 Release key        : RWRj+Sb8UH/pwLCycZ+3OroR5UV9aaLUDVbKxNcj+gYoOOZv0LrzMH1a
```

We have a Crossbar instance running in our system. We also have a job that runs frequently (multiple times an hour) that creates a lot of short connections to the Crossbar instance. Each session is created using Anonymous Authentication. We've noticed that over time, Crossbar continued to use more and more memory until we had to restart the host after about ten days.

I restarted our Crossbar instance with debug logging enabled this morning. Then, I was able to connect to our running Crossbar instance with `pyrasite` and used `objgraph`, `gc`, and `tracemalloc` to diagnose the issue.  Using `objgraph`, I could see that the number of sets being used continued to increase throughout the day:

```
>>> objgraph.show_growth()
set                 231936   +168223
Traceback              446      +446
Statistic              262      +262
StatisticDiff          184      +184
tuple                12792      +109
function             17094       +75
_Traces                 27       +27
Snapshot                27       +27
weakref               5663       +16
member_descriptor      791       +11
```

Then, using `tracemalloc`, I was able to track down that `crossbar/router/router.py:160` seemed to continue growing all day:
```
/opt/crossbar/lib/python3.5/site-packages/crossbar/router/router.py:160: size=27.8 MiB (+1733 KiB), count=74147 (+7921), average=394 B
  File "/opt/crossbar/lib/python3.5/site-packages/crossbar/router/router.py", line 160
    self._authid_to_sessions[details['authid']] = set([session])
  File "/opt/crossbar/lib/python3.5/site-packages/crossbar/router/session.py", line 767
[...]

PR #1203, opened 2017-10-19
Stop retaining old authids and empty sets in router.

rrueth:
When all of the given sessions had disconnected for a given authid, the authid and empty set were
being held in a map. This meant that when Anonymous Authentication was used, you could have lots of
authids and empty sets being kept around, but the authid would be unlikely to be used again. This
commit removes the authid and empty set from the map, allowing that memory to be garbage collected.

Addresses #1202

### DIFF
--- a/crossbar/router/router.py
+++ b/crossbar/router/router.py
@@ -173,6 +173,8 @@ class Router(object):
         Internal helper.
         """
         self._authid_to_sessions[details['authid']].discard(session)
+        if not self._authid_to_sessions[details['authid']]:
+            del self._authid_to_sessions[details['authid']]
         self._authrole_to_sessions[details['authrole']].discard(session)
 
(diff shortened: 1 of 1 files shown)


OOPMan:
Looks good to me

rrueth:
@oberstet, who would be a good person from the Crossbar team to review and merge this change?

rrueth:
Thanks, @meejah!

meejah:
Hmm, 'something' failed on master (python3.6 only) after this merge -- but I can't repeat it locally.

meejah:
(But: build was green here before the merge ... weird)

meejah:
it was "sort order in sets"; I've got a PR in for that -- this PR was Almost Certainly not the cause

rrueth:
Thanks for finding the issue, @meejah! Good to know it likely wasn't due to this change.

PR #1204, opened 2017-10-24
Batched trace edgecase

meejah:
This fixes an edgecase of the batching tracing dispatching that was revealed after an unrelated merge; fixes also the test-case to use `OrderedSet` so it reliably (would have) repeated the failure.


meejah:
@oberstet this fixes the master build-breakage (which happens just because of some random set-sorting stuff). Essentially just moving (more of) the filtering outside of the "notify some chunks" loop to catch a case where the last chunk becomes empty (because it consists of only the "publishing" session).

PR #1216, opened 2017-11-05
Implement auth settings for longpoll connections

### LINKED ISSUE #1205: Web transport + longpolling does not support authentication?
I'm looking to add dynamic authentication to an app I'm building, which is currently configured to use a web transport and a `longpoll` endpoint.  I'm going through the various auth examples, and trying to figure out where to add the auth options.  Digging through the code in [`checkconfig.py`](https://github.com/crossbario/crossbar/blob/274605e41d0cbe789b25583aaa37e7741e550b9c/crossbar/common/checkconfig.py), it looks like the [`check_transport_auth()`](https://github.com/crossbario/crossbar/blob/274605e41d0cbe789b25583aaa37e7741e550b9c/crossbar/common/checkconfig.py#L565) function is only called for websocket endpoints, not for longpolling.  

Based on that, it looks like it's not currently possible to add authentication to a longpoll setup.  Is there something I'm missing here?

markerikson:
Fixes #1205 .

Modified config validation to accept an `auth` entry in `longpoll` sections
Passed auth options into WampLongPollResource
Updated WAMP transport and Session classes to reference auth settings if available

### DIFF
--- a/crossbar/router/longpoll.py
+++ b/crossbar/router/longpoll.py
@@ -289,6 +289,11 @@ class WampLongPollResourceSession(Resource):
         self._serializer = transport_details['serializer']
         self._session = None
 
+        if(hasattr(self._parent, '_auth')):
+            # Create a config entry so that the auth settings are picked up by the session
+            self._config = {'auth' : self._parent._auth}
+
+
         # session authentication information
         #
         self._authid = None
@@ -583,7 +588,8 @@ class WampLongPollResource(Resource):
                  queueLimitBytes=128 * 1024,
                  queueLimitMessages=100,
                  debug_transport_id=None,
-                 reactor=None):
+                 reactor=None,
+                 auth=None):
         """
         Create new HTTP WAMP Web resource.
 
@@ -605,6 +611,8 @@ class WampLongPollResource(Resource):
         :type debug_transport_id: str
         :param reactor: The Twisted reactor to run under.
         :type reactor: obj
+        :param auth: Optional authentication settings for the WAMP connection
+        :type auth: obj
         """
         Resource.__init__(self)
 
@@ -621,6 +629,7 @@ class WampLongPollResource(Resource):
         self._killAfter = killAfter
         self._queueLimitBytes = queueLimitBytes
         self._queueLimitMessages = queueLimitMessages
+        self._auth = auth
 
         if serializers is None:
             serializers = []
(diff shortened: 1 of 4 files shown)


markerikson:
Looks like the Travis check didn't like the use of extra blank lines.  If I can fix that, could we get this merged in?

crossbar/worker/router.py
```diff
@@ -689,12 +689,15 @@ def _create_resource(reactor, path_config, templates, log, cbdir, _router_sessio
 
         path_options = path_config.get('options', {})
 
+        auth = path_config.get('auth')
+
```
meejah: Does path_config always have an 'auth' key? Should this be `.get('auth', None)`..?

crossbar/router/longpoll.py
```diff
@@ -289,6 +289,11 @@ def __init__(self, parent, transport_details):
         self._serializer = transport_details['serializer']
         self._session = None
 
+        if(hasattr(self._parent, '_auth')):
+            # Create a config entry so that the auth settings are picked up by the session
```
meejah: don't need `()` around the if; just `if hasattr(..)`.

Perhaps the comment should say when this would be true? What sort of parent will have `._auth`? What sort of parent *won't* have this attribute?

Is there a reason that a "config" object (or "auth=" kwarg) isn't simply passed into this `__init__` directly?

markerikson: Good questions. It's been a couple months since I looked at this, so I'll have to swing back around in the near future and refresh my memory on what I was thinking when I made these changes :)

meejah: Okay :+1:

I've not sure if I've ever delved directly into the longpoll things, but am familiar with auth and the code "in general" -- so feel free to ping me on #autobahn in Freenode or Crossbar.io Gitter: https://gitter.im/crossbario/public if you have questions etc.

meejah: (BTW, if the way things are organized currently seems "weird" -- it might be because it's weird and needs some refactoring! So please do get in touch, especially if you think there's a better way -- we're not afraid to refactor if it makes the code nicer/easier to read!)

markerikson:
It looks like this never did get merged in.  Is there anything blocking it from being merged?

haizaar:
I understand there is still no auth support for longpolling transport, right?

PR #1246, opened 2018-02-13
WIP: reduce pkg_resources usage

om26er:


crossbar/worker/router.py
```diff
@@ -607,7 +606,8 @@ def _create_resource(reactor, path_config, templates, log, cbdir, _router_sessio
                 raise ApplicationError(u"crossbar.error.invalid_configuration", emsg)
             else:
                 try:
-                    static_dir = os.path.abspath(pkg_resources.resource_filename(path_config['package'], path_config['resource']))
+                    static_dir = os.path.abspath(os.path.join(os.path.dirname(
+                        importlib.import_module(path_config['package']).__file__), path_config['resource']))
                 except Exception as e:
```
meejah: Can we use `pkgutil.get_data()` here, as described here http://pyinstaller.readthedocs.io/en/latest/spec-files.html#using-data-files-from-a-module

crossbar/controller/cli.py
```diff
@@ -94,26 +95,37 @@
...
+    if isinstance(name_or_module, str):
+        name_or_module = importlib.import_module(name_or_module)
+
+    try:
+        return name_or_module.__version__
+    except AttributeError:
+        return ''
+
+
 def get_node_classes():
```
meejah: I like replacing all the "get a package's version" places with a call to a single function (i.e. this one)

meejah:
I think we should enumerate what the 'entry-points' stuff is trying to accomplish, and how it currently does that. The general goal is that "something that is not Crossbar" can add a "personality" type and thus change what class is instantiated (i.e. "community" vs. "fabric" is the current concrete use-case).

Also, are we SURE that we *can't* just use `pkg_resources.iter_entry_points` when using PyInstaller...?

crossbar/worker/router.py
om26er: I looked deeper into this and found that pkutil.get_data() is only used to get data files and **not** directories, any such attempt throws a IsADirectoryError exception. But the good thing is pkg_resources.resource_filename() is working fine for pyinstaller provided `datas`, so I reverted those changes.

om26er:
> I think we should enumerate what the 'entry-points' stuff is trying to accomplish, and how it currently does that. The general goal is that "something that is not Crossbar" can add a "personality" type and thus change what class is instantiated (i.e. "community" vs. "fabric" is the current concrete use-case).

The currently proposed code in this PR is just a glue to get this all working and is not a solution and we should not land this PR with that. I understand the use of entry-points here and proposed to use one of the "recommended" ways of creating plugins https://packaging.python.org/guides/creating-and-discovering-plugins/#using-namespace-packages

If we don't want to go that route, then I can look into your suggestion elsewhere http://twistedmatrix.com/documents/current/api/twisted.plugin.html

crossbar/worker/process.py
```diff
@@ -40,7 +40,7 @@
 __all__ = ('run',)
 
 
-def get_argument_parser(parser):
+def get_argument_parser(parser=argparse.ArgumentParser()):
     parser.add_argument('--reactor',
```
meejah: No, we shouldn't do this: this default will only be instantiated ONCE, which is not what you intend I don't think. Instead, say `parser=None` and in the function itself `if parser is None: parser = argparse.ArgumentParser()`; this way you'll get a new parser each time `get_argument_parser()` is called -- it's probably only called once here, but..

crossbar/worker/router.py
meejah: Ah, okay, good to know.

oberstet:
No, this is not public api. Users pluggin in their own classes is an explicit non-goal. This is private apiAm 14.02.2018 11:10 nachm. schrieb Omer Akram <notifications@github.com>:
I think we should enumerate what the 'entry-points' stuff is trying to accomplish, and how it currently does that. The general goal is that "something that is not Crossbar" can add a "personality" type and thus change what class is instantiated (i.e. "community" vs. "fabric" is the current concrete use-case).

The currently proposed code in this PR is just a glue to get this all working and is not a solution and we should not land this PR with that. I understand the use of entry-points here and proposed to use one of the "recommended" ways of creating plugins https://packaging.python.org/guides/creating-and-discovering-plugins/#using-namespace-packages
If we don't want to go that route, then I can look into your suggestion elsewhere http://twistedmatrix.com/documents/current/api/twisted.plugin.html

crossbar/worker/process.py
om26er: If this function is called without an argument, a new parser is created (old logic, that was happening previously as well, though inside run()) and if we provide a parser as an argument, it uses that, allowing code reusability for cli.py

meejah: No, a new parser is only created ONCE, at import time. So if you call it a second time, you'll get the same parser again.

oberstet:
@om26er can we split this PR up? rgd fallout from removing pkg_resources (due to pyinstaller) touches these areas:

1. reading package/module versions => you have a fix for this in this PR, and that could be merged right away!
2. WAMPlets => deprecate that altogether (feature removed)
3. reading Jinja2 template files, node release key, .. static files from within the package
4. import user modules (eg in container workers, when user specifies the name of the class to be loaded in the container)
5. the "plugin" stuff (class overriding) for CF/CFC

---

I think 5. deserves some thought rgd design .. for that, I will comment on https://github.com/crossbario/crossbar-fabric-package/issues/35

om26er:
> @om26er can we split this PR up? rgd fallout from removing pkg_resources (due to pyinstaller) touches these areas:
> 
>     reading package/module versions => you have a fix for this in this PR, and that could be merged right away!
>     WAMPlets => deprecate that altogether (feature removed)
>     reading Jinja2 template files, node release key, .. static files from within the package
>     import user modules (eg in container workers, when user specifies the name of the class to be loaded in the container)
>     the "plugin" stuff (class overriding) for CF/CFC
> 
> I think 5. deserves some thought rgd design .. for that, I will comment on crossbario/crossbar-fabric-package#35

That makes sense, I was waiting before things were "agreed". I'll split this into multiple PRs.

om26er:
@oberstet https://github.com/crossbario/crossbar/pull/1249 please review and merge.

oberstet:
ok, cool! other 2 PRs landed. I'll close this one. next would be these 2:

* reading Jinja2 template files, node release key, .. static files from within the package
* import user modules (eg in container workers, when user specifies the name of the class to be loaded in the container)

The last one (the "plugin" stuff (class overriding) for CF/CFC) is more tricky: I will comment in https://github.com/crossbario/crossbar-fabric-package/issues/35

om26er:
> reading Jinja2 template files, node release key, .. static files from within the package

as discussed with @meejah -- We don't need to make any change for this one. We only need to add "data" files using PyInstaller and then pkg_resources will work as is.

regarding pkgutil.get_data() <-- this only returns complete file path and NOT directory. So we can't use it anyways as our code depends on "directories" instead of files.

So I'll get item 4 and wait for your comments on 5.

om26er:
@oberstet I still have the work in here that added a new option to crossbar command line to be able to start a worker, shall that have an option to specify the name of the worker to start ?

PR #1247, opened 2018-02-15
Use router from cls not import

masell:
looser coupling in routerfactory

### DIFF
--- a/crossbar/router/router.py
+++ b/crossbar/router/router.py
@@ -490,7 +490,7 @@ class RouterFactory(object):
             if arg in realm.config.get('options', {}):
                 setattr(options, arg, realm.config['options'][arg])
 
-        router = Router(self, realm, options, store=store)
+        router = self.router(self, realm, options, store=store)
 
         self._routers[uri] = router
         self.log.debug("Router created for realm '{uri}'", uri=uri)


masell:
I dont see how the tests can fail on this patch.
Setting self.router to Router:
https://github.com/crossbario/crossbar/pull/1247/files#diff-0e77a2d272403b3dca5b32474584c159R402

using self.router instead of Router:
https://github.com/crossbario/crossbar/pull/1247/files#diff-0e77a2d272403b3dca5b32474584c159R493

meejah:
Yeah, this looks good to me. Have you signed a CAA? https://github.com/crossbario/crossbar/blob/master/legal/individual_caa.pdf See also: https://github.com/crossbario/crossbar/blob/master/docs/pages/about/Contributing-FAQ.md

It seems the builds are failing because `coverage html` doesn't like the invalid python file (that's used for a test) so definitely not from this change. I wonder why master build didn't fail?

oberstet:
Yeah, as meejah mentioned, we need a signed CAA before merging.

Note: this feature (being able to override classes within crossbar) is NOT a user feature. You should consider it an internal implementation details of crossbar, we don't guarantee anything (can break anytime, be removed anytime). Of course, technically, we cannot stop anyone using it;) However, note the AGPL license of crossbar will require to comply with that (a custom class needs to be published under AGPL again ..).

masell:
> Why do we need this agreement?
> 
> Without this agreement, you issue a license to your code to anybody who uses the project as is, i.e. under the current license.
> 

Im fine with the above.

Also, Im not going to sign anything.

oberstet:
@masell "not sign anything": thats cool for me! just note the requirements of the AGPL 3.0 in that case, should you use a modified version of Crossbar.io.

masell:
@meejah 

> I wonder why master build didn't fail?

Yeah, this is worrying.

meejah:
The likely reason that master didn't fail is that "coverage" released a new version that exhibits this bug (having to load all .py files, despite syntax errors) and when master was built it didn't have the bug. Nevertheless, that should be explored :)

oberstet:
@masell pls remove the following package immediately from PyPi as it violates our trademarks and copyrights: https://pypi.python.org/pypi/crossbar-modified/17.12.1

@goeddea ^

goeddea:
@masell - thanks for removing this!

To clarify: Our objection is to you using the name "Crossbar.io" for  a forked version of the software. Your rights under the AGPL are a separate matter.

PR #1255, opened 2018-02-27
Add commandline option to start a worker

om26er:


crossbar/common/reloader.py
```diff
@@ -52,7 +52,7 @@ def get_module_path_and_mtime(m):
     :type m: obj
     """
     res = (None, None)
-    if m and hasattr(m, '__file__') and (m.__file__.endswith('.py') or m.__file__.endswith('.pyc')):
+    if m and hasattr(m, '__file__') and m.__file__ and (m.__file__.endswith('.py') or m.__file__.endswith('.pyc')):
```
om26er: @meejah @oberstet added this change because it seems the package `zope` contains the `__file__` attribute, though its value is None.

So without this change crossbar is failing to start.

meejah: Hmm, is that a "namespace package" thing? Also could probably simplify above a bit with `getattr(m, '__file__', None)` or similar. Also `.pyo` is a valid ending but maybe we're ignoring those on purpose for some reason?

om26er: yeah that makes sense, pushed the change.

For the latter, not really sure.

PR #1265, opened 2018-03-09
Wamp scram prototype

meejah:
See https://github.com/wamp-proto/wamp-proto/issues/135 and https://github.com/crossbario/autobahn-python/pull/956

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -524,6 +524,16 @@ def check_transport_auth_cryptosign(config):
         raise InvalidConfigException('logic error')
 
 
+def check_transport_auth_scram(config):
+    """
+    Check a WAMP-SCRAM configuration item.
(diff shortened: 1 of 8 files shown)


oberstet:
This looks cool! The new code is nicely factored into files, and it's very easy to follow: one new file, and 3 changed ones .. for a complete new authentication method. Neat.

meejah:
There's one thing to consider before merging this and that is dependencies. (Also, it doesn't do anything with channel-binding at all yet).

This does depend on `passlib` for the `saslprep` function (similar to the Autobahn req). Do we: leave it up to "others" to install `autobahn[scram]` or `autobahn[all]` to get that, or always install passlib ourselves, or ...?

(Also, if a requirement *is* added here, do I *just* add it to `requirements-min.txt` and you do the pinning before a release, or what?)

PR #1273, opened 2018-03-19
fixing crash when watching a directory...

kneufeld:
When a file changed in a watched directory the following exception occurred

  File "crossbar/controller/process.py", line 849, in on_filesystem_change
    self._node._reactor.callLater(worker.watch_timeout, self.start_guest, worker_id, worker_config, details)
builtins.AttributeError: 'NodeControllerSession' object has no attribute 'start_guest'

It seems like there was an api name change and this one call got missed.

### DIFF
--- a/crossbar/controller/process.py
+++ b/crossbar/controller/process.py
@@ -846,7 +846,7 @@ class NodeControllerSession(NativeProcessSession):
                         if options['watch'].get('action', None) == 'restart':
                             self.log.info("Filesystem watcher detected change {fs_event} - restarting guest in {watch_timeout} seconds ..", fs_event=fs_event, watch_timeout=worker.watch_timeout)
                             # Add a timeout large enough (perhaps add a config option later)
-                            self._node._reactor.callLater(worker.watch_timeout, self.start_guest, worker_id, worker_config, details)
+                            self._node._reactor.callLater(worker.watch_timeout, self.start_worker, worker_id, worker_config, details)
                             # Shut the worker down, after the restart event is scheduled
                             # FIXME: all workers should have a stop() method ..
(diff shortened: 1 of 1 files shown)


meejah:
Thanks for the PR!

Also, CC @om26er is this fallout from recent PyInstaller-required changes?

Even though this is a trivial change I think you'll still need to go through the minor formality of signing an "Individual CAA" (https://github.com/crossbario/crossbar/blob/master/legal/individual_caa.pdf) because Crossbar is GPL.

meejah:
Or, @oberstet I see a "CAA Unneeded (minor change)" label, do tiny changes like this not need a CAA?

meejah:
Also there appears to be a general problem with the AppVeyor setup (those failures are not related to this PR).

om26er:
> Also, CC @om26er is this fallout from recent PyInstaller-required changes?

No, it seems the regression was introduced here https://github.com/crossbario/crossbar/commit/b66528217acdefaec152b2ca166029a1ec48db6a#diff-28aad7f67e74d04a75f6904bb7a26034L802

om26er:
A wild guess based on the commit I pasted above, instead of renaming the call to self.start_worker, it should be self._start_guest_worker, no ?

kneufeld:
I tried both `self._start_guest_worker` and `self.start_worker` and they both work but I guessed that calling the non-dunder better as `start_worker` does a check and calls `_start_guest_worker` but I really have no idea and have no opinion either way.

oberstet:
@meejah 

> "CAA Unneeded (minor change)" 

Correct, changes that are really minor (such as one line changes, typos) do not require a CAA. We just need to be careful about "trivial" .. @goeddea 

So this PR is "trivial" in that sense, and hence we can merge without a CAA.

PR #1296, opened 2018-04-24
policy_refactor

oberstet:
This PR is big, as it refactors the code so that the node controller and workers take their behavior and features from **policy classes**


oberstet:
superseded by https://github.com/crossbario/crossbar/pull/1299

PR #1326, opened 2018-05-15
Adding instructions to config.json for VSCode

oddjobz:
Specifically the configuration snippet to add to VSCode so it recognises files called "config.json" as Crossbar configs that should conform to crossbar.json. (i.e. removes the need for the $schema at the top of the .json file)


oberstet:
oh, that's neat! no schema line ...

PR #1336, opened 2018-05-26
Empty commit to trigger arm64 snap build

om26er:
build.snapcraft.io now support arm64 builds (along with amd64, i386 and armhf). We could either land this and expect the build to trigger automatically OR @oberstet could open this page https://build.snapcraft.io/user/crossbario/crossbar and click 'Build manually now' button. I can't do that as it requires admin rights.


oberstet:
arm64 snaps -.- yay!

PR #1339, opened 2018-05-28
Fix Container Session Shutdown

oddjobz:
This is critical for the new CFC code when deleting a Realm (as the specific use-case) and I suspect it may be affecting applications that accesses containers via RawTCP transports, if they attempt to stop / start those containers. The missing code meant that in such a scenario "stop_component" would not pass the "yield component._stopped" line, hence the component would be left running and the "on_component_stopped" event would not be issued. I sufficient quantity there may be a leak issue, but the immediate issue is when trying to re-create a session with the same name - which in the case of the realm controller, fails.
New code seems to work Ok for CFC, I can create a realm, connect to a realm, delete a realm, then re-create a realm with the same name, and it all looks good.


crossbar/worker/container.py
```diff
@@ -322,33 +363,22 @@ def close_wrapper(orig, was_clean, code, reason):
 
                 return r
 
-            # FIXME: due to history, the following is currently the case:
-            # ITransportHandler.onClose is implemented directly on WampWebSocketClientProtocol,
-            # while with WampRawSocketClientProtocol, the ITransportHandler is implemented
-            # by the object living on proto._session
-            #
             if isinstance(proto, WampWebSocketClientProtocol):
```
meejah: Is the statement in the comment now not true?

crossbar/worker/container.py
```diff
@@ -243,6 +243,47 @@ def panic(fail, msg):
...
+                        self.log.info('Stopping Container')
+                        self.shutdown()
+                    return return_value
+                #
+                #   Assign onClose handler
+                #
+                session.onClose = partial(close_wrapper, session.onClose)
                 return session
 
             except Exception:
```
meejah: There's a LOT of code in this "try/except Exception" block; can we at least move the inner-functions out? How is the new close_wrapper here for raw-sockets only? (i.e. I don't see any code that actually checks what "session" is..?)

crossbar/worker/container.py
```diff
@@ -243,6 +243,47 @@ def panic(fail, msg):
...
+                        self.log.info('Component shutdown Ok ({})'.format(component_id))
+                    else:
+                        self.log.error('Component shutdown Errored ({})'.format(component_id))
+                    #
+                    #   Tidy up .. self.components is cleaned up in "stop_component"
+                    #
+                    component = self.components[component_id]
+                    self._publish_component_stop(component)
+                    component._stopped.callback(component.marshal())
+                    #
```
meejah: Maybe this needs a better coment? What is this callback doing etc? (Is `_stopped` a `Deferred`.?) I guess what I'm confused about is where this data from the `.marshal()` end up going?

crossbar/worker/container.py
oddjobz: Mm, I guess it is, but I took the comment in the context of an explanation of what needed fixing and why .. this code should fix the issue, and implicitly describes what was in the comment. I can leave the comment in if you like and just take out the "FIXME"?

crossbar/worker/container.py
oddjobz: Yes, you're right, it should be more granular, let me reorganise ..

crossbar/worker/container.py
oddjobz: Mm, this is pre-existing code I've just duplicated (see close_wrapper from on_connect_success) .. could probably be refactored to remove the duplication, and yes, there's lots of places where more comments would be good .. ;) .. let me see if I can refactor and comment .. but to answer your question; the problem I was seeing was this code not being called, then in stop_component a call to "yield _stopped" where stopped is "Deferred()" .. which means stop_component was effectively hanging waiting for a callback that would never come. All this is doing it signalling that the remaining component shutdown code is ok to continue .. I don't think the marshal data goes anywhere, it's just the code that was already there. Whereas it may be safe to remove, I'm not sufficiently familiar with / confident of the unit tests to remove stuff because it looks obsolete to me ..

oddjobz:
@meejah Ok, have refactored a little, how's that looking now? I'm always a little wary of refactoring or changing too much and making the PR difficult to read .. but I agree unit tests would be great, and there is certainly some scope for refactoring in places for readability.

crossbar/worker/container.py
meejah: Okay, I'd like to keep the comment then (maybe with "NOTE" instead of "FIXME"...?) but this, I recall, being *very* confusing (and also the isinstance() is weird anyway, which is what it's explaining).

crossbar/worker/container.py
```diff
@@ -283,72 +326,20 @@ def panic(fail, msg):
         def on_connect_success(proto):
             component = ContainerComponent(component_id, config, proto, None)
             self.components[component_id] = component
-
-            # FIXME: this is a total hack.
-            #
```
meejah: This is still true too, but maybe it needs a bit more of a comment: basically "all" we're doing here (in both cases) is attaching an "on close" listener; the wrapper is a hack around the fact that there's no such thing as "an on-close listener" :/

Are XXX/FIXME comments a bad use of this? I personally like keeping them around as a sort of "permission" to future people that "yes, this is weird and it would be great if it could be fixed properly" .. but maybe they're just confusing?

meejah:
Sorry, should have made all those comments "a review" maybe. Definitely easier to read now, but also I'm worried we lost some checks and comments ;) and the logging needs some changes

crossbar/worker/container.py
oddjobz: Ok, have inserted your comment here with a FIXME. I've not really understood "why" this is a hack or the difference between the wrapper and having an on-close listener .. I will at some point dig into it and find out, although if in the meantime there's something I could read that explains it more concisely that would be great .. :) .. in terms of FIXME, I guess there are different ways of working (?) .. for me FIXME indicates something broken or incomplete hence something I wouldn't expect to see in production code. If a feature is intentionally incomplete or needs a redesign or refactor, I either use a # TODO, or create a ticket. As this fix addresses specific brokenness, my automatic response was to remove what I thought was the associated # FIXME .. to indicate it's no longer 'broken' .. but I can adapt to fit the codebase :)

crossbar/worker/container.py
```diff
@@ -228,21 +228,64 @@ def start_component(self, component_id, config, reload_modules=False, details=No
...
+            if was_clean:
+                self.log.info('Close connection to component ({}) with # {}'.format(component_id, code))
+            else:
+                self.log.error('Lost connection to component ({}) with # {}'.format(component_id, code))
+            #
+            #   Report an error reason if we have one ...
+            #
+            if reason:
+                self.log.warn(str(reason))
+            #
```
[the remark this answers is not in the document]
oddjobz: Ok, in my defence, this isn't my code, it's just what was there previously .. :) .. however I'll see if I can improve it a little .. reason is only (apparently) supplied for tcp sockets and not raw sockets, which I think is the reason for the conditional ..

crossbar/worker/container.py
```diff
@@ -228,21 +228,64 @@ def start_component(self, component_id, config, reload_modules=False, details=No
...
+            self._publish_component_stop(self.components[component_id])
+            self.components[component_id]._stopped.callback({})
+            #
+            #   Report and exit ...
+            #
+            self.log.info('Container hosting ({}) components, exit mode is ({})'.format(
+                          len(self.components),
+                          self._exit_mode))
+            if not self.components and self._exit_mode == self.SHUTDOWN_ON_LAST_COMPONENT_STOPPED:
```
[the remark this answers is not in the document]
oddjobz: Yeah, works for me ... :) .. I typically don't hit return after the opening bracket which leaves me with lots of tabs or a pep8 complaint, but I like your format better .. :)

crossbar/worker/container.py
```diff
@@ -283,72 +326,20 @@ def panic(fail, msg):
...
+            # for RawSockets the onClose handler is attached to session.proto (above)
             if isinstance(proto, WampWebSocketClientProtocol):
                 proto.onClose = partial(close_wrapper, proto.onClose)
-
-            elif isinstance(proto, WampRawSocketClientProtocol):
-                # FIXME: doesn't work without guard, since proto_.session is not yet there when
-                # proto comes into existance ..
-                if proto._session:
-                    proto._session.onClose = partial(close_wrapper, proto._session.onClose)
-            else:
```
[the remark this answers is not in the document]
oddjobz: Ok, my understanding of the problem was that the "if" was happening before proto._session had been instantiated, so that test in that position was obsolete as proto_session would be None, even if a session was in the process of being created .. so the assignment now happens "in" create_session, which only happens after the session has been created ... ?

oddjobz:
My fault, I'm not familiar with the code base and some of the style, and I'm trying to juggle CF, CFC, CFC-UI, while trying to learn Vue.js, with limited success .. I'd like to look at everything in more depth, I just need a timezone with more hours in it .. ;) .. I'll get there eventually, honest! :) .. have pushed another version ..

crossbar/worker/container.py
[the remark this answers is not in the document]
meejah: There should be one listener only, but it looks from the code like in websocket cases there will be TWO listeners attached: if it's a `WampWebSocketClientProtocol` it'll get the `close_wrapper` attached to `session.onClose` (because there's no "if" there) and also get `proto.onClose` attached (right?).

Maybe there are some CTS tests that cover this, I'll look -- I recall this stuff being really hard to get right but also can't remember the details, so want to make sure we don't break behavior :)

meejah:
No need to apologize, there's definitely some complex code in here :) as far as the "wrapper" thing being (in my opinion) "a hack": the intended way to use these things is to subclass something and override the method to get the notification (e.g. `onClose`).

However, here we've just "happened upon" an object and we wish to know when its `onClose` is called -- but there's no such thing as "an onClose listener" right now (as your new comment already points out). The cleaner solution would be to use `autobahn.util.ObservableMixin` and have the `protocol` implementation notify these as appropriate -- the WAMP stuff all does this now, so that's an example of how this works.

That said, wrapping methods like this *does* work, it's just awkward and can be error-prone and hard to trace what's happening. Also the error-handling gets weird: the original code here (before your changes) is still wrong, because if the (possibly user-code overridden) `onClose` method has an error, our whole shutdown thing will be skipped -- with a listener approach, errors can be caught and logged for each listener making the code easier to trace when there's bugs.

Anyway I'll run this branch through CTS and see if anything pops out...

oddjobz:
:) Many thanks .. I kind of get the feeling I'd need a number of months working with the code before I got all the in's and out's of the best way to do things (with regards to fixing stuff like this properly), but if I can pick bits up from you guys while being productive on the UI side at the same time, I'll get there so long as you can put up with my blunt force learning .. ;)

meejah:
There seem to be a couple problems with the wrappers popping up from CTS, still looking into it..

meejah:
BTW, is there a ticket describing the actual problem? (The above descriptions talks about the solution, but not how to re-produce the original problem you're seeing)

oddjobz:
@meejah no, it popped up as an issue while writing the remove_realm code for CFC. The problem affects closing a component that's configured on a WampRawSocket .. in this instance it will effectively leak memory as it fails to shut the component down properly. The observed effect for me is that I can't re-create a realm of the same name (following a delete) because the shutdown didn't actually fully remove the old instance of the realm router). I fixed it rather than logging it because it was a blocker for the CFC stuff I'm working on .. would you like me to file a ticket for it?

meejah:
Usually I'd ask you to file a ticket ;) but I just fixed this a slightly different way (and yes, the problem was that rawsockets weren't getting cleanup handlers attached properly). See: https://github.com/crossbario/crossbar/pull/1341

oddjobz:
@meejah if it's fixed I'm happy .. :) .. did we keep the change to router.py, it was a 1-liner but also critical ..

quick question;

`del self.components[component_id]`

Is it needed? the remaining part of stop_component also does this?

meejah:
Oh, does it? Okay I'll look closer, was just copy-pasta-ing the original code. I'll put the router changes in a new PR I guess. Thanks for the catch!

oddjobz:
At the end of "stop_component" ..
```
       # the component.proto above normally already cleaned it up
        if component_id in self.components:
            del self.components[component_id]
```
one or other seemed obsolete .. :)

oberstet:
Have been reading through your conversation here: really great to see how smoothly and quickly you started to discuss one of the definitely more complex pieces of the code base already=)

@meejah I totally agree that the complexity in this case is partially due to the old-schoolish design of the WebSocket layer of ABPy. An observer pattern also for the WebSocket layer would remove _some_ of the complexity in the CB codebase here. if we could find a way to add that in ABPy that preserves backwards compatibility with existing code using ABPy at the WebSocket level (like Daphne or others), then I'd be +1 on this, but given the context for this year, I'd probably give it "medium-priority" .. details.

@oddjobz CB has some areas where the code base is definitely non-trivial;) It is expected to take months to get into all of that. Tbh, I am impressed by the speed you got into! Like, you seem to be close to where you have the complete chain from CFC-UI, CFC-API and backend, and CF/CB management side of things "under control" .. the interplay between those wheels .. which is one of those "high-complexity topics" in my view. In this case the complexity comes from the many wheels involved, and the distributed nature of the things happening.

PR #1341, opened 2018-05-30
Unify websocket vs rawsocket code-paths, fixing a rawsocket bug

meejah:
RawSocket objects would often not (never?) have their _session set
up yet so weren't getting appropriate cleanup handlers added (which
causes an internally-used `on_component_stop` handler to not fire)


oberstet:
merged via https://github.com/crossbario/crossbar/pull/1344

PR #1346, opened 2018-06-05
docs: fix environment variable example

wodCZ:
Took me some time to realize that the variable was not evaluated when specified with `()` instead of `{}`


oberstet:
uups, thanks!

PR #1351, opened 2018-06-12
Dynamic SCRAM authenticator

MickVanDuijn:
This is the basis for the dynamic SCRAM authenticator. The principals are retrieved by calling the dynamic authenticator, and the rest (client signature, server proof, etc) remains the same as when they were stored in the config file. This way, we can dynamically add users with SCRAM. The authenticator can be as simple as retrieving the correct principal from a database.

This is the extension to dynamic authenticators, building on the code of https://github.com/crossbario/crossbar/pull/1265

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -527,36 +527,36 @@ def check_transport_auth_scram(config):
     """
     Check a WAMP-SCRAM configuration item.
     """
-    for k in [u'type', u'principals']:
-        if k not in config:
-            raise InvalidConfigException(
-                "missing mandatory attribute '{}' in WAMP-SCRAM configuration".format(k)
-            )
-    if config[u'type'] != u'static':
+    if u'type' not in config:
         raise InvalidConfigException(
-            "Only static SCRAM configuration allowed currently"
+            "missing mandatory attribute '{}' in WAMP-SCRAM configuration".format(u'type')
         )
-
-    # check map of principals
-    for authid, principal in config['principals'].items():
-        check_dict_args({
-            'kdf': (True, [six.text_type]),
-            'iterations': (True, [six.integer_types]),
-            'memory': (True, [six.integer_types]),
-            'salt': (True, [six.text_type]),
-            'stored-key': (True, [six.text_type]),
-            'server-key': (True, [six.text_type]),
-            'role': (False, [six.text_type]),
-        }, principal, "WAMP-SCRAM - principal '{}' configuration".format(authid))
-        available_kdfs = (u'argon2id-13', u'pbkdf2')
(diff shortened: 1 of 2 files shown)


MickVanDuijn:
To respond to this directly:
https://github.com/Zefiros-Software/crossbar/blob/849a93dbb860f61b8d1ce5e123728483b0076990/crossbar/router/auth/scram.py#L106

I believe you can not compute it, as this would require the hashed password. The hashed password is never stored, for security reasons.

meejah:
hi, thanks for the PR!

I took a quick look, but I think some code is missing? ``_init_dynamic_authenticator`` and ``_marshal_dynamic...`` methods aren't defined?

Also, if you haven't yet, you'll need to sign a CAA (e.g. https://github.com/crossbario/crossbar/blob/master/legal/individual_caa.pdf)

MickVanDuijn:
Hi, thanks for the info! I just emailed the signed agreement.

Regarding the functions you mentioned, those are defined on the PendingAuth parent class
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/pending.py#L166
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/pending.py#L181

Multiple (if not all) other dynamic authenticators also make use of those functions, so I figured it would be good practice to do that here as well.

For `wampcra` see:
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/wampcra.py#L129
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/wampcra.py#L148

For `ticket` see:
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/ticket.py#L94
https://github.com/crossbario/crossbar/blob/977c708255fc48ef56180615c34663a7ec3f7b42/crossbar/router/auth/ticket.py#L148

meejah:
Ahh, okay sorry with my "glancing" I was expecting to see those in the new code. I'll take a closer look :)

oberstet:
@MickVanDuijn 

> I just emailed the signed agreement.

thanks! got it.

MickVanDuijn:
@oberstet Great!
@meejah No problem! Just thought I'd give you some pointers, so you have less digging to do!

MickVanDuijn:
@meejah is there anything I can do to help speed this up?

meejah:
Thanks for the new feature and sorry for the delay in reviewing!

PR #1352, opened 2018-06-12
Feature/get component

oddjobz:
Ok, this is a summary of minor tweaks;
- Two fixes for manhole;
  1. incoming credentials are binary and need to be decoded before being compared to a string
  2. private and public have been switched at some point (!)
- Add pid and startup time to node controller publication for consistency (i.e. so total of publications relating to startup deliver same basic info at get_node)
- Remove realm from internal listing, this was part of an earlier patch that seems to have gone missing ..
- Implement "get_router_component", despite being listed under registrations, it seems it was missing (!)


oberstet:
cool! rgd get_router_component missing: yeah, as you likely will have noticed by now, especially the read side of the management API (within CB) could need more love;) up to now, there wasn't much executing it, so it didn't get attention .. which is changing now .. good!

PR #1353, opened 2018-06-24
Personality auth options

meejah:
Allows Personality to add new authentication methods


oberstet:
aah, the bits missing that made the other PR incomplete .. cool! and the "personality" approach for pluggability looks sustainable .. we can add "as we need". also good.

PR #1354, opened 2018-06-26
fix web config handling

meejah:


meejah:
Ah, yes, thanks good catch! :)

meejah:
I'm pretty sure "paths" is a dict as well (at least it is in the existing config for the Autobahn example router) and in any case the code is indexing with a ``str`` (that is ``['/']``) so it's obviously expecting a dict?

meejah:
In any case ``x in foo` works whether foo is a list or a dict (or any collection) so it shouldn't really matter, but as far as "documenting" things dicts are more-correct ;)

oberstet:
yeah, dict it is .. actually, any Mapping: https://github.com/crossbario/crossbar/blob/master/crossbar/common/checkconfig.py#L1729

oddjobz:
Yeah, sorry, I was just looking at the default in the original code, looking at the surrounding code, yeah, dict .. :)

oberstet:
I am backing this out .. it breaks stuff. The code before did work for me

Issue #1357, opened 2018-07-03
async is a keyword since py 3.7


eratulrai:
Hi,
  I upgraded my crossbar installation on windows10  , after the installation finished successfully
Whenever I try to run crossbar init or version command , it throws following error:
```
PS D:\crossbar> crossbar init
Traceback (most recent call last):
  File "c:\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python37\Scripts\crossbar.exe\__main__.py", line 9, in <module>
  File "c:\python37\lib\site-packages\crossbar\__init__.py", line 172, in run
    _personalities = personalities()
  File "c:\python37\lib\site-packages\crossbar\__init__.py", line 205, in personalities
    from crossbar import personality as standalone
  File "c:\python37\lib\site-packages\crossbar\personality.py", line 42, in <module>
    from crossbar.worker.router import RouterController
  File "c:\python37\lib\site-packages\crossbar\worker\router.py", line 51, in <module>
    from crossbar.worker.controller import WorkerController
  File "c:\python37\lib\site-packages\crossbar\worker\controller.py", line 214
    def start_profiler(self, profiler=u'vmprof', runtime=10, async=True, details=None):
                                                                 ^
SyntaxError: invalid syntax
PS D:\crossbar> crossbar -version
Traceback (most recent call last):
  File "c:\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Python37\Scripts\crossbar.exe\__main__.py", line 9, in <module>
  File "c:\python37\lib\site-packages\crossbar\__init__.py", line 172, in run
    _personalities = personalities()
  File "c:\python37\lib\site-packages\crossbar\__init__.py", line 205, in personalities
    from crossbar import personality as standalone
  File "c:\python37\lib\site-packages\crossbar\personality.py", line 42, in <module>
    from crossbar.worker.router import RouterController
  File "c:\python37\lib\site-packages\crossbar\worker\router.py", line 51, in <module>
    from crossbar.worker.controller import WorkerController
  File "c:\python37\lib\site-packages\crossbar\worker\controller.py", line 214
    def start_profiler(self, profiler=u'vmprof', runtime=10, async=True, details=None):
[...]

PR #1361, opened 2018-07-07
python3.7 compat: remove use of now reserved keyword 'async'

om26er:
Release summary of python 3.7 says "async and await are now reserved keywords."

https://docs.python.org/3/whatsnew/3.7.html#summary-release-highlights

fixes https://github.com/crossbario/crossbar/issues/1357


oberstet:
ah, right, I see. wasn't too hard it seems;)

PR #1364, opened 2018-07-13
Reverse proxy str bytes

chrisbarber:


chrisbarber:
Actually, using `path` here is plain wrong and does not match the docs. Fix incoming...

oberstet:
awesome! thanks for this .. and your other PRs!

PR #1380, opened 2018-08-23
amrhf/aarch64 docker image fixes

om26er:
Fixes armhf and aarch64 builds, fixes some docker warnings, removes old cruft.
* Use latest python base images for amrhf and aarch64 builds
* Add missing libffi-dev build dependency
* Remove old x86 image Dockerfile
* Change Dockerfile(s) to read from "right" node directory


oberstet:
pls see:

* https://github.com/crossbario/autobahn-python/tree/master/docker

PR #1397, opened 2018-09-12
Adding missing publish

oddjobz:
Publish an "on_realm_stopped" event, not sure if this is the "best" place for it .. but it looks right, and I'm certainly now getting the events Ok ...

### DIFF
--- a/crossbar/worker/router.py
+++ b/crossbar/worker/router.py
@@ -312,6 +312,7 @@ class RouterController(WorkerController):
             u'detached_sessions': sorted(detached_sessions)
         }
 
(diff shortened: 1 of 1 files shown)


meejah:
(when CI goes green, please merge)

oberstet:
will certainly do the job!

it is missing exclude of caller (will work, but is different from what we do elsewhere .. but there is more cleanup to do: eg some "stop_xx" events publish on the ID of the stopped XX, others will also publish stopping detail information):

`self.publish(..., options=PublishOptions(exclude=caller))`

oberstet:
ah, our comments overlapped in time;) cool for merging "as is" for me (we can do the cleanups rgd event payload / options probably in a dedicated systematic round .. later)

PR #1398, opened 2018-09-14
correctly pass on any configured "role"

meejah:
"function"-style components have no other way to ask for a particular auth-role, and have a "role" config item which was never correctly passed on.


oberstet:
aah;)

PR #1406, opened 2018-09-26
requirements-min.txt: remove idna.

asafkahlon:
A short "git grep idna" shows there's no direct use in idna
in crossbar.

Signed-off-by: Asaf Kahlon <asafka7@gmail.com>

### DIFF
--- a/requirements-min.txt
+++ b/requirements-min.txt
@@ -1,4 +1,3 @@
-idna<2.6,>=2.5
 click>=6.7
 setuptools>=36.0.1
 zope.interface>=4.4.2


oberstet:
this is a recommended depdendency for use of TLS with Twisted. pls see the note here https://twisted.readthedocs.io/en/twisted-18.7.0/core/howto/ssl.html#connecting-to-public-servers

> To properly validate your hostname parameter according to RFC6125, please also install the “service_identity” and “idna” packages from PyPI. Without this package, Twisted will currently make a conservative guess as to the correctness of the server’s certificate, but this will reject a large number of potentially valid certificates. service_identity implements the standard correctly and it will be a required dependency for TLS in a future release of Twisted.

asafkahlon:
OK, I guess can agree with that... But is there any reason to hold the idna version under 2.6 as stated at the requirements file?

PR #1408, opened 2018-10-03
Adding Fallback Resource

oddjobz:
Ok, so I don't know if there's a better way of doing this, but this feels fairly straightforward. Essentially, if you have a static resource and add to the options;
```python
"default_file" : "index.html"
```
In the event that the resource is unable to deliver a URL, it will attempt to deliver "index.html" (from the configured 'directory') instead. So for example for a single page application, you might have;
```python
"/": {
    "type": "static",
    "directory": "../web",
    "options": {
        "enable_directory_listing": true,
        "default_file": "default.html"
    }
},
```
In this case, accessing a valid path will deliver the expected result, but any attempt to access "/" followed by an unknown word will deliver the file;
```python
/unknown => ../web/default.html
```
This should suffice for most SPA's, but if you need a directory hierarchy, you can always add;
```python
"subfolder": {
    "type": "static",
    "directory": "../web",
    "options": {
        "enable_directory_listing": false,
        "default_file": "default.html"
    }
},
```
Which will force any unknowns on the subfolder to map to the same pattern, i.e;
```python
/subfolder/unknown => ../web/default.html
```
If this is an acceptable solution, then I would envisage the same patch for the "archive" resource.

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -1298,7 +1298,8 @@ def check_web_path_service_static(personality, config):
         check_dict_args({
             'enable_directory_listing': (False, [bool]),
             'mime_types': (False, [Mapping]),
-            'cache_timeout': (False, list(six.integer_types) + [type(None)])
+            'cache_timeout': (False, list(six.integer_types) + [type(None)]),
+            'enable_fallback': (False, [str])
         }, config['options'], "'options' in Web transport 'static' path service")
 
 
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -1299,7 +1299,7 @@ def check_web_path_service_static(personality, config):
             'enable_directory_listing': (False, [bool]),
             'mime_types': (False, [Mapping]),
             'cache_timeout': (False, list(six.integer_types) + [type(None)]),
-            'enable_fallback': (False, [str])
+            'default_file': (False, [str]),
         }, config['options'], "'options' in Web transport 'static' path service")
 
 
--- a/crossbar/webservice/base.py
+++ b/crossbar/webservice/base.py
@@ -64,8 +64,8 @@ class ResourceFallback(File):
     def __init__(self, path, config, **kwargs):
         File.__init__(self, path, **kwargs)
         directory = config.get('directory', '')
-        file = config.get('options', {}).get('enable_fallback')
-        self.path = '{}/{}'.format(directory, file)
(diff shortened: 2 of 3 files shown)


crossbar/webservice/base.py
```diff
@@ -56,6 +57,17 @@ def set_cross_origin_headers(request):
...
+class ResourceFallback(File):
+    """
+    Handle requests for non-existent URL's
+    """
+    def __init__(self, path, config, **kwargs):
+        File.__init__(self, path, **kwargs)
+        directory = config.get('directory', '')
+        file = config.get('options', {}).get('enable_fallback')
+        self.path = '{}/{}'.format(directory, file)
+
```
meejah: This should use `os.path.join`

crossbar/common/checkconfig.py
```diff
@@ -1298,7 +1298,8 @@ def check_web_path_service_static(personality, config):
         check_dict_args({
             'enable_directory_listing': (False, [bool]),
             'mime_types': (False, [Mapping]),
-            'cache_timeout': (False, list(six.integer_types) + [type(None)])
+            'cache_timeout': (False, list(six.integer_types) + [type(None)]),
+            'enable_fallback': (False, [str])
         }, config['options'], "'options' in Web transport 'static' path service")
```
meejah: minor nit-pick, but it's nice to have trailing commas on stuff (so e.g. NEXT time this is changed, there's only a one-line diff ;) )

crossbar/webservice/base.py
oddjobz: yeah, absolutely, fixed :)

crossbar/common/checkconfig.py
oddjobz: No, good point, my natural reaction is that a comma is superfluous. however it's definitely much neater in PR's just to see one line rather than two ... fixed :)

oddjobz:
Re; names, sure I have no preference or opinion on naming, I find it too subjective, so I'm happy to go with whatever others are happy with .. :) .. have adjusted to "default_file" .. :)

oddjobz:
@meejah yeah, I know what you mean re; more options .. I initially looked at a bunch of permutations for routing and had to stop myself. This was on my list for the UI, and I was prompted by a query someone posted with exactly the same issue (SPA routing). In this instance there was a very concrete issue / requirement, so I just aimed to solve that one problem, there are a number of potential 'other' features possible here, but maybe wait until there's a solid use-case for them to ensure the solution is testable against something real.

PR #1431, opened 2018-11-03
Don't rely on parent context for subscription

oddjobz:
I was seeing missing "unsubscribe" events, and events for non-existent subscriptions .. looks like any client disconnection where the client has > 1 subscription was at risk of putting out incorrect unsubscribe events. This seems to fix it for me in testing ...


om26er:
Cool, this looks good to me.

meejah:
Yeah :lgtm: -- non-optimal Python loop continuation binding :/

Issue #33, opened 2014-03-13
authmethods: ["anonymous"] --> cookie-based if cookie exists


goeddea:
When a cookie is set,

`authmethods: ["anonymous"]`

results in

`authmethod: "cookie.mozilla_persona"`

being used

oberstet:
I can confirm this. It's a bug. Here is a log:

```console
AutobahnJS debug enabled
(index):13 Ok, AutobahnJS loaded 0.9.6
autobahn.min.jgz:33 trying to create WAMP transport of type: websocket
autobahn.min.jgz:33 using WAMP transport type: websocket
autobahn.min.jgz:33 WebSocket transport send [1,"realm1",{"roles":{"caller":{"features":{"caller_identification":true,"progressive_call_results":true}},"callee":{"features":{"progressive_call_results":true}},"publisher":{"features":{"subscriber_blackwhite_listing":true,"publisher_exclusion":true,"publisher_identification":true}},"subscriber":{"features":{"publisher_identification":true}}},"authmethods":["anynonymous"],"authid":"joe"}]
autobahn.min.jgz:33 WebSocket transport receive [2,4671093781727030,{"authrole":"frontend","authmethod":"wampcra","authprovider":"cookie","roles":{"broker":{"features":{"publisher_identification":true,"pattern_based_subscription":true,"subscription_meta_api":true,"subscription_revocation":true,"publisher_exclusion":true,"subscriber_blackwhite_listing":true}},"dealer":{"features":{"pattern_based_registration":true,"registration_revocation":true,"shared_registration":true,"caller_identification":true,"registration_meta_api":true,"progressive_call_results":true}}},"authid":"joe"}]
(index):49 connected session with ID 4671093781727030
(index):50 authenticated using method 'wampcra' and provider 'cookie'
(index):51 authenticated with authid 'joe' and authrole 'frontend'
autobahn.min.jgz:33 WebSocket transport send [48,1744880512532480,{},"com.example.add2",[2,3]]
autobahn.min.jgz:33 WebSocket transport receive [50,1744880512532480,{},[5]]
(index):57 call result 5
```

oberstet:
The client asks for anonymous authentication, and CB answers with `"authmethod":"wampcra","authprovider":"cookie"`

oberstet:
This is how a log looks now (client wants to authenticate as anyonmous, which isn't even allowed by the node config.):

```console
AutobahnJS debug enabled
[...]

PR #1434, opened 2018-11-06
no cookie auth if not in authmethods #33

markope:
If cookie is not in the authmethods given by the connecting client, then no cookie auth should be executed. 

The proposed solution allows the cookie auth initially on transport level but in the context of the onHello of the session the auth is reevaluated. If "cookie" is not in the given authmethods, the already given auth will not be transfered to the session level, but will be revoked. In this case the cookie will also be invalidated. If the following auth process goes through however, then the cookie will be revalidated with the then correct role. 

I think this is closer to the behaviour expected by the user. 
One detail remains:
It is a little unexpected that authentication methods will be tried in order only cookie auth is considered first regardless of the position in the authmethods array. This should be documented or fixed too. 

Note, that there is another way to invalidate a cookie, which could be used as a workaround too:
`session.leave('wamp.close.logout')` will invalidate the cookie according to [here](https://crossbar.io/docs/Ticket-Authentication/)

### DIFF
--- a/crossbar/router/session.py
+++ b/crossbar/router/session.py
@@ -617,9 +617,16 @@ class RouterSession(BaseSession):
 
             # if the client had a reassigned realm during authentication, restore it from the cookie
             if hasattr(self._transport, '_authrealm') and self._transport._authrealm:
-                assert u'cookie' in authmethods
-                realm = self._transport._authrealm
-                authextra = self._transport._authextra
+                if u'cookie' in authmethods:
+                    realm = self._transport._authrealm
+                    authextra = self._transport._authextra
+                else:
+                    # revoke authentication and invalidate cookie (will be revalidated if following auth is successful)
+                    self._transport._authmethod = None
+                    self._transport._authrealm = None
+                    self._transport._authid = None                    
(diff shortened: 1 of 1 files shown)


oberstet:
> If cookie is not in the authmethods given by the connecting client, then no cookie auth should be executed.

yes, that is how crossbar should behave, and apparently it does not:(

the problem is: this change would break all other transport-level authentication methods: eg consider client TLS certificate based authentication. this is also pre-authenticating already at the transport-level (before WAMP kicks in, before onHello). the change works by erasing the pre-auth info (from transport), but does not consider where the transport level info came from (eg from a cookie, from TLS, from ..).

so a fix that is confined to the situation you care about (cookie etc) needs to erase the transport level auth info, but only for `self._authprovider = u'cookie'`

https://github.com/crossbario/crossbar/blob/master/crossbar/router/protocol.py#L276

actually it would be good to have an issue filed first describing the expected and the observed behavior. and referencing this PR. one more reason for this is: we need to nail the docs on this _now_ too - as the docs should carry the agreed expected behavior (identical to observed after the fix of course).

PR #1436, opened 2018-11-09
Adding new test to check unsubscribe/detach fix

oddjobz:
This is a unit test change to go with; https://github.com/crossbario/crossbar/pull/1431
This test fails prior to the fix, and succeeds with the fix applied.

I'm afraid this took quite a while for me to get my head around, I'm not sure if this pattern is ideal moving forward, but for now it should be good enough to pick up on the edge case in question. Specifically I'm testing subscribe/unsubscribe (which was on anyway) and subscribe / detach (which was where the issue was).

### DIFF
--- a/crossbar/router/test/test_broker.py
+++ b/crossbar/router/test/test_broker.py
@@ -34,6 +34,7 @@ from twisted.trial import unittest
 
 import txaio
 import mock
+import time
 
 from autobahn.wamp import types
 from autobahn.wamp import message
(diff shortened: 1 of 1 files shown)


crossbar/router/test/test_broker.py
```diff
@@ -511,6 +512,197 @@ class TestSession(ApplicationSession):
...
+                            deletes.remove(args[2])
+
+                    test.assertEqual(len(created), 0, 'incorrect response sequence for on_create')
+                    test.assertEqual(len(subscribes), 0, 'incorrect response sequence for on_subscribe')
+                    test.assertEqual(len(unsubscribes), 0, 'incorrect response sequence for on_unsubscribe')
+                    test.assertEqual(len(deletes), 0, 'incorrect response sequence for on_delete')
+
+                reactor.callLater(1, all_done)
+                time.sleep(1)
+
```
meejah: This stuff is what `Clock` is for: faking time. So we don't want to wait a real second, we want to tell the (fake) reactor: go forward in time 1 second. For example, like this: https://github.com/crossbario/autobahn-python/blob/master/autobahn/websocket/test/test_websocket.py#L183

crossbar/router/test/test_broker.py
```diff
@@ -511,6 +512,197 @@ class TestSession(ApplicationSession):
...
+                    subscribes = list(subscriptions)
+                    unsubscribes = list(subscriptions)
+                    deletes = list(subscriptions)
+
+                    for args, argv in self._service_session._events:
+                        if args[0] == 'wamp.subscription.on_create':
+                            test.assertEqual(args[1], self._service_session._session_id, 'on_create: session id is incorrect!')
+                            test.assertTrue(args[2]['id'] in created, 'on_create: subscription id is incorrect!')
+                            created.remove(args[2]['id'])
+
```
meejah: This sort of pattern always makes me wonder: should this be like 4 separate tests instead? (Maybe not, but ..)

crossbar/router/test/test_broker.py
```diff
@@ -511,6 +512,197 @@ class TestSession(ApplicationSession):
...
+    def test_subscribe_detach(self):
+        """
+        Make sure the wamp.* event sequence delivered as part of a subscribe/detach sequence
+        is correct and contains the correct session id and subscription id values.
+        """
+        router = self.router
+        test = self
+
+        class Session(mock.MagicMock):
+            """
```
meejah: Does this need to subclass Mock? (i.e. are there other method-calls that we're expecting other than the ones in this fake Session?). If so, it might be a nicer pattern (and more explicit/obvious) to do something like:
```
class Session(object):
    some_method = Mock()
```

...or similar.

crossbar/router/test/test_broker.py
```diff
@@ -511,6 +512,197 @@ class TestSession(ApplicationSession):
...
+            pub's we are expecting.
+            """
+            _private = []
+            _events = []
+            _authrole = 'trusted'
+            _session_id = 0
+
+            def __init__(self, *args, **kwargs):
+                super().__init__(*args, **kwargs)
+
```
meejah: Remove this `__init__`, it's not doing anything (I mean the whole thing, not just the super call_

oddjobz: Yeah, it was, but no longer, forgot to remove .. :)

crossbar/router/test/test_broker.py
oddjobz: Yeah the sleep thing is interesting, take out the sleep and it fails .. however, I've just noticed that if I reduce the callLater to zero, and remove the sleep it works .. :)

crossbar/router/test/test_broker.py
meejah: Okay, fair enough.

crossbar/router/test/test_broker.py
meejah: Yeah, there's .. probably a lot of methods. I guess it just looked weird to me to subclass it, but amounts to about the same thing .. so, good like this :)

meejah:
oh, CI is sad because of the unused `time` import

oddjobz:
time import fixed :)

oberstet:
@oddjobz adding new unit (or integration) tests: awesome!! in particular such a tricky one that is covering an important feature.

PR #1441, opened 2018-11-23
This is a reworking of Pull#1354

oddjobz:
I'm not sure how 1354 breaks elsewhere, but the current code doesn't work with a number of use-cases including demo.crossbar.io. This patch does work with demo.crossbar.io and my local examples .. whereas I could see that 1354 could break something, I'm hoping this one can't ...

### DIFF
--- a/crossbar/worker/transport.py
+++ b/crossbar/worker/transport.py
@@ -384,6 +384,8 @@ class RouterWebTransport(RouterTransport):
         # create root web service
         if '/' in self._config.get('paths', []):
             root_config = self._config['paths']['/']
+        elif '/' in self._config.get('web', {}).get('paths', {}):
+            root_config = self._config['web']['paths']['/']
         else:
             root_config = {'type': 'path', 'paths': {}}
         root_factory = self._worker.personality.WEB_SERVICE_FACTORIES[root_config['type']]


crossbar/worker/transport.py
```diff
@@ -384,6 +384,8 @@ def _create_web_factory(self, create_paths=False, ignore=[]):
         # create root web service
         if '/' in self._config.get('paths', []):
             root_config = self._config['paths']['/']
+        elif '/' in self._config.get('web', {}).get('paths', {}):
+            root_config = self._config['web']['paths']['/']
         else:
```
meejah: Maybe outside the scope of this PR, but I wonder if `RouterWebTransport` should "know" if it's a Universal or plain Web transport? (i.e. the above would allow a "web" -> "paths" thing in a plain-web config -- but hopefully that's caught earlier at config-validation time?)

meejah: (I think that was maybe the problem with 1354: I didn't see that this class supports two different things)

oddjobz: Mmm, don't know .. for now I was just looking to fix the immediate problem .. moving forward, it might be interesting to explore the possibility of replacing checkconfig with automatic json schema validation .. i.e. we now have a json schema for the configuration (albeit it's not completely up-to-date or indeed complete), but it could be used to implicitly validate a config file without the need for actual code. This would force the schema to be kept up-to-date and make editing config files 100% verifiable in the editor. (assuming you use something with schema validation built-in, i.e. vscode or similar) Might make it easier to see what the config structure should look like to the code ...

PR #1451, opened 2018-12-19
Make snap sharable to be consumed by WAMP based components

om26er:
Make use of the [content interface of snapcraft](https://forum.snapcraft.io/t/the-content-interface/1074). This allows other snaps to use crossbar without shipping their own copy of crossbar.


om26er:
Landing this now because a) it doesn't change any python code and b) I am blocked to test something.

PR #1452, opened 2018-12-23
wamp.session.get serialization

meejah:
Fixes https://github.com/crossbario/autobahn-python/issues/1096

### DIFF
--- a/crossbar/newsfragments/1096.bugfix
+++ b/crossbar/newsfragments/1096.bugfix
@@ -1,0 +1,1 @@
+fixed serialization error using wamp.session.get
--- a/crossbar/router/service.py
+++ b/crossbar/router/service.py
@@ -238,7 +238,17 @@ class RouterServiceAgent(ApplicationSession):
         if session_id in self._router._session_id_to_session:
             session = self._router._session_id_to_session[session_id]
             if not is_restricted_session(session):
-                return session._session_details
+                details = session._session_details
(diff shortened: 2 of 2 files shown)


oberstet:
ah, this should use the marshaling code there is already, as in `return session._session_details.marshal()`

meejah:
Hmm, `SessionDetails` doesn't have a `.marshal()` currently (in `types.py` only `SessionIdent` does currently it seems) .. but I can definitely put the code there instead..

oberstet:
@meejah ah yes, you are right, the spec includes (optional) transport info https://wamp-proto.org/_static/gen/wamp_latest.html#wamp-session-get

IMO the best would be to "fix this once and for all" now by extending https://github.com/crossbario/autobahn-python/blob/master/autobahn/wamp/types.py#L326 with transport details (consolidating the info there, instead of wherever we currently put it).

We could create a new `TransportDetails` class with a `marshal` method and have a `transport` attribute on `SessionDetails` which is an instance of the former. The `marshal` of `SessionDetails` could use the `marshal` of `TransportDetails` object.

I would leave in the attributes currently not in the spec .. maybe file an issue on wamp-proto to gather attention/consensus. Overall my main concern is to consolidate/cleanup stuff ..

oberstet:
superseded by https://github.com/crossbario/crossbar/pull/1466

Issue #1453, opened 2018-12-25
"Crossbar has started event"


om26er:
I have an app that is based on crossbar, its a pluggable app that exposes functionalities of a Linux based system desktop over WAMP. This app also implements "network discovery" using [zeroconf](https://pypi.org/project/zeroconf/), currently it relies on the `node.pid` file to exist before broadcasting service name, this works sometimes but other times it sends the broadcast a little too early, causing a failed connection for clients.

We need a way to check if crossbar is fully running, something that's emitted after https://github.com/crossbario/crossbar/blob/master/crossbar/node/node.py#L482

A "hack" would be to just create an empty file inside `cbdir`, a more sophisticated solution might actually be to send some sort of event using twisted's reactor.

thoughts @meejah @oberstet ?

om26er:
I am a noob when it comes to Twisted but I proposed a minor change that could address this case.

oddjobz:
Hi, not sure if I'm totally understanding, but could you get away with something like this;

oddjobz:
Hi, not sure if I'm totally understanding, but could you get away with something like this; adding this to config.json will enable a unix domain socket;
```xml
            "transports": [
                {
                    "type": "rawsocket",
                    "endpoint": {
                        "type": "unix",
                        "path": "/tmp/uds_watch.sock"
                    }
                },
```
With a twisted client;
```python
#!/usr/bin/env python3

from twisted.internet import protocol, reactor


class Monitor(protocol.Protocol):
    
    def connectionMade(self):
        print('Call "after startup" event trigger here')

    def connectionLost(self, reason):
        print('Call "after shutdown" event trigger here')


class MonitorFactory(protocol.ReconnectingClientFactory):

    protocol = Monitor
    maxDelay = 1
    initialDelay = 1

    def clientConnectionLost(self, connector, reason):
        print('Crossbar has gone down!')
        protocol.ReconnectingClientFactory.clientConnectionLost(self, connector, reason)

    def clientConnectionFailed(self, connector, reason):
        print('Waiting for Crossbar to come up ...')
        protocol.ReconnectingClientFactory.clientConnectionFailed(self, connector, reason)


if __name__ == '__main__':
    reactor.connectUNIX('/tmp/uds_watch.sock', MonitorFactory())
    reactor.run()
```

meejah:
[...]

PR #1454, opened 2018-12-25
Emit event crossbar-ready when its fully started

om26er:
A proposal to fix https://github.com/crossbario/crossbar/issues/1453


oberstet:
this is a twisted event, not a WAMP management event, but we do want the latter - closing

Issue #1456, opened 2018-12-29
Remove dependency on six


om26er:
Seems we no longer support Python 2, so we probably need to get rid of all the `six` usage in the code base. Unless if we use it for something that I don't know ?

meejah:
Yes, we could probably remove all `six` usage (at least, I can't think of a reason why not).

om26er:
Ok, I proposed my initial effort [here](https://github.com/crossbario/crossbar/pull/1457), let the CI run and see how things are.

PR #1457, opened 2018-12-29
Remove six usage

om26er:
We no longer support Python 2 but still our code uses `six` in many places for compatibility purposes. Lets move away from six. closes https://github.com/crossbario/crossbar/issues/1456


crossbar/bridge/rest/test/__init__.py
```diff
@@ -87,10 +85,10 @@ def _utcnow():
 
 def makeSignedArguments(params, signKey, signSecret, body):
 
-    params[b'timestamp'] = [network_string(_utcnow())]
+    params[b'timestamp'] = [bytes(_utcnow(), 'latin-1')]
```
om26er: Reference on why that's using `latin-1` encoding: https://pythonhosted.org/six/#six.b

crossbar/common/twisted/endpoint.py
```diff
@@ -293,7 +292,7 @@ def _create_tls_client_context(config, cbdir, log):
         for cert_fname in [os.path.abspath(os.path.join(cbdir, x)) for x in (config['ca_certificates'])]:
             cert = crypto.load_certificate(
                 crypto.FILETYPE_PEM,
-                six.u(open(cert_fname, 'r').read())
+                str(open(cert_fname, 'r').read())
```
om26er: Not fully sure about this one though, help ?

crossbar/bridge/rest/test/__init__.py
meejah: Maybe make your comment above a comment in the code? (Because I can 100% assure you I'm going to ask that question again if I randomly look at this line ;) )

crossbar/common/twisted/endpoint.py
meejah: I am *fairly* certain that because you opened with `"r"` instead of `"rb"` you'll get unicode/str back in Python3 -- so I think you can just drop the `str(..)` entirely.

crossbar/_logging.py
```diff
@@ -274,12 +273,7 @@ def _make_json(_event):
...
             traceback = event["log_failure"].getTraceback()
 
-            if not six.PY3:
-                traceback = traceback.decode('utf-8')
-                linesep = os.linesep.decode('utf-8')
-            else:
-                linesep = os.linesep
-
+            linesep = os.linesep
             eventText = eventText + linesep + traceback
```
meejah: Maybe we can just replace `linesep` with `os.linesep` throughout this code (i.e. drop the local)?

crossbar/_logging.py
```diff
@@ -353,7 +347,7 @@ def color_json(json_str):
     Given an already formatted JSON string, return a colored variant which will
     produce colored output on terminals.
     """
-    assert(type(json_str) == six.text_type)
+    assert(type(json_str) == str)
     return highlight(json_str, lexers.JsonLexer(), formatters.TerminalFormatter())
```
meejah: Not this PR's fault, but asserts like this should really be `assert isinstance(json_str, str)` or so

crossbar/bridge/mqtt/wamp.py
```diff
@@ -65,7 +63,7 @@ def _mqtt_topicfilter_to_wamp(topic):
     Convert a MQTT topic as used in MQTT Subscribe (and hence ptoentially containing
     special characters "+" and "#") to a WAMP URI and a match policy.
     """
-    if type(topic) != six.text_type:
+    if type(topic) != str:
         raise TypeError('invalid type "{}" for MQTT topic filter'.format(type(topic)))
```
meejah: Should use `isinstance` probably...

crossbar/bridge/mqtt/wamp.py
```diff
@@ -119,7 +117,7 @@ def _mqtt_topicname_to_wamp(topic):
     """
     Convert a MQTT topic as used in MQTT Publish to a WAMP URI.
     """
-    if type(topic) != six.text_type:
+    if type(topic) != str:
         raise TypeError('invalid type "{}" for MQTT topic name'.format(type(topic)))
```
meejah: isinstance

crossbar/bridge/rest/test/__init__.py
meejah: (Oh, why isn't my other comment showing here? Anyway just in case: maybe turn the above in to a code-comment with the link please :) )

crossbar/common/checkconfig.py
```diff
@@ -290,15 +281,12 @@ def get_config_value(config, item, default=None):
     if item in config:
         # for string valued items, check if the value actually point to
         # an enviroment variable (e.g. "$PGPASSWORD")
-        if type(config[item]) == six.text_type:
+        if type(config[item]) == str:
             match = _ENV_VAR_PAT.match(config[item])
```
meejah: isinstance

crossbar/common/checkconfig.py
```diff
@@ -689,7 +677,7 @@ def check_endpoint_backlog(backlog):
     :param backlog: The backlog parameter for listening endpoints to check.
     :type backlog: int
     """
-    if type(backlog) not in six.integer_types:
+    if type(backlog) != int:
         raise InvalidConfigException("'backlog' attribute in endpoint must be int ({} encountered)".format(type(backlog)))
```
meejah: isinstance

crossbar/common/checkconfig.py
```diff
@@ -702,7 +690,7 @@ def check_endpoint_port(port, message="listening/connection endpoint"):
     :param port: The port to check.
     :type port: int
     """
-    if type(port) not in six.integer_types:
+    if type(port) != int:
         raise InvalidConfigException("'port' attribute in {} must be integer ({} encountered)".format(message, type(port)))
```
meejah: isinstance

crossbar/node/worker.py
```diff
@@ -188,7 +187,7 @@ def log(self, childFD, data):
             self._log_entries.append(repr(data))
             return
 
-        if type(data) != six.text_type:
+        if type(data) != str:
             data = data.decode('utf8')
```
meejah: isinstance

crossbar/router/auth/cryptosign.py
```diff
@@ -204,7 +202,7 @@ def authenticate(self, signed_message):
         as a challenge previously, XORed with the 32 bytes transport channel ID (if available).
         """
         try:
-            if type(signed_message) != six.text_type:
+            if type(signed_message) != str:
                 return types.Deny(message=u'invalid type {} for signed message'.format(type(signed_message)))
```
meejah: isinstance

oberstet:
overall, I'm cool with the PR .. removing six makes sense now that we only support py3 only.

@meejah what's the rationale with preferring `isinstance` vs `type` again? because py3 makes everything an object now? is it more a "stylistic" thing (which is a good reason as well), or are there real effects? just interested .. I am cool with changing to isinstance (or leaving it at `type` - for now)

crossbar/_logging.py
om26er: done, it seems there was only one instance, other places in that file are already using os.linesep

crossbar/common/twisted/endpoint.py
om26er: I looked up the upstream documentation and indeed, this expects the buffer as `bytes` https://pyopenssl.org/en/stable/api/crypto.html#OpenSSL.crypto.load_certificate, So now I am opening with `rb` and **not** converting it to a string.

crossbar/bridge/rest/test/__init__.py
om26er: cool, I removed the _utcnow() function and reused it from autobahn.

om26er:
I do have the code ready (locally) that replaces **all** instances of `type(foo) == bar` with `isinstance(foo, bar)`. One good thing is that it makes our code uniform i.e. `isinstance` in all places, currently its a mixture of both.

I think it would be better if that was proposed in the different PR, will help keep things simple.

Also for this PR, it would probably be a good idea to "squash and merge" as there are quite a few commits.

crossbar/common/twisted/endpoint.py
meejah: :+1:

om26er:
This is ready to land. The CI failure is a false flag, apparently connection to aws timeout. Is there a way to rerun https://travis-ci.org/crossbario/crossbar/jobs/476928243 ?

om26er:
Also, is it fine to do `isinstance()` changes in a separate branch or shall I push those changes to this one ?

oberstet:
I'd be cool with piggybacking `isinstance` changes on this PR as well, np. makes sense anyways, as it is logically "the same fix"

om26er:
> I'd be cool with piggybacking `isinstance` changes on this PR as well, np. makes sense anyways, as it is logically "the same fix"

sure, pushed the changes. Lets wait for the CI.

PR #1465, opened 2019-01-10
Add Ubuntu Snap installation to docs

codelectron:


om26er:
We should probably add a small section under "Installation on Linux" page on how to install Crossbar from snap. The current doc (as brought from old docs) would make it sound as if snaps are Ubuntu specific.

om26er:
Lets land this now, I'll update the snap documentation as needed.

PR #1468, opened 2019-01-12
Docker notes

oberstet:


oberstet:
doesn't touch any code, only readme and docker build stuff

Issue #1463, opened 2019-01-08
Security issue - information leak through "call" operations


haizaar:
Good day,

I have crossbar:pypy3-18.9.2 serving a realm with dynamic authorizer. For publish, subscribe and register operation attempts, request is forwarded to authorizer first and foremost (I am publishing with acks) so authorizer have a chance to return `wamp.error.not_authorized` without giving up any sensitive information.

For call operations, if no one registered for the requested procedure, `wamp.error.no_such_procedure` is returned, *without calling authorizer first*. This way a malicious client can sniff out all URIs that have a registrants behind them.

This looks like a security leak to me - I would've expected "call" operations to behave similarly to others, namely being forwarded to authorizer first.

If there is registrant behind the URI, then authorizer is called as expected, so the problem only with URIs that do not have registrants.

haizaar:
Similar issue with "register" operation - if someone else is already registered for the URI, then `wamp.error.procedure_already_exists` before consulting with authorizer.

meejah:
Yes, this does indeed sound like an information-disclosure problem. I agree an unauthorized client shouldn't be able to enumerate valid procedure URIs

oberstet:
agreed: a slight information leakage that seems unnecessary. authorization before checking for existence would fix it.

oberstet:
I've added a "bug" label, as this is security related though this "only" leaks some info - it cannot be directly exploited, since knowing the proc exists still doesn't allow to call it (as the authorizer kicks in). but of course there are other aspects. anyways. we should fix it;)

haizaar:
I agree this is not "breach"-grade issue, but it can be exploited to obtain quite sensitive information. For example if you sell appliances that connect to a WAMP portal, then each portal user can find out how many appliances in total are available, track your sales progress, etc.

oberstet:
fixed via https://github.com/crossbario/crossbar/pull/1467

haizaar:
Thanks for the quick fix!
It missed 19.1.1 release though, right?

oberstet:
yes, it missed 19.1.1. I pushed a new release that contains the fix (and a couple of others):

* https://github.com/crossbario/crossbar/tree/v19.1.2
* https://pypi.org/project/crossbar/19.1.2/
* https://snapcraft.io/crossbar

oberstet:
crossbar 19.1.2 docker images published:

* https://hub.docker.com/r/crossbario/crossbar/tags/
[...]

PR #1498, opened 2019-02-13
Ticket1463 register info leak

meejah:
missing piece of #1463


haizaar:
Thank you. When do you plan to craft a new release?

PR #1500, opened 2019-02-13
REST bridge: remove WAMP ticket auth forwarding

meejah:


markope:
That's a scary pull request. Is this really removing ticket authentication from crossbar?
We are using this feature.

oberstet:
Ah, no;) What's removed is only code that forwarded HTTP to WAMP-Ticket Auth - that code is a security problem (and the design approach unfortunately "broken by design"). we are working on a replacement for the ability to authenticate the incoming HTTP to Web services defined in crossbar ..

PR #1507, opened 2019-02-21
[1388] fix reactor forwarding to worker processes

### LINKED ISSUE #1388: Reactor configuration does not propagate to workers
Sometime before v18.7 support for the `reactor` attribute in worker configuration was dropped, and
reactor configuration for the controller was moved from command-line argument to environment variables (`CROSSBAR_REACTOR`).

However, worker processes do not run on the reactor specified via environment variable but on the default reactor (IOCP on Winidows).

asodeur:
fixes #1388

### DIFF
--- a/crossbar/node/controller.py
+++ b/crossbar/node/controller.py
@@ -503,7 +503,9 @@ class NodeController(NativeProcess):
             args.extend(['--title', options['title']])
 
         # forward explicit reactor selection
-        #
+        # TODO: would make more sense to start whatever we are currently running
+        #  on but there seems to be no way to get the current reactor's shortName
+        #  in twisted
         reactor = os.environ.get('CROSSBAR_REACTOR', None)
         if reactor:
             args.extend(['--reactor', reactor])
--- a/crossbar/node/controller.py
+++ b/crossbar/node/controller.py
@@ -502,11 +502,18 @@ class NodeController(NativeProcess):
         if options.get('title', None):
             args.extend(['--title', options['title']])
 
-        # forward explicit reactor selection
-        # TODO: would make more sense to start whatever we are currently running
-        #  on but there seems to be no way to get the current reactor's shortName
-        #  in twisted
-        reactor = os.environ.get('CROSSBAR_REACTOR', None)
+        # forward current reactor selection
+        # TODO: not a twisted expert... is it really so hard to get the current reactor's shortName?
+        #
+        def get_current_reactor_shortname():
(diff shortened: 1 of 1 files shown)


oberstet:
not sure .. this needs some deeper look before merging, because: the command line argument to the worker should be retained, but we should make sure the reactor selected in the node controller is actually forwarded as a command line argument when the worker is spawned

asodeur:
Latest version is the best `twisted` lets you do IMHO. The reactors do not know about their `shortName`s and the installers do not know about the reactors (only the module of the `install` function, see [here](https://github.com/twisted/twisted/blob/6ac66416c0238f403a8dc1d42924fb3ba2a2a686/src/twisted/application/reactors.py#L60) and [here](https://github.com/twisted/twisted/blob/6ac66416c0238f403a8dc1d42924fb3ba2a2a686/src/twisted/plugins/twisted_reactors.py)).

Under the bold assumption the `install` function will reside in the same module as the reactor impl the new version will find the `shortName`. On casual inspection the only reactor not working seems to be `iocp` but that one is broken anyway and I am loosing hope it is ever getting fixed. (Guess it shows I am slowly developping a serious aversion to `twisted` on Windows ...)

oberstet:
mmh, ok: which of the following "works" on windows these days - and which one do you actually want to use btw?

```
from twisted.application import reactors
reactors.installReactor('iocp')
```

and

```
from twisted.application import reactors
reactors.installReactor('select')
```

and

```
from twisted.application import reactors
reactors.installReactor('wfmo')
```

asodeur:
We are running the `select` reactor which is working fine for us and is supposed to be stable [see](https://twistedmatrix.com/documents/current/core/howto/choosing-reactor.html). However, AFAIK the `select` reactor has a limit on simultaneous connections <512 which might be way too low for many use-cases.

The other two reactors are still experimental.

The `iocp` reactor supposedly can handle more connections but has severe issues with high load/large messages. `test_tcp.LargeBufferWriterProtocol` of this [PR](https://github.com/twisted/twisted/pull/1025/files) is a simple and fairly reliable reproducer. Yet, this might be you're only option if you need many connections.

Last time I tried `wfmo` could not run crossbar, yet. Not sure if that has improved in the meantime.

oberstet:
Rgd the windows `iocp` problems with high load, this is now merged on twisted trunk https://github.com/twisted/twisted/pull/1025

Rgd `wfmo`: never used that as it is quite new and I left windows years ago

Issue #750, opened 2016-05-01
crossbar/worker/test/examples/syntaxerror.py causing build issue


tpetazzoni:
The file `crossbar/worker/test/examples/syntaxerror.py` apparently contains an intentional syntax error. Unfortunately, when byte-compiling, this causes an error, which is annoying for distributions or build systems that aim at checking byte-compiling issues.

Here is the issue:

```
PYTHONPATH="/home/thomas/projets/buildroot/output/target/usr/lib/python3.5/sysconfigdata/:/home/thomas/projets/buildroot/output/target/usr/lib/python3.5/site-packages/" /home/thomas/projets/buildroot/output/host/usr/bin/python3.5 support/scripts/pycompile.py /home/thomas/projets/buildroot/output/target/usr/lib/python3.5
Traceback (most recent call last):
  File "/home/thomas/projets/buildroot/output/host/usr/lib/python3.5/py_compile.py", line 125, in compile
    _optimize=optimize)
  File "<frozen importlib._bootstrap_external>", line 697, in source_to_code
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/home/thomas/projets/buildroot/output/target/usr/lib/python3.5/site-packages/crossbar/worker/test/examples/syntaxerror.py", line 33
    class # noqa
               ^
SyntaxError: invalid syntax

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/thomas/projets/buildroot/output/host/usr/lib/python3.5/compileall.py", line 145, in compile_file
    optimize=optimize)
  File "/home/thomas/projets/buildroot/output/host/usr/lib/python3.5/py_compile.py", line 129, in compile
    raise py_exc
py_compile.PyCompileError:   File "/home/thomas/projets/buildroot/output/target/usr/lib/python3.5/site-packages/crossbar/worker/test/examples/syntaxerror.py", line 33
    class # noqa
               ^
SyntaxError: invalid syntax
```

Would it be possible to find a solution that does not break byte compilation? For the moment, we have to patch this file to avoid the syntax error.

oberstet:
FWIW, this test is to test when there is a syntax error in user code that is to be started as a worker under Crossbar.io - it makes sure we properly catch that and process it (in the node controller). It is a quite important test.

Obviously, we could exclude the tests from distribution - but people will then complain about the tests missing (some distros have rules about that).

[...]

PR #1513, opened 2019-03-02
[SNAP] byte-compile python files to improve startup

om26er:
Given the read-only nature of snaps, the python cache files are not written, making each start of the snap as a "cold-start", this PR changes that and given the runtime environment of a snap is predictable and exactly the same across different systems, the change is supposed to be reliable.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,9 +1,6 @@
 name: crossbar
-# Fallback version
 version: 'latest'
-version-script: |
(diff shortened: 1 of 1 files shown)


om26er:
Ok, so given the conversation in https://github.com/crossbario/crossbar/issues/750, I guess the the approah to delete the syntaxerror.py file is the right one to do. I could also try rename that file before compilation and restore to old name after that.

oberstet:
yeah, deleting the file is just fine!

PR #1532, opened 2019-03-24
1531.cryptosign improvements

meejah:


meejah:
Note this needs https://github.com/crossbario/autobahn-python/pull/1146 as well

PR #1534, opened 2019-03-24
Docker r1932

oberstet:


om26er:
Does this mean, we can re-enable arm64 snap builds ?

oberstet:
yes, however, I'm not 100% sure the PR will work;)

python packaging is still in a semi-sad state: eg environment markers (for having vmprof as a dependency conditional to machine arch) works in requirements.txt when used from pip - but _not_ when installed from pypi via setuptools ("pip install crossbar") and hence we need this total hack here: https://github.com/crossbario/crossbar/blob/14051335af2522dca6efd82eba1d4943f25a6f1a/setup.py#L60

PR #1550, opened 2019-04-04
Don't use psutil.Process().io_counters() on MacOS

rrueth:
The [`psutil` documentation](https://psutil.readthedocs.io/en/release-3.4.2/#psutil.Process.io_counters) says that `io_counters()` is not supported on MacOS:

Return process I/O statistics as a namedtuple including the number of read and write operations
performed by the process and the amount of bytes read and written. For Linux refer to /proc
filesysem documentation. On BSD there’s apparently no way to retrieve bytes counters, hence -1 is
returned for read_bytes and write_bytes fields. OSX is not supported.

### DIFF
--- a/crossbar/common/monitor.py
+++ b/crossbar/common/monitor.py
@@ -29,6 +29,7 @@
 #####################################################################################
 
 import datetime
+import sys
 import psutil
 
 from twisted.internet.threads import deferToThread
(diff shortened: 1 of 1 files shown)


oberstet:
the fix should be done by extending/modifying the code block that exists - and not simply return from the ctor prematurely (returning in the middle of the ctor "is bad"). also, not sure: have you signed a CAA already? we would need that, but only once! sorry for annoyance ..

oberstet:
fixed via https://github.com/crossbario/crossbar/pull/1561

PR #1554, opened 2019-04-12
Better naming for nighly snap builds

om26er:
Today all daily snap builds have their version number set to last stable version of crossbar because we read from https://github.com/crossbario/crossbar/blob/master/crossbar/_version.py#L31

This PR changes that and now we use `git describe` to include the commit hash for daily builds. The format for daily builds `19.3.5+git5.g0fc8f58b`

Note: If we tag a new release and push that, then the build that is triggered won't have the git commit hash and only contain the version number, just like what we currently have, so this change will also help us distinguish between "stable" and "nightly" in snapcraft.io web interface.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,6 @@
 name: crossbar
-version: 'latest'
-version-script: python3 -c "exec(open('crossbar/_version.py').read()); print(__version__)"
+version: latest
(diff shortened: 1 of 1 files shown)


oberstet:
the CI fails were unrelated to this PR. I fixed those and now force merged this PR here - it "will work";)

Issue #413, opened 2015-08-27
WebSocket options not checked


oberstet:
The implementation of `check_websocket_options` in `checkconfig.py` is incomplete:
* lacks actual checking of options
* lacks WebSocket compression checking

Also make sure above is in sync with the docs at:

* https://github.com/crossbario/crossbardocs/blob/master/pages/docs/administration/router/transport/WebSocket-Options.md
* https://github.com/crossbario/crossbardocs/blob/master/pages/docs/administration/router/transport/WebSocket-Compression.md

PR #1580, opened 2019-05-06
more websocket options checking

meejah:
Most things were already partially checked, this increases the checking. See #413


oberstet:
looks good, merging. the sphinx failure (some version conflict) in CI is definitely unrelated ..

Issue #1563, opened 2019-04-17
Strange things happening on YIELD error


Jopie64:
Hello,

I'm working with WampSharp (backend) and autobahn.js (frontend) to connect to the crossbar router version cpy3-19.3.5 and ran into an issue when a progressive RPC was yielded from WampSharp. The router reported:

> YIELD received for non-pending request ID 180

After that it disconnected. There are 2 things I noted in this situation.

1. The client had multiple running progressive calls, but it looked like they didn't receive a response when the callee was thrown out. Hence the caller isn't aware of the callee not reacting any longer. On a normal disconnect, the caller does receive responses for the pending RPC's.
2. I traced back where Crossbar threw the callee out in the first place, and it looks like it is handling the YIELD of non-pending RPC as a protocol error:
https://github.com/crossbario/crossbar/blob/57a2393f62c32723b72cc9fcb5a26eb0a5365f57/crossbar/router/dealer.py#L1279
I think it should not handle a YIELD of a non-pending RPC as a protocol error because it can happen due to races. Warning about it is OK to me but killing the session is a bit rigorous I think.

But at least it should handle the disconnect gracefully.

meejah:
How can it happen due to races? (and, if it does, that's a bug in one of the client libraries, likely or in crossbar itself)

Usually crossbar is fairly intolerant of protocol errors, because if a client has gotten to an illegal state, "guessing" what it'll do isn't good. In this case, for example, if it is sending a `YIELD` for a request that hasn't been issued, will it get "stuck" when (if?) the actual request comes through?

Jopie64:
Example of a race:

- Consider a pending RPC
- Caller sends `CANCEL` with `killnowait`
- Dealer forwards it as `INTERRUPT` to callee, considers RPC done
- At the same time (before receiving `INTERRUPT`) callee sends `YIELD`
- `YIELD` is received while RPC is unknown by dealer

Still please also consider the first problem which is actually of bigger importance: the ungracefull disconnect of the callee without responding to the callees (who did nothing wrong) that the pending RPC's are cancelled.

oberstet:
here is more discussion that might be relevant https://github.com/wamp-proto/wamp-proto/issues/160

Jopie64:
Turns out the same problem occurs when callee reacts with `ERROR` on a cancel!

https://github.com/crossbario/crossbar/blob/cfa1feb938a7409ebae103649d1b47de29226101/crossbar/router/dealer.py#L1401

[...]

PR #1592, opened 2019-05-25
Partly fix YIELD/ERROR race

Jopie64:
Hello all,

About issue #1563, I want to, as we say in Dutch, 'Throw the bat in the hen house'. It means something like to 'stir things up' :) Because, at our company we run against this issue quite often.

I agree that the WAMP router better be strict about protocol errors. Sessions that cause them are probably buggy and might cause issues in the WAMP router when not taken care of.
But in the case of a `YIELD` or invocation `ERROR` response, it could be that, due to a race, the referenced invocation is not there anymore. (Because it was `CANCEL`'ed recently.) In that case it was not due to a protocol error and the router should just ignore these responses.

When the invocation ID was the ID of a different session, a different check kicks in and it will throw a ProtocolError exception anyway.

So since we run against this issue in our company, I hope this pull request is approved soon. Or when you have a different suggestion, feel free to offer that.

### DIFF
--- a/crossbar.json
+++ b/crossbar.json
@@ -41,7 +41,7 @@
                 4, 6
             ],
             "default": 4
-        },       
+        },
         "check_ip_host": {
             "description": "The hostname to use for this connection",
             "type":"string"
@@ -162,19 +162,7 @@
                 "max_frame_size": {
                     "description": "Maximum size in bytes of incoming WebSocket frames accepted or 0 to allow any size. (default: 0)",
                     "type": "integer",
-                    "minimum": 0
-                },
-                "require_masked_client_frames": {
-                    "description": "Require all WebSocket frames received to be masked. (default: true)",
-                    "type": "boolean"
-                },
-                "apply_mask": {
-                    "description": "Actually apply WebSocket masking (both in- and outgoing). (default: true)",
-                    "type": "boolean"
-                },
-                "max_frame_size": {
-                    "description": "Maximum size in bytes of incoming WebSocket frames accepted or 0 to allow any size. (default: 0)",
-                    "type": "integer",
+                    "minimum": 0,
                     "examples": [
                         0
                     ]
@@ -213,7 +201,7 @@
                 },
                 "flash_policy": {
                     "description": "This is new - fixme!",
-                    "type": "string"                    
+                    "type": "string"
                 },
                 "compression": { "$ref": "#/definitions/check_compression" },
                 "require_websocket_subprotocol": {
@@ -237,7 +225,7 @@
                 },
                 "endpoint": { "$ref": "#/definitions/listening_endpoint" },
                 "serializers": { "$ref": "#/definitions/check_listening_serializers" },
-                "url": { "$ref": "#/definitions/url" },                        
+                "url": { "$ref": "#/definitions/url" },
                 "options": { "$ref": "#/definitions/check_websocket_options" }
             }
         },
@@ -431,7 +419,7 @@
                 "tls": {
                     "description": "Options associated with a TLS endpoint",
                     "type": "object",
-                    "properties": {       
+                    "properties": {
                         "key": {
                             "description": "TLS Key file",
                             "type": "string"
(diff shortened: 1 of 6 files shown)


crossbar/router/dealer.py
```diff
@@ -1276,9 +1276,6 @@ def processYield(self, session, yield_):
                         if invocation_sent:
                             self._call_store.pop_queued_call(invocation_request.registration)
 
-        else:
-            raise ProtocolError(u"Dealer.onYield(): YIELD received for non-pending request ID {0}".format(yield_.request))
-
     def processInvocationError(self, session, error):
```
meejah: I feel like we should at least log this (even if only at DEBUG level?)  in case there is an underlying problem.

crossbar/router/dealer.py
```diff
@@ -1396,6 +1393,3 @@ def processInvocationError(self, session, error):
             #
             invoke = self._invocations[error.request]
             self._remove_invoke_request(invoke)
-
-        else:
-            raise ProtocolError(u"Dealer.onInvocationError(): ERROR received for non-pending request_type {0} and request ID {1}".format(error.request_type, error.request))
```
meejah: Same, I'd say this should be at least logged here..

crossbar.json
```diff
@@ -213,7 +201,7 @@
                 },
                 "flash_policy": {
                     "description": "This is new - fixme!",
```
meejah: following whitespace fixups all look good :+1:

crossbar.json
```diff
@@ -162,19 +162,7 @@
...
-                    "description": "Require all WebSocket frames received to be masked. (default: true)",
-                    "type": "boolean"
-                },
-                "apply_mask": {
-                    "description": "Actually apply WebSocket masking (both in- and outgoing). (default: true)",
-                    "type": "boolean"
-                },
-                "max_frame_size": {
-                    "description": "Maximum size in bytes of incoming WebSocket frames accepted or 0 to allow any size. (default: 0)",
-                    "type": "integer",
```
meejah: why was this config removed?

Jopie64: Because those 3 were duplicates. See commit message.

crossbar/router/dealer.py
Jopie64: Yes. I can agree with that.

Jopie64:
About keeping track of recently closed invocations:
I think that will make it harder. How long would you store this? It could be that network is slow at the moment. What if 'recently closed invocation' is already removed and then a `YIELD` is received? Protocol Error anyway? That would be bad I think. A fixed size queue: How big should the queue be? What when a lot of invocations were cancelled at roughly the same time?

About logging:
Yea I agree with you that that is a good idea. I'll fix soon.

oberstet:
cool! seems like it's now logging at level "debug" instead of bailing out on protocol error, so for me this is ready for merge=) ah, last one for merge is: flake8 is complaining https://travis-ci.org/crossbario/crossbar/jobs/537309107

Jopie64:
About this flake error: is this a new thing? I ran the tests at my PC back then and it looked like they succeeded. How can this error be fixed?

oberstet:
You can run `make autoformat` locally to fix it ..

background: flake8 isn't happy with the whitespace formatting here:

```
rossbar/router/dealer.py:1281:17: E128 continuation line under-indented for visual indent
crossbar/router/dealer.py:1403:17: E128 continuation line under-indented for visual indent
crossbar/router/dealer.py:1404:17: E128 continuation line under-indented for visual indent
```

Jopie64:
Autoformat target didn't seem to be defined in the makefile. But I tried to do formatting myself following other logging examples. So hopefully flake is satisfied now...

Jopie64:
Thanks for reviewing/merging!! Any idea when a new docker version will be released?

oberstet:
Cool, thanks for contributing! Rgd docker images with the code: will come with release 19.6.1 (not yet here, but soonish ..)

oberstet:
@Jopie64 ok, I tagged and published a release v19.6.1 with this code, including docker images:

* https://hub.docker.com/r/crossbario/crossbar
* https://hub.docker.com/r/crossbario/crossbar-armhf
* https://hub.docker.com/r/crossbario/crossbar-aarch64

Jopie64:
@oberstet Thanks again, now for this quick release! We're going to update soon.

Issue #1609, opened 2019-07-08
Component on_ready event


oberstet:
Container and Router embedded components should fire an `on_ready` deferred, and `start_router_component`/`start_container_component` should only return _after_ that deferred has fired.

See also: https://github.com/crossbario/crossbarfx/issues/266

meejah:
Hmm, I see that currently `start_router_component` doesn't return a Deferred at all -- it could certainly return one that waits for the session's `on_join` to run.

Are you talking about a new `on_ready` API for client code to use?
e.g. the example router in the Autobahn repository has a custom authenticator which is just an `ApplicationSession` subclass -- so it doesn't signal any notion of "readiness" (beyond "`onJoin` has completed").

oberstet:
the goal is to have both crossbar router embedded and crossbar container components (which are currently based on ApplicationSession - and I wouldn't want to change/expand _that_ now) to:

Provide such components ^ a deferred in `extra`:

https://github.com/crossbario/crossbar/blob/ffcc896618d1bc8fdeba75aa2f990dcf9ad630dd/crossbar/worker/container.py#L211
https://github.com/crossbario/crossbar/blob/ffcc896618d1bc8fdeba75aa2f990dcf9ad630dd/crossbar/worker/router.py#L580

Then, that deferred is supposed to fire when the user code in `ApplicationSession.onJoin` finishes successfully.

> question is: who is firing ^ deferred? anything else than AB/CB code would break user code. changing AB/CB: no problem.

And `start_router_component` and `start_container_component` should only return when above deferred has fired.

Which means calls to these management procedures (`start_router_component`/`start_container_component`) will only return when the user session transport has connected, the session has joined and the `onJoin` has finished.

And that again would resolve https://github.com/crossbario/crossbarfx/issues/266 immediately.

It's clear that the new component API has this solved in a much cleaner way - and moving router/container components to the new component API would allow us to solve the specific issue here too - BUT: that is a quite invasive change .. which I'm afraid to do now.

What do you think?

meejah:
`Session` / `ApplicationSession` already fires a "ready" event when all `onJoin` **and** all registered `.on("join", ..)` handlers have run. I have an integration test for the "real" issue here, and yes simply having `start_router_component` wait properly for `onJoin` to complete fixes this case .. so I don't think we need a new API or a new concept of "ready" (I can't think of a use-case where "`onJoin` has completed" is not sufficient).

meejah:
[...]

PR #1615, opened 2019-07-17
wait for onjoin in container-components

meejah:


meejah:
see https://github.com/crossbario/crossbar/issues/1609

PR #1619, opened 2019-08-21
Click.option-s should have help text

ulziibay:
## What?
Add `help` text to click.option

## Why?
Click.option should ideally have a `help` text defined to be useful.


ulziibay:
Thanks for the thumbs up!

Issue #1607, opened 2019-07-05
New container shutdown and restart modes


oberstet:
# Shutdown modes

Currently, container workers allow only 2 shutdown modes:

* u'shutdown-manual': only shutdown container if instructed so (via management API)
* u'shutdown-on-last-component-stopped': automatically shutdown the container worker when the last component has stopped (container is hosting no more components)

We should add:

* u'shutdown-on-any-component-failed': shutdown the container worker if any hosted component fails
* u'shutdown-on-any-component-stopped': shutdown the container worker if any hosted component ends (either with success or with error)

# Restart modes

* u'restart-always': whenever the container worker ends, automatically restart it
* u'restart-on-failed': whenever the container workers ends with error, automatically restart it

oberstet:
example:

```json
{
    "type": "container",
    "options": {
        "pythonpath": [
            ".."
        ],
        "shutdown": "shutdown-on-any-component-failed",
        "restart": "restart-always"
    },
    "components": [
        {
            "type": "class",
            "classname": "myapp.MySession",
            "realm": "realm1",
            "transport": {
                "type": "websocket",
                "endpoint": {
                    "type": "tcp",
                    "host": "127.0.0.1",
                    "port": 8080
                },
                "url": "ws://127.0.0.1:8080/ws"
            }
        }
    ]
}
```

meejah:
While writing tests for the shutdown code, I see that there's a special 2-second timeout during `node` startup that basically says "if a component shuts down 'too quickly' / faster-than-2-seconds" the whole node shuts down.

Since we have an explicit shutdown mode for that now, perhaps it should be removed (and, arguably, the default should really be "shutdown-on-any-component-failed" to be mostly-compatible with the old code-path).

See https://github.com/crossbario/crossbar/blob/master/crossbar/node/node.py#L681

meejah:
There will be some interactions between the options (which could depend on which order we decide them, too). Arbitrarily choosing to decide "shutdown" first.

- If the mode is `shutdown-on-any-component-stopped` then we'll never try to re-start anything (because the first thing to stop will cause the container to shut down)
- We probably want a `restart-never` too (which is the default)
- Need to "recursively" restart, right? (That is, if a component fails and we're on anything but `restart-never` we attempt to re-start it .. now if that attempt fails we should keep trying right?)
[...]

PR #1622, opened 2019-08-29
Ticket1607 shutdown modes

meejah:
Adds the two new shutdown modes mentioned in #1607 .. along with a couple other minor fixups discovered while testing this.

(Note this doesn't yet include the restart modes mentioned).


meejah:
see https://github.com/crossbario/crossbar/pull/1623

PR #1626, opened 2019-09-07
Ticket1607 remove special-case node shutdown

meejah:
See https://github.com/crossbario/crossbar/issues/1607#issuecomment-526741763


oberstet:
cool, looks good -- only flake is complaining;) https://travis-ci.org/crossbario/crossbar/jobs/582140697

PR #1633, opened 2019-09-21
open-sourced proxy workers

oberstet:


oberstet:
we also need an example in crossbar-examples ..

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+    def set_backend(self, other_session):
+        self._backend = other_session
+
+    def onMessage(self, msg):
+        if isinstance(msg, self.do_not_forward):
+            return super(ProxyServerSession, self).onMessage(msg)
+        # TODO: should always destroy 'the other side' if one side
+        # (including the backend) goes away (e.g. message IDs would
+        # get out of sync if we just auto re-connected to the backend)
+        if not self._backend or not self._backend._transport:
```
meejah: I *think* this means we need to watch for Goodbye etc messages and take appropriate action..?

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+        self._front_to_back = dict()  # session_id -> ProxySession backend
+        self._backend_transport_config = transport_config
+
+    def has_role(self, role):
+        """
+        we just forward any role at all onto the 'real' backend
+
+        XXX TODO: is this true? or should we config which roles are
+        okay? or is that up to 'the real backend' itself?
+        """
```
meejah: This probably does want to be a configuration-thing (and/or a hook-function a user can pass in somewhere).

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+                authrealm=realm,
+                authid=authid,
+                authrole=authrole,
+            ))
+
+            def _join(*args):
+                self.log.info("ProxyRouter joined to backend session", )
+                backend_connected_d.callback(s)
+
+            s.on('join', _join)
```
meejah: We should handle other cases here; `on("disconnect")`, `on("leave")`, ... (this is connected to the other XXX/TODO above, I believe).

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+            return s
+
+        # XXX FIXME TODO look at transport-type etc
+        factory = WampWebSocketClientFactory(create_session, url="ws://localhost:9000/ws")
```
meejah: Yeah, this means we are just hard-coding "websocket" as the only proxy-able connection; need to handle others via config.

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+
+        def got_backend(backend):
+            self._front_to_back[details['session']] = backend
+
+            def backend_connected(_, details):
+                assert backend._session_id is not None
+
+                # always just forward to frontend
+                backend.onMessage = frontend._transport.send
+                frontend.set_backend(backend)
```
meejah: I'm not 100% sure this is correct behavior -- are there any message-kinds we **don't** want to forward (like the `do_not_forward` going the other way)...?

crossbar/worker/proxy.py
```diff
@@ -0,0 +1,537 @@
...
+        d.addCallback(got_backend)
+        return d
+
+    def _session_left(self, session, details):
+        self.log.info(
+            "ProxyRouter _session_left {session}",
+            session=session._session_id,
+        )
+        del self._front_to_back[session._session_id]
+
```
meejah: :+1: looks like we keep this dict accurate

crossbar/worker/proxy.py
oberstet: oh, I see: proxy is forwarding messages at the transport layer without actually creating a WAMP session to the backend using our machinery in eg ApplicationSession. 

so with the current design, yes, we would need to add all WAMP session establishment/close message handling to proxy. plus auth handling as this is part of the existing machinery.

that is, proxy needs to process these messages

* 531:class Hello(Message):
* 776:class Welcome(Message):
* 1027:class Abort(Message):
* 1113:class Challenge(Message):
* 1188:class Authenticate(Message):
* 1263:class Goodbye(Message):

and proxy can plain forward (both directions) all other messages - parsed/reserialized under the hood because the backend connection can use a different transport+serializer than the incoming client connection (IOW: we cannot simply forward at the websocket level)

see also my comment at the very end of this PR

crossbar/worker/proxy.py
oberstet: yes. with the current proxy design:

* the proxy handling is _above_ the raw transport - as it is a) authenticating the incoming WAMP and b) is translating between different WAMP serializers between frontend/backend transport
* the proxy handling is _below_ a full blown WAMP session - as it is a) not processing WAMP messages other than for session establishment (incl auth) nor session teardown and b) is not a 2-legged WAMP cross-link (which would require proxy to be a degenerate router, and establish a full blown WAMP session to the backend .. like rlinks do)

oberstet:
alright, now that I have dived deeper into this: pls lets discuss on the dev call today:

* we might want to keep proxy simple, and not authenticate WAMP in the proxy at all, but simply forward at the transport level - also because we will soon have rlinks
* I created an example config based on this assumption to be discussed here: https://github.com/crossbario/crossbar-examples/blob/master/proxy/.crossbar/config-node1.json

oberstet:
superseded by PR #407

PR #1635, opened 2019-09-21
Snap shared runtime

om26er:
* This enables third-party snaps to reuse the crossbar runtime environment, without the need to ship their own copy of things
* Helps projects like https://github.com/deskconn/deskconnd reduce their installed footprint.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,6 +1,5 @@
 name: crossbar
-version: latest
-version-script: git describe --always | sed -e 's/-/+git/;y/-/./' | tail -c +2
+adopt-info: crossbar
 summary: Crossbar.io - Polyglot application router.
 description: |
   Crossbar.io is a networking platform for distributed and microservice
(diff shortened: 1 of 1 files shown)


oberstet:
looks useful, but pls let me first understand: so this is about reusing the dependencies of crossbar inside the crossbar snap/environment from a different snap?

* does this expose any security risk for the crossbar (env)?
* we don't make any guarantees in crossbar about our own dependencies to external world - we can remove or add a dependency at any time
* crossbar has no python level official api whatsoever - importing a crossbar class is definitely discouraged/unsupported, and the AGPL3 will apply to the importing code (as a `from crossbar import xyz` is "linking" in the GPL sense)

om26er:
>     * does this expose any security risk for the crossbar (env)?

No, it does not, the change only provides read-only access to the content of the snap (which itself is read-only in nature)

>     * we don't make any guarantees in crossbar about our own dependencies to external world - we can remove or add a dependency at any time

Sure we only add/remove dependencies when crossbar needs it and what my change exposes is exactly that, so the promise can't be broken by us.

>     * crossbar has no python level official api whatsoever - importing a crossbar class is definitely discouraged/unsupported, and the AGPL3 will apply to the importing code (as a `from crossbar import xyz` is "linking" in the GPL sense)

You are right and I just changed my project to no longer import crossbar and replaced that will a shell script https://github.com/deskconn/deskconnd/blob/master/server.sh

> slightly OT, but I was curious;) https://deskconn.org seems down .. it loads forever ..

I didn't get around to setting up the website, I only have the domain...for now ;-)

PR #1643, opened 2019-10-07
docs for archive service

meejah:


oberstet:
cool, thanks=) merged and slightly expanded the example config - this is what I now use most of the time .. the feature is quite useful. for completeness I also filed a docs issue for the other webservice recently moved to oss https://github.com/crossbario/crossbar/issues/1644

PR #1646, opened 2019-10-23
Oss proxy workers new.2

meejah:
re-worked OSS proxy workers

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -612,6 +612,10 @@ def check_transport_auth(personality, auth, ignore=[], checks=None):
         'cookie': check_transport_auth_cookie,
         'cryptosign': check_transport_auth_cryptosign,
         'scram': check_transport_auth_scram,
+
+        # FIXME
+        'proxy-cryptosign': lambda _: None,
+        'proxy': lambda _: None,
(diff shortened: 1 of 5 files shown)


oberstet:
that's a great start! looking at the code, it would appear for example relatively easy to even add such behavior:

* in `Frontend.onMessage`, wait for the first `autobahn.wamp.message.Hello` received to read the `realm` the client wants to join.
* only then establish a backend connection, to a backend randomly selected from a list of backend transports for the respective `realm` configured in the node config

with above deferred backend connection, we could distribute to different backend routers based on realm, which could be quite useful I guess

what do you think?

meejah:
switching backends based on the realm we notice in `Hello`: yes, that should be straightforward I think! (...as long as that's the first message, and we don't have to buffer stuff .. those things are true though right?)

I'll also look into the mux-ing.

oberstet:
> as long as that's the first message, and we don't have to buffer stuff .. those things are true though right?)

yeah, exactly! which simplifies things, as we don't need to keep state .. other than "backend connected"

oberstet:
rgd mux'ing, I once started this approach: https://github.com/crossbario/autobahn-python/pull/989

PR #1650, opened 2019-10-31
proxy improvements

meejah:
This makes the backend config "not actually hard-coded" and also spies on the Hello message so a backend per realm can be configured. Currently, this works like this:

 - a backend config MAY have a `"realm"` key
 - a backend with no `"realm"` key is a "default backend"
 - when a client connects, we wait for the `Hello`:
    - if a backend exists with a matching realm, it is used
    - if none match, and there's a "default backend", it is used
    - if neither of the above, an error occurs

### DIFF
--- a/crossbar/worker/proxy.py
+++ b/crossbar/worker/proxy.py
@@ -126,7 +126,6 @@ class Frontend(WampWebSocketServerProtocol):
     _backend_transport = None  # valid if we're completely setup
     _transport_d = None  # a Deferred if we're waiting for the backend
 
-
     def onConnect(self, request):
         print("frontend: onConnect: {}".format(request))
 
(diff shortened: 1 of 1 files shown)


meejah:
(I haven't even tried WebSocket backends .. I think they'll fail because we need special-case code for "listen for when the connection is open" :/ but that can come next)

oberstet:
great! I think the decision logic for when a client connects is perfect. That makes a lot of sense.

Also it opens up (future) the option to add  "lazily-created multiplexed connections" - that this run multiple incoming connections over 1 backend (per-realm).

The latter could be done using a home grown muxing scheme over eg rawsocket (add muxing support to rawsocket).

We could also re-use the muxing that is built into QUIC (also see https://github.com/wamp-proto/wamp-proto/issues/363) and run 1 QUIC stream (within a QUIC connection) per incoming WAMP session. Those sessions can be to one or multiple different realms.

IOW / IMO: ideally proxy workers will be able to MUX sessions flexibly both _from_ incoming connections and _to_ backend connections - at a per-message basis, bidirectional and stateless (other than initial HELLO processing)

PR #1667, opened 2019-12-07
please update https://pypi.org/simple/autobahn/

vaibhav-rbs:
please update https://pypi.org/simple/autobahn/ with latest, autobahn-19.11.2.tar.gz or don't update requirements-min.txt autobahn==19.11.2


oberstet:
no, this is correct: crossbar GH works with autobahn GH

PR #1668, opened 2019-12-10
move where backend transports are defined

meejah:
As discussed here: https://github.com/crossbario/crossbar-examples/pull/128#issuecomment-548265605

This moves the proxy "backend" configurations underneath the paths in a proxy. 

Anyway. If this looks like a better way to organize the configuration, I can fix up the example. I didn't move where the "start the backend" code is managed (in the `proxy` worker, `ProxyController`). It probably makes sense to put that (and e.g. `_find_backend_for`) in the transport / service code (`ProxyWebSocketService`) since that's where the backends are configured now. Then (when multiplexing works) there will be one "actual backend connection" for every configured "backend" transport in the config (which might make the whole thing easier to think about for people doing the configuration).

### DIFF
--- a/crossbar/node/node.py
+++ b/crossbar/node/node.py
@@ -750,6 +750,8 @@ class Node(object):
     def _configure_native_worker_proxy(self, worker_logname, worker_id, worker):
         yield self._configure_native_worker_common(worker_logname, worker_id, worker)
 
+        backends = []
+
         # start transports on proxy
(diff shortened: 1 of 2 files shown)


meejah:
Hmm, even if the backend-handling is moved to `ProxyWebSocketService` there can still be duplicate / confusion with the backends. It really needs to be per-path I guess...

PR #1671, opened 2019-12-16
OSS proxy workers refactor

meejah:
Further refinement to https://github.com/crossbario/crossbar/pull/1668 (which moves where in the config backends are configured). This also cleans up the code a bunch and handles closed-connections properly

### DIFF
--- a/crossbar/node/node.py
+++ b/crossbar/node/node.py
@@ -750,6 +750,8 @@ class Node(object):
     def _configure_native_worker_proxy(self, worker_logname, worker_id, worker):
         yield self._configure_native_worker_common(worker_logname, worker_id, worker)
 
+        backends = []
+
         # start transports on proxy
         for i, transport in enumerate(worker.get('transports', [])):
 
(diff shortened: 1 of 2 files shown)


oberstet:
looks good! progress;) now, rgd the CI fails: one is flake8 complaining .. the other seems unrelated, and I'm not sure why it pops up:

https://travis-ci.org/crossbario/crossbar/jobs/625539231#L682

wsaccel (a native cpyext) package fails for pypy .. but wsaccel should _not_ be used on pypy anyway:

https://github.com/crossbario/autobahn-python/blob/master/setup.py#L78

meejah:
I think the pypy failures are due to `lz4` dependency via `autobahn[compress]` .. should be `lz4-cffi` for PyPy support. (But also, do we need `lz4` at all? I don't see it mentioned in autobahn's code anywhere..)

meejah:
(If removing `lz4` is the right thing, see https://github.com/crossbario/autobahn-python/pull/1280 )

Issue #1684, opened 2020-01-21
Cleanup legacy string literals


om26er:
One candidate for change is/are `u"my_string"` (single and double quoted variants)

Since crossbar has been Py3+ only for over a year, I see there is very little legacy code. Are there any other instances that anyone could think of ?

oberstet:
yes, but I wouldn't touch that other code (now) ;)

real answer:

eg one could propose changing `@inlineCallbacks` to `async def`. for this specific example, I wouldn't support that until/unless we have a) added our performance tests to the CI and b) demonstrate no adverse effect on the PRs' CI run then. 

IOW: it really depends for the cases other then the one you listed (string literals) .. fixing that is harmless and of course cool! it further cleans up the code a bit - which is always a nice thing=)

PR #1693, opened 2020-02-02
cleanup: remove unicode strings

om26er:
* Replaces all instances of u'str' with 'str'
* Replaces all instances of u"str" with "str"

closes https://github.com/crossbario/crossbar/issues/1684


om26er:
This PR also fixes the sphinx CI matrix which was failing due to a version conflict. Other failures are unrelated.

oberstet:
I haven't reviewed this line-by-line, as the PR touches large number of lines .. I assume, the first commit in this PR (removing `u`) only contains changes done automatically using some shell/sed magic?

apart from that, the last commit https://github.com/crossbario/crossbar/pull/1693/commits/f62e0cd88ae2047a0023afed1a745d002ac4742d

needs to be backed out again as:

* it must be updated using `make freeze` https://github.com/crossbario/crossbar/blob/master/Makefile#L64
* it cannot use platform specifiers, as pip doesn't support combining those with hashes

om26er:
This should be good now

PR #1694, opened 2020-02-02
first-cut dealer timeout/cancel implementation

meejah:
First hack at a timeout + cancel from the Dealer. I'm sure there are other cases to consider (like if one side or the other doesn't support cancel).

### DIFF
--- a/crossbar/router/dealer.py
+++ b/crossbar/router/dealer.py
@@ -67,6 +67,7 @@ class InvocationRequest(object):
         'forward_for',
         'canceled',
         'error_msg',
+        'timeout_call',
     )
 
     def __init__(self, id, registration, caller, call, callee, forward_for):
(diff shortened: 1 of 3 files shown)


oberstet:
@meejah I've prepared everything for a v20.2.1 release of CB: do you want to land this PR already or wait?

meejah:
lets wait, and do it in the "next" release (in case of fallout etc?)

oberstet:
> lets wait, and do it in the "next" release (in case of fallout etc?)

ok! once we have it on master, we can give it some more testing, probably also on the demo instances before pushing a new release

meejah:
Yeah .. "should" be fairly safe as it's kind of "opt-in" by providing the new `timeout=` kwarg, but ...

oberstet:
CI still complains: flake8 (and sphinx, I restarted that one to see if it disappears ..)

meejah:
Hrmmm, hard to make flake8 happy: doesn't like newline before or after the `and` and also doesn't like a blackslash in there ... so I guess it has to be one line :/

Oh, the docs say there's no way to make it happy here, so I'm going with the "soon to be best practice" suggestion (and ignoring W503 in the flake8): https://www.flake8rules.com/rules/W503.html

PR #1711, opened 2020-02-17
Fallback path support packages

om26er:
Improvement to https://github.com/crossbario/crossbar/pull/1408, now supporting reading fallback from python packages as well.

closes https://github.com/crossbario/crossbar/issues/1712


crossbar/webservice/static.py
```diff
@@ -155,8 +155,8 @@ def create(transport, path, config):
         #
         fallback = static_options.get('default_file')
         if fallback:
-            resource.childNotFound = ResourceFallback(path, config)
+            resource.childNotFound = ResourceFallback(os.path.join(static_dir, fallback))
```
oberstet: we can remove `ResourceFallback` altogether and just use `File` here directly

crossbar/webservice/static.py
```diff
@@ -131,7 +131,7 @@ def create(transport, path, config):
 
             raise ApplicationError("crossbar.error.invalid_configuration", "missing web spec")
 
-        static_dir = static_dir.encode('ascii', 'ignore')  # http://stackoverflow.com/a/20433918/884770
+        static_dir_enc = static_dir.encode('ascii', 'ignore')  # http://stackoverflow.com/a/20433918/884770
```
oberstet: why do we need to change the variable name?

crossbar/webservice/base.py
```diff
@@ -62,11 +62,8 @@ class ResourceFallback(File):
     """
     Handle requests for non-existent URL's
     """
-    def __init__(self, path, config, **kwargs):
+    def __init__(self, path, **kwargs):
```
oberstet: we can completely remove `ResourceFallback` (see also the other comment below)

oberstet:
also: ideally, we had an issue file for this, a changelog entry refering the issue/PR and adjustment to docs. (I skip a CI test .. for this PR - but I think the changelog/docs would be quick to add ..)

crossbar/webservice/base.py
om26er: done

crossbar/webservice/static.py
om26er: done

crossbar/webservice/static.py
```diff
@@ -155,8 +155,8 @@ def create(transport, path, config):
         #
         fallback = static_options.get('default_file')
         if fallback:
-            resource.childNotFound = ResourceFallback(path, config)
+            resource.childNotFound = File(os.path.join(static_dir, fallback))
         else:
-            resource.childNotFound = Resource404(transport.templates, static_dir)
+            resource.childNotFound = Resource404(transport.templates, static_dir_enc)
```
oberstet: needs to be "static_dir"

crossbar/webservice/static.py
```diff
@@ -143,7 +143,7 @@ def create(transport, path, config):
         cache_timeout = static_options.get('cache_timeout', DEFAULT_CACHE_TIMEOUT)
         allow_cross_origin = static_options.get('allow_cross_origin', True)
 
-        resource = static_resource_class(static_dir, cache_timeout=cache_timeout, allow_cross_origin=allow_cross_origin)
+        resource = static_resource_class(static_dir_enc, cache_timeout=cache_timeout, allow_cross_origin=allow_cross_origin)
```
oberstet: needs to be "static_dir"

crossbar/webservice/static.py
oberstet: needs to be "static_dir"

om26er: We use the existing `static_dir` variable in the newly added code below, It was being encoded and overwritten previously. As `os.path.join()` expects both paths to be string.

The alternate could be to just avoid the renaming and call `static_dir.decode('ascii')` where needed. Your call!

om26er:
> also: ideally, we had an issue file for this, a changelog entry refering the issue/PR and adjustment to docs. (I skip a CI test .. for this PR - but I think the changelog/docs would be quick to add ..)

Filed https://github.com/crossbario/crossbar/issues/1712, added a change log entry and also created another issue to track a CI test for this feature. Only need to update the docs, though I wonder what would be the "right" place for that...

oberstet:
perfect, ready to merge!

PR #1718, opened 2020-02-26
close websocket connection by calling the right method

om26er:


oberstet:
fixed via https://github.com/crossbario/crossbar/pull/1723

PR #1721, opened 2020-02-27
Fix error in wamp.session.list and wamp.session.count …

DZabavchik:
When any sessions that are not derived from `RouterSession` or `PendingAuth` are present (native workers, authorizers, authenticators etc.) calls to `wamp.session.list` or `wamp.session.count` fail with ```object has no attribute '_session_details'```

Whe list of authroles is provided (filter_authroles) attempt to get `_session_details` may fail
```if filter_authroles is None or session._session_details.authrole in filter_authroles```

The PR uses `hasattr(session, '_session_details')` which is already used in other places of the same file.

Not part of the fix, but something to consider adding is `wamp.session.summary`.  Which would return counts of sessions grouped by authrole.  Alternative to it is calling `wamp.session.count` for each role separately, which makes it O(M * N) as it has to enumerate all N sessions for M roles.

### DIFF
--- a/crossbar/router/service.py
+++ b/crossbar/router/service.py
@@ -192,6 +192,30 @@ class RouterServiceAgent(ApplicationSession):
         if not isinstance(failure.value, ApplicationError):
             super(RouterServiceAgent, self).onUserError(failure, msg)
 
+    @wamp.register('wamp.session.summary')
+    def session_summary(self, filter_authroles=None, details=None):
+        """
+        Get summary of sessions currently joined on the router grouped by authroles.
+
+        :param filter_authroles: If provided, only return sessions counts for authroles from this list.
+        :type filter_authroles: None or list
+
+        :returns: Summary of WAMP session counts (associative array role -> count, order undefined).
+        :rtype: list
+        """
+        assert (filter_authroles is None or type(filter_authroles) == list)
+        session_counts = {}
+        for session in self._router._session_id_to_session.values():
+            if not is_restricted_session(session):
+                if filter_authroles is None or (hasattr(session, '_session_details') and session._session_details.authrole in filter_authroles):
+                    auth_role = session._session_details.authrole
+                    if auth_role in session_counts:
+                        session_counts[auth_role] += 1
+                    else:
+                        session_counts[auth_role] = 1
+
+        return session_counts
+
     @wamp.register('wamp.session.list')
     def session_list(self, filter_authroles=None, details=None):
         """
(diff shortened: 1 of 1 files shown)


DZabavchik:
[SessionListWithAuthTest.zip](https://github.com/crossbario/crossbar/files/4264520/SessionListWithAuthTest.zip)
Test that was used to reproduce the issue and verify fix.

oberstet:
the fix: +1.

the new "session_summary": I can definitely see why this might be useful, and it also looks perfect from impl as well - the problem is, that it exposes new API in the URI space `"wamp.*"` - which is restricted to be on the public WAMP spec. crossbar.io strictly adheres to that.

so either we add such thing first to the spec (at least an issue, and a PR with text proposal, and "rough consensus" among implementors), or the proc is exposed on a different URI, this is a "problem" ..

DZabavchik:
I assumed that WAMP spec was prescriptive, listing required endpoints and specific implementations can extend interface with private endpoints (vs. restrictive where it limits interface).

I can remove "session_summary" altogether.  I just thought that as long as data contract on existing endpoints is not violated, newer endpoints can be added with subsequent discussion and proposals for spec changes.

Let me know. Tnx

oberstet:
> I can remove "session_summary" altogether

yeah, that would be great! possibly save the respective "session_summary" in another PR, or even just an issue, so that we can come back to that as well!

because, yes, in general WAMP reserves every URI that starts with "wamp.". let me try to collect some relevant infos/pointers.

for its own definitions and use, WAMP reserves

https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#reserved-uris

WAMP then introduces a set of such URIs and their meaning - as _officially standardized_ behavior:

https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#predefined-uris
https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#registration-meta-api
https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#subscription-meta-api
https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#session-meta-api
https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#event-history
https://wamp-proto.org/_static/gen/wamp_latest_ietf.html#testament-meta-procedures

WAMP _implementations_ such as Crossbar.io or other routers can define their own additional URIs of course. And they do .. eg Crossbar.io has a whole bunch of them on prefixes "crossbar.".

We can also of course add to the official WAMP spec! To do that, the proposal would best be filed as an issue on the WAMP repo first. Other implementors then have a chance to chime in. Ideally, after a PR to the spec text and some iteration, it is merged to the WAMP spec. An implementation in Crossbar.io could be done in parallel, as the development of the WAMP spec always has been done as co-iterating spec text and experimental implementations ..

DZabavchik:
Done. PR only contains bug fix.

I was unable to find an example of any `crossbar.` prefixes or a good place for `crossbar.session.summary` other than `crossbar/router/service.py`.  It appears to be the best module for this because it has access to `_router._session_id_to_session.values()`.

Issue #1720, opened 2020-02-26
'FrontendProxyProtocol' object has no attribute 'close'


om26er:
I have seen that appear in logs, the relevant code is https://github.com/crossbario/crossbar/blob/fbdc8c758532acdc2d93b4cd4a7203748e1548f7/crossbar/worker/proxy.py#L298 -- Shall we call `sendClose()` instead ?

(sorry for not having complete log)

om26er:
attempted https://github.com/crossbario/crossbar/pull/1718#issuecomment-591620519

oberstet:
> Shall we call sendClose() instead ?

no, as sendClose is specific to WebSocket .. eg RawSocket doesn't have it. Proxy workers should in principle work with both WebSocket and RawSocket as frontend protocol (even though we likely only need/want WebSocket on the demo instances eg) .. "close should work" ..

meejah:
Just so I remember, this bug is for when the frontend-to-backend is using a WebSocket..

PR #1723, opened 2020-03-02
ticket 1720: close not propagated properly from backend

meejah:
Neither websocket nor rawsocket backends were propagating close() properly (if the backend closed) -- but for different reasons. See #1720


oberstet:
mmh, not sure where that `ImportError: cannot import name 'ContextManager'` https://travis-ci.org/crossbario/crossbar/jobs/657099258 is coming from - it doesn't seem to be related to this PR (I think I've seen it on other of our repos very recently), and it happens for all PyPy tests .. maybe some dependency was updated?

PR #1733, opened 2020-03-26
proxy-workers re-factoring

meejah:
Still needs more authentication details on the backend.

### DIFF
--- a/crossbar/router/auth/pending.py
+++ b/crossbar/router/auth/pending.py
@@ -28,10 +28,37 @@
 #
 #####################################################################################
 
+import abc
 from autobahn.wamp import types
 from autobahn.wamp.exception import ApplicationError
 
 __all__ = ('PendingAuth',)
+
+
+class IRealmContainer(abc.ABC):
+    """
+    An object the authentication system can query about the existince
+    of realms and roles.
+    """
+
+    @abc.abstractmethod
+    def has_realm(self, realm):
+        """
+        :returns: True if the given realm exists
+        """
+
+    @abc.abstractmethod
+    def has_role(self, realm, role):
+        """
+        :returns: True if the given role exists inside the realm
+        """
+
+    @abc.abstractmethod
+    def get_service_session(self):
+        """
+        :returns: ApplicationSession suitable for use by dynamic
+            authenticators
+        """
 
 
 class PendingAuth:
@@ -45,27 +72,28 @@ class PendingAuth:
 
     AUTHMETHOD = 'abstract'
 
-    def __init__(self, session, config):
+    def __init__(self, pending_session_id, transport_info, realm_container, config):
         """
+        :param int pending_session_id: the Session ID if this succeeds
 
-        :param session: The authenticating session.
-        :type session: obj
+        :param dict transport_info: information about the session's rtansport
+
+        :param IRealmContainer realm_container: access configured realms / roles
+
         :param config: Authentication configuration to apply for the pending auth.
         :type config: dict
         """
-        # The session that is authenticating
-        self._session = session
 
         # Details about the authenticating session
         self._session_details = {
-            'transport': session._transport._transport_info,
-            'session': session._pending_session_id,
+            'transport': transport_info,
+            'session': pending_session_id,
             'authmethod': None,
             'authextra': None
         }
 
         # The router factory we are working for
-        self._router_factory = session._router_factory
+        self._realm_container = realm_container
 
         # WAMP-Ticket configuration to apply for the pending auth
         self._config = config
@@ -142,12 +170,18 @@ class PendingAuth:
             return types.Deny(ApplicationError.NO_SUCH_ROLE, message='no authrole assigned')
 
         # if realm is not started on router, bail out now!
-        if self._realm not in self._router_factory:
-            return types.Deny(ApplicationError.NO_SUCH_REALM, message='no realm "{}" exists on this router'.format(self._realm))
+        if not self._realm_container.has_realm(self._realm):
+            return types.Deny(
+                ApplicationError.NO_SUCH_REALM,
+                message='no realm "{}" exists on this router'.format(self._realm)
+            )
 
         # if role is not running on realm, bail out now!
-        if self._authrole and not self._router_factory[self._realm].has_role(self._authrole):
-            return types.Deny(ApplicationError.NO_SUCH_ROLE, message='realm "{}" has no role "{}"'.format(self._realm, self._authrole))
+        if self._authrole and not self._realm_container.has_role(self._realm, self._authrole):
+            return types.Deny(
+                ApplicationError.NO_SUCH_ROLE,
+                message='realm "{}" has no role "{}"'.format(self._realm, self._authrole)
+            )
 
     def _init_dynamic_authenticator(self):
         self._authenticator = self._config['authenticator']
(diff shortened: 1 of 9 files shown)


crossbar/worker/proxy.py
```diff
@@ -60,6 +57,284 @@ class ProxyWorkerProcess(worker.NativeWorkerProcess):
...
+
+        # if we have a backend connection, it'll be here (and be a
+        # Session instance)
+        self._backend_session = None
+
+        # in _transport_info are (possibly) the following keys (gleaned from router/session.py):
+        # client_cert
+        # channel_id
+        # type (websocket, etc)
+        # protocol (websocket subprotocol?) None for rawsocket
```
oberstet: yes, this is the websocket subprotocol (if the transport is over websocket): a string like `"wamp.2.json"` and similar

crossbar/router/auth/pending.py
```diff
@@ -28,12 +28,39 @@
...
 #####################################################################################
 
+import abc
 from autobahn.wamp import types
 from autobahn.wamp.exception import ApplicationError
 
 __all__ = ('PendingAuth',)
 
 
+class IRealmContainer(abc.ABC):
```
oberstet: this is great! it clearly defines the interface the proxy needs in the code for the incoming connection side to check for existence of realm-role requested by the client in the backend - which can be the router or poxy controller. cool, make sense

crossbar/router/auth/pending.py
```diff
@@ -45,27 +72,28 @@ class PendingAuth:
 
     AUTHMETHOD = 'abstract'
 
-    def __init__(self, session, config):
+    def __init__(self, pending_session_id, transport_info, realm_container, config):
```
oberstet: yeah, perfect! so this removes the (deep) session object reference, which was the source of the strong binding to routing internals the authenticator actually does not need

crossbar/router/auth/pending.py
```diff
@@ -45,27 +72,28 @@ class PendingAuth:
...
         :type config: dict
         """
-        # The session that is authenticating
-        self._session = session
 
         # Details about the authenticating session
         self._session_details = {
-            'transport': session._transport._transport_info,
-            'session': session._pending_session_id,
+            'transport': transport_info,
```
oberstet: I think this makes a lot of sense: the job of collecting infos for a transport should not be done in authenticators, but the stuff that creates and calls authenticators - IOW, exactly as this change does: provide it from outside. fwiw, this seamingly harmless "transport info" thing has been haunting us in different places;) this cleans up one of those

crossbar/worker/proxy.py
```diff
@@ -60,6 +57,284 @@ class ProxyWorkerProcess(worker.NativeWorkerProcess):
...
+        is considered running and is_open() would return true, as soon as this callback
+        has completed successfully.
+
+        :param transport: The WAMP transport.
+        :type transport: object implementing :class:`autobahn.wamp.interfaces.ITransport`
+        """
+        # print("ProxySession.onOpen: {}".format(transport))
+        self.transport = transport
+
+    from functools import singledispatchmethod
```
oberstet: oh, this is new to me: what's that? what does it do on onMessage?

crossbar/worker/proxy.py
```diff
@@ -89,352 +363,140 @@ def onConnect(self):
...
         # the frontend is speaking
-        self._frontend.forward_message(msg)
-
-        # ... should we do this explicitly, or just count on the
-        # "disconnect" propogating properly
-        if False and isinstance(msg, Goodbye) and self._frontend:
-            self._frontend.transport.loseConnection()
-            self._frontend = None
+        # print("BackendProxySession.onMessage: {}".format(msg))
+        if isinstance(msg, (message.Welcome, message.Challenge, message.Abort, message.Goodbye)):
```
oberstet: exactly! this is the crucial code that defines the behavior of the new WAMP proxy:

* it handles frontend and backend wamp opening handshake / authentication, but _not_
* session-level wamp message types, which it instead forwards directly (without any processing at the app level, but with re-serialization if required)

crossbar/worker/proxy.py
```diff
@@ -60,6 +57,284 @@ class ProxyWorkerProcess(worker.NativeWorkerProcess):
...
+            except Error:
+                pass
+            self._backend_session = None
+            self.transport = None
+
+
+ITransportHandler.register(ProxySession)
+
+
 class BackendProxySession(Session):
```
oberstet: This looks already great! as far as I see, the only piece missing is `onChallenge` and adding WAMP-cryptosign authentication, using the node private key and forwarding the incoming (already authenticated) principal `authid`, `authrole` and `realm` in the `authextra` used in the backend connection. All the other pieces are already there. Awesome=)

oberstet: I just realize: for now, it's even simpler than that: because if the backend router worker listens on UDS, the filesystem is authenticating the incoming proxy connections, and all the backend authenticator needs to do is pick out authid, authrole and realm from authextra and return that as principal.

crossbar/worker/proxy.py
```diff
@@ -89,352 +363,140 @@ def onConnect(self):
...
-        websocket_factory = WebSocketServerFactory()
-        websocket_factory.protocol = FrontendProxyProtocol
+    return connected_d
 
-        resource = WebSocketResource(websocket_factory)
 
-        service = ProxyWebSocketService(transport, path, config, resource)
-        websocket_factory._service = service
+# implements IRealmContainer
+class ProxyController(RouterController):
```
oberstet: fantastic! can_map_backend and map_backend and auxiliary functions, this makes a lot of sense and nicely factors the code with little binding in between. I like that a lot .. it just seams natural from now looking over the implementation.

crossbar/worker/proxy.py
meejah: See the `_hello_received` which is decorated with `@onMessage.register(message.Hello)` .. basically now `onMessage` will (possibly) call different functions depending on the **type** of the first argument. So if a `Hello` arrives, `_hello_received` is called; for anything without a "special" message, `onMessage` itself runs.

Basically a replacement for the `if isinstance(first_arg, SomeClass):  .. elif .. elif` pattern.

meejah: (When wrote that I expected to have more special-cases than just Hello ;) so could now simplify it to not use those decorators..?)

crossbar/worker/proxy.py
meejah: Yeah. And I don't think it's worth supporting anything besides "nothing, it's UDS" or "cryptosign-proxy" for the actual backend connections, right? (I mean, `wampcra` and `wamp-scram` or tickets)

oberstet: >  And I don't think it's worth supporting anything besides "nothing, it's UDS" or "cryptosign-proxy" for the actual backend connections, right?

yes, agreed, only 2 methods: nothing and cryptosign

crossbar/worker/proxy.py
oberstet: interesting! thanks for explaining. however, yes, in this particular case at least, yeah. I'd prefer a basic/dumb if/else style. in the situation here, using the magic decorator IMO doesn't bring benefits, but obfuscates things.

hiding magic behind decorators is cool - sometimes. but for singledispatchmethod, I am not really convinced. because it distributes decision logic and hides it behind a magic decorator. 

taken to extreme, eg consider

```python
@arithmetic('add', a, b=3):
def my_add(self, a):
    pass

@arithmetic('mul', a, b=3):
def my_mul(self, a):
    pass
`
@arithmetic():
def my_any(self, op, args*):
    pass
`` 

awesome, beautiful, isn't it? ;)

nope. this now hides dynamic (run-time) message dispatching (is it add or mul or what) and computation behind a decorator - which is defined somewhere else

Comment (meejah) [reply crossbar/worker/proxy.py]: okay, took it out

**Comment by oberstet** [comment]
travis notification did not work .. the CI run has actually finished https://travis-ci.org/github/crossbario/crossbar/builds/667361986

**Comment by meejah** [comment]
@oberstet see last commit for the proxy versions of anonymous + cryptosign auth (could maybe write it as a generic `PendingAuthProxy` which has an instance of Anonymous or Cryptosign to call-through to).

File: crossbar/worker/proxy.py [hunk 9]
```diff
@@ -60,366 +60,438 @@ class ProxyWorkerProcess(worker.NativeWorkerProcess):
...
+                authmethods = [
+                    '{}-proxy'.format(x)
+                    for x in self._backend_session._authenticators.keys()
+                ]
+
+                self._backend_session.join(
+                    self._realm,
+                    authmethods=authmethods,
+                    authid=None,
+                    authrole=self._authrole,  # same role as front-end?
```
oberstet: I would argue `authrole==None` would make sense - but that it doesn't matter because it is not used in the backend?

in general, my view:

the backend should authenticate the incoming proxy connection either "anonymous" (with a fixed credential defined only by the backend) or "cryptosign" (_only_ relying on the pubkey presented by the proxy, _not_ anything requested by the proxy in authid/authrole/realm).

_if_ that authentication succeeds, then the backend will (without wamp MUX) authenticate the incoming connection using the credentials supplied in "cb_proxy_*"

in this case the realm/authid/authrole requested in HELLO in the proxy=>backend connection is never actually used - and hence should be set (or omitted) by the proxy in the first place

PR #1741, opened 2020-04-03
Proxy polish

oberstet:


oberstet:
ok, I actually started reviewing the proxy code in detail, saw some smaller things and started to move around and "polish" a bit here and there. at some point, I decided a PR would be quicker than lots of comments;) this PR should also fill some minor gaps. pls have a look whether this actually is an improvement in your eyes (eg I tried to make code paths more explicit/easier to follow) ..

crossbar/worker/proxy.py
```diff
@@ -60,50 +73,67 @@ class ProxyWorkerProcess(worker.NativeWorkerProcess):
...
-        # protocol (websocket subprotocol?) None for rawsocket
-        # peer
-        # http_headers_received
-        # http_headers_sent
-        # http_response_lines
-        # websocket_extensions_in_use
-        # cbtid
-
-        # for rawsocket, it's type, protocol=None and peer
+        self._aauthextra = None
```
meejah, line 137: Typo? (`self._authextra` instead?)

crossbar/worker/proxy.py
```diff
@@ -126,248 +203,280 @@ def onMessage(self, msg):
...
-                realm=self._realm,
-                authid=self._authid,
-                authrole=self._authrole,
-                authmethod=self._authmethod,
-                authprovider=self._authprovider,
-                authextra=self._authextra,
-                #  custom=custom,
-            )
+            if isinstance(msg, message.Hello) or isinstance(msg, message.Abort) or isinstance(msg, message.Authenticate):
+                raise ProtocolError("Received {} message while proxy frontend session is already joined".format(msg.__class__.__name__))
```
meejah, line 230: Is Abort only during the handshake phase?

crossbar/worker/proxy.py
```diff
@@ -126,248 +203,280 @@ def onMessage(self, msg):
...
-        elif isinstance(res, types.Challenge):
-            msg = message.Challenge(
-                res.method,
-                res.extra,
-            )
+            # https://wamp-proto.org/_static/gen/wamp_latest.html#session-closing
+            elif isinstance(msg, message.Goodbye):
+                if self._backend_session:
+                    yield self._controller.unmap_backend(self, self._backend_session)
+                    self._backend_session = None
```
meejah, line 235: Ah, yes +1 for `unmap_backend()` :)

crossbar/worker/proxy.py
```diff
@@ -126,248 +203,280 @@ def onMessage(self, msg):
...
+                        "proxy_realm": accept.realm,
+                        "proxy_authid": accept.authid,
+                        "proxy_authrole": accept.authrole,
+                        "proxy_authmethod": accept.authmethod,
+                        "proxy_authprovider": accept.authprovider,
+
+                        # this is the authextra returned from the frontend authenticator, which
+                        # would normally be returned to the client
+                        "proxy_authextra": accept.authextra,
+                    }
```
meejah, line 303: ah, nice .. I guess this can also ultimately allow someone doing `dynamic` auth on the actual-backend to get this information direct from the "real" client.

crossbar/worker/proxy.py
oberstet: yeah, typo: fixed

crossbar/worker/proxy.py
oberstet: yes: once a session is established, GOODBYE is the only message allowed to initiate the WAMP closing handshake (well, and the handshake is merely a GOODBYE reply from the peer) - also https://wamp-proto.org/_static/gen/wamp_latest.html#difference-between-abort-and-goodbye

crossbar/worker/proxy.py
oberstet: this should also make it simpler to later add backend session caching/muxing

crossbar/worker/proxy.py
oberstet: yeah, I've further expanded the info. this is the `authextra` an authenticating (frondend) client receives now:

```
{
   'x_cb_node': 'node1',
   'x_cb_worker': 'worker001',
   'x_cb_peer': 'unix:None',
   'x_cb_pid': 19998,
   'x_cb_proxy_node': 'node1',
   'x_cb_proxy_worker': 'worker003',
   'x_cb_proxy_peer': 'tcp4:127.0.0.1:36902',
   'x_cb_proxy_pid': 20005
}
```

so both the frontend proxy and backend proxy contributed to the `authextra` ultimately returned to the client.

oberstet: Screenshot from 2020-04-05 18-31-14

Issue #1511, opened 2019-02-25
HttpOnly for Cookie Authentication


manuelmager:
# Goal
Mitigate XSS attacks by preventing client javascript from accessing the Authentication Cookie content.

# Why is this important for you
We want to use the cookie to remember that a user is already authorized so as to allow the user to not provide credentials with every connection. Storing this information in a non-HttpOnly cookie is a security risk when it comes to cross site scripting.

# Proposed Solutions
## Recommended: always HttpOnly
Add `;HttpOnly` when creating a cookie in the `cookiestore.py`. I know of no use case that requires having the cookie accessible for client-side javascript, so from my point of view it seems to be a good idea to always set Cookies to be HttpOnly.

Replace
https://github.com/crossbario/crossbar/blob/1e18ceef1b444f60bca118cd5e57eb7ffbd8f2bf/crossbar/router/cookiestore.py#L137
with
`return cbtid, '%s=%s;max-age=%d;HttpOnly' % (self._cookie_id_field, cbtid, cbtData['max_age'])`

## Alternative: configurable HttpOnly
Add an optional configuration parameter `http_only` to allow defining whether the cookie should be inaccessible to client-side javascript. In case there is a use case for not using HttpOnly.

oberstet:
this needs to be added here: https://github.com/crossbario/crossbar/blob/master/crossbar/router/protocol.py#L254

it would be trivial to add 2 lines to set `HttpOnly`, `SameSite` and such - however, we might want to have knobs to control setting of these headers in the node config - so it will be "a couple of lines" instead of just 2.

dshymanskyi-vivasg:
Hi i created [pull request](https://github.com/crossbario/crossbar/pull/1753), please check it out as soon as you have time for it.

yankos:
Does this issue resolved?

oberstet:
yeah, it's solved!

PR #1753, opened 2020-04-15
Configurable cookie headers  #issue-1511

dshymanskyi-vivasg:
Added functionality to set cookie headers in node config. 
Related to this [issue](https://github.com/crossbario/crossbar/issues/1511)

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -663,7 +663,8 @@ def check_transport_cookie(personality, cookie, ignore=[]):
         'length': (False, [int]),
         'max_age': (False, [int]),
         'store': (False, [Mapping]),
-        'headers': (False, [Mapping])
+        'http_strict': (False, [bool]),
+        'same_site': (False, [str])
     }, cookie, "WebSocket cookie configuration")
 
     if 'name' in cookie:
@@ -699,6 +700,16 @@ def check_transport_cookie(personality, cookie, ignore=[]):
             pass
         else:
             raise InvalidConfigException('logic error')
+
+    if 'http_strict' in cookie:
+        http_strict = cookie['http_strict']
+        if not isinstance(http_strict, bool):
+            raise InvalidConfigException("invalid type {} for attribute 'http_strict' in cookie - must be bool".format(type(http_strict)))
+
+    if 'same_site' in cookie:
+        same_site = cookie['same_site']
+        if same_site not in ["Strict", "Lax", "None"]:
+            raise InvalidConfigException("invalid attribute value '{}' for attribute 'same_site' - must be one of 'Strict', 'Lax', 'None'".format(same_site))
 
 
 def check_endpoint_backlog(backlog):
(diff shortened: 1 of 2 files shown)


oberstet:
thanks for the PR! looked into it: this adds the general ability to set HTTP cookie headers from the config file. the issue you linked is about a new boolean option `http_only` in the config, that when enabled will add `;HttpOnly` to the end value returned by `self.factory._cookiestore.create()` .. which is different;)

dshymanskyi-vivasg:
Sorry for late response. Maybe I misunderstood you, but when i read your comment to issue:
> it would be trivial to add 2 lines to set HttpOnly, SameSite and such - however, we might want to have knobs to control setting of these headers in the node config - so it will be "a couple of lines" instead of just 2.
I thought about what i did in this pull request 

For example if you set in config.json for websocket transport
`

      "type": "websocket",
      "cookie": {
           "store": {
                "type": "file",
                "filename": "cookies.dat"
            },
            "headers": {
                "HttpOnly": true,
                "SameSite": "Strict"
           }
`

to end of cookies will be added ";HttpOnly;SameSite=Strict;"

oberstet:
> For example if you set in config.json for websocket transport

yes, I understand that. it would also allow `"Foobar": "23"`. we don't want that flexibility. we want exactly 1 additional **bit** in the config: http_strict yes/no with default no

dshymanskyi-vivasg:
Left only "http_strict" and "same_site" cookie headers. 
Now its like
`

      "type": "websocket",
      "cookie": {
           "store": {
                "type": "file",
                "filename": "cookies.dat"
            },
            "http_strict": true,
            "same_site": "Strict" // Can only be one of 'Strict'|'Lax'|'None'
      }
`

oberstet:
yeah, this looks great! thanks for the changes.

so let's take this chance and really nail this. I have reread through

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies

it seems, there are actually 3 modifiers we might want a user to enable?

* SameSite
* HttpOnly
* Secure

if you agree, how about

```
   'secure': (False, [bool]),
   'http_only': (False, [bool]),
   'same_site': (False, [str])
```

for the node config? that way, there is a direct correspondence between configuration option name and HTTP cookie header modifier name.

in my eyes, ^ is desirable, and then this change also covers all options ..

dshymanskyi-vivasg:
Added secure modifier, but what confuse me the most and main reason why i didn't add secure modifier in previous commit was this comment in code
https://github.com/crossbario/crossbar/blob/1e18ceef1b444f60bca118cd5e57eb7ffbd8f2bf/crossbar/router/cookiestore.py#L134

oberstet:
> do NOT add the "secure" cookie attribute! "secure" refers to the scheme of the Web page that triggered the WS, not WS itself!!

mmh, yeah. not sure anymore;) let's see. the option is harmless since it needs to be activated explicitly

dshymanskyi-vivasg:
Well... let it be then :) Thanks for your time and for merge :)

PR #1757, opened 2020-04-25
dynamic proxy frontend authenticators

oberstet:


crossbar/worker/proxy.py
```diff
@@ -469,7 +469,7 @@ def _process_Hello(self, msg):
             )
             try:
                 # call into authenticator for processing the HELLO message
-                hello_result = self._pending_auth.hello(msg.realm, details)
+                hello_result = yield as_future(self._pending_auth.hello(msg.realm, details))
             except Exception as e:
```
meejah, line 473: I think this should be `as_future(self._pending_auth.hello, msg.realm, details)` right? So that the "actual call" happens later? (e.g. on twisted it'll result in an `ensureFuture` call or a `ensureDeferred` if it's a coroutine from `async def`).

oberstet: yeah! that was wrong ..

oberstet:
superseded by https://github.com/crossbario/crossbar/pull/1760

PR #1760, opened 2020-04-28
more proxy / async auth

meejah:


oberstet:
looks good to me .. probably it doesn't catch everything, but let's merge it! only thing is flake8 ..

oberstet:
merged via https://github.com/crossbario/crossbar/pull/1761

PR #1763, opened 2020-04-28
Proxy func authenticator

meejah:
What do you think of this?

It's a `type: function` authenticator, so you provide a (possibly-async) function that returns the actual authentication function. This allows the setup function to be async (e.g. establish database connection) as well as the actual authentication function.

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -492,8 +492,14 @@ def check_transport_auth_cryptosign(config):
     if 'type' not in config:
         raise InvalidConfigException("missing mandatory attribute 'type' in WAMP-Cryptosign configuration")
 
-    if config['type'] not in ['static', 'dynamic']:
-        raise InvalidConfigException("invalid type '{}' in WAMP-Cryptosign configuration - must be one of 'static', 'dynamic'".format(config['type']))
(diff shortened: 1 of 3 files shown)


meejah:
For  `expose_controller=` we could pass an additional arg to `create_authenticator()`

meejah:
Okay I've also added some "controller" stuff now .. (that part still has an issue that `realm is None` when we try to do the controller call.

oberstet:
so the `type==function` stuff looks great! yeah, this should allow us to run custom python authenticator easily in the proxy workers. =)

---

rgd expose_controller: sorry, I should have been more precise, I was thinking slightly different:

* I think the best place for that code is not here, but in a user provided custom authenticator code
* that user code could make use of the "expose_controller" option .. but
* there is the "realm is None" issue you've seen .. and actually, I'm no longer certain we should use this route.
* the alternative would be to simply open an uplink connection to the master node from the authenticator _user_ code

---

so in summary, I am +1 on: removing the couple of lines rgd expose_controller and then merge this PR, and give it a shot and configure it on proxy workers in a test bed ..

oberstet:
merged via https://github.com/crossbario/crossbar/pull/1764

PR #1778, opened 2020-06-12
ensure we always have publisher/caller information

meejah:
See also https://github.com/crossbario/autobahn-python/pull/1393 (which is needed to make this work).


oberstet:
so this is just a stop-gap "solution", until we have rectified https://github.com/crossbario/autobahn-python/issues/1392, right?

because all of `sessionid`, `authid` and `authrole` MUST (well, should, because of of above unfixed issue) always be available to rlinks

PR #1785, opened 2020-06-29
Allow cryptosign-key to be set by env-variable

jegger:
This allows to set the (or more than one) public-key by env-variable.

I read that you need a "Contributor Assignment Agreement" to merge the code. Before I go trough that I want to check out if that PR makes sense to you and if you want that to functionality in your code-base.


oberstet:
rgd the change, my view: -1

the ability to read values in a static node configuration file was intended as a quick way solely for secrets in static authentication config items - not as a general facility for all kinds of value (eg for public keys or other identifiers). I don't think this would be a sustainable approach (like using a dynamic authenticator and read whatever/from-wherever or remote management via crossbarfx).

PR #1815, opened 2020-09-26
fix: remove None value

stone99:
None value lead this error:
```
  File "d:\programdata\anaconda3\envs\skynet\lib\site-packages\autobahn\wamp\websocket.py", line 60, in onOpen
    self._session.onOpen(self)
  File "d:\programdata\anaconda3\envs\skynet\lib\site-packages\crossbar\router\session.py", line 338, in onOpen
    self._transport._transport_info[u'channel_id'] = binascii.b2a_hex(channel_id).decode('ascii')
TypeError: 'NoneType' object does not support item assignment
```


oberstet:
this is not a solution ..

PR #1817, opened 2020-10-05
Fix typo on role attribute

CanNuhlar:
Previously authrole was written as atuhrole on router configuration page. This PR fixes that typo.


oberstet:
welcome, and thanks for contributing!

Issue #1819, opened 2020-10-27
PyTrie 0.4.0 drops python2 support


henxing:
The `requirements-latest.txt` and `requirements-min.txt` files need to be updated to restrict the version of PyTrie crossbar depends on to `pytrie>=0.3, <0.4.0`, since version `0.4.0` drops python 2 support.

### crossbar version output
```
$ crossbar version
/usr/local/lib/python2.7/dist-packages/OpenSSL/crypto.py:12: CryptographyDeprecationWarning: Python 2 is no longer supported by the Python core team. Support for it is now deprecated in cryptography, and will be removed in a future release.
  from cryptography import x509
     __  __  __  __  __  __      __     __
    /  `|__)/  \/__`/__`|__) /\ |__)  |/  \
    \__,|  \\__/.__/.__/|__)/~~\|  \. |\__/

 Crossbar.io        : 18.4.1 (Crossbar.io COMMUNITY)
   Autobahn         : 19.10.1 (with JSON, MessagePack, CBOR, UBJSON)
   Twisted          : 20.3.0-EPollReactor
   LMDB             : 1.0.0/lmdb-0.9.24
   Python           : 2.7.17/CPython
 Frozen executable  : no
 Operating system   : Linux-5.4.0-1028-aws-x86_64-with-Ubuntu-18.04-bionic
 Host machine       : x86_64
 Release key        : RWRoWUpWxaQIXo+I/gl5/ghE46xcFKM4YXQlHck9kRBNLQI+rElVETMF
```

### Steps to reproduce

```
$ pip install crossbar
...
$ pip install pytrie==0.4.0
...
$ crossbar version
Traceback (most recent call last):
  File "/usr/local/bin/crossbar", line 11, in <module>
    sys.exit(run())
  File "/usr/local/lib/python2.7/dist-packages/crossbar/controller/cli.py", line 1114, in run
    if options.personality not in get_installed_personalities():
  File "/usr/local/lib/python2.7/dist-packages/crossbar/controller/cli.py", line 67, in get_installed_personalities
    from crossbar.controller.personality import Personality as CommunityPersonality
  File "/usr/local/lib/python2.7/dist-packages/crossbar/controller/personality.py", line 37, in <module>
    from crossbar.controller.node import Node
  File "/usr/local/lib/python2.7/dist-packages/crossbar/controller/node.py", line 61, in <module>
    from crossbar.router.router import RouterFactory
[...]

PR #1820, opened 2020-10-27
Pins pytrie dependency (1819)

henxing:
[PyTrie](https://pypi.org/project/PyTrie/) version 0.4.0 removes python 2 support, this restricts
valid versions of the pytrie module to `pytrie>=0.3, <0.4.0`.


oberstet:
crossbar does not support python 2

https://github.com/crossbario/crossbar/blob/5aa0682c6ef1edee8ede920873e16571bf4e68a6/setup.py#L105

so using pytrie 0.4 is fine

oberstet:
btw: thanks for CAA we received!

also, I had a look at the CI run for the last crossbar release

https://travis-ci.com/github/crossbario/crossbar/jobs/369843467#L380

and that seems to still had run PyTrie 0.3.1

are you saying you have run into a problem with pytrie 0.4 on python 3. if so, then we need to examine that definitely! however, as said, crossbar does not support python 2 (it did very long ago ..)

henxing:
My mistake @oberstet, I was under the impression that it did support Python 2. Issue #1819 was noticed while running against Python 2, I assume it isn’t an issue for Python 3.

henxing:
@oberstet one comment, the [installation docs](https://crossbar.io/docs/Installation-on-Linux/) still reference Python 2.7, maybe those should be updated?

Issue #1822, opened 2020-10-30
msgpack not listed in requirements-min, requires separate installation


Emantor:
Currently crossbar does not contain `msgpack` in requirements-min.txt, but unconditionally requires it through import of the autobahn `MsgPackObjectSerializer`. Stacktrace from a CI run of labgrid without `msgpack` being explicitly added:
```
crossbar: /home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/OpenSSL/crypto.py:12: CryptographyDeprecationWarning: Python 3.5 support will be dropped in the next release of cryptography. Please upgrade your Python.
crossbar:   from cryptography import x509
crossbar: 
crossbar: Traceback (most recent call last):
crossbar:   File "/home/travis/virtualenv/python3.5.7/bin/crossbar", line 10, in <module>
crossbar:     sys.exit(run())
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/__init__.py", line 175, in run
crossbar:     _personalities = personalities()
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/__init__.py", line 213, in personalities
crossbar:     from crossbar import personality as standalone
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/personality.py", line 41, in <module>
crossbar:     from crossbar.worker.router import RouterController
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/worker/router.py", line 30, in <module>
crossbar:     from crossbar.worker.transport import TransportController
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/worker/transport.py", line 49, in <module>
crossbar:     from crossbar.bridge.mqtt.wamp import WampMQTTServerFactory
crossbar:   File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/crossbar/bridge/mqtt/wamp.py", line 52, in <module>
crossbar:     from autobahn.wamp.serializer import JsonObjectSerializer, MsgPackObjectSerializer, CBORObjectSerializer, UBJSONObjectSerializer
crossbar: ImportError: cannot import name 'MsgPackObjectSerializer'
[...]

PR #1823, opened 2020-10-30
Remove the zlmdb and numpy dependencies from crossbar

Emantor:
Remove numpy and zlmdb from the crossbar dependencies and remove the version checking code for them.
zlmdb was used for the time_ns function, which is now imported from txaio since 2f26fef9de4ac87f39291593a777fdd5f88cc8ef.


Emantor:
Looks like the CI fails because of #1822.

oberstet:
these dependencies should not be removed: we plan to expand both their use in the codebase

Issue #1618, opened 2019-08-14
Crossbar fails to start on Linux when CONFIG_TASK_IO_ACCOUNTING is disabled


pdcastro:
Seeing as Crossbar is able to run on OSX where the kernel does not provide the "IO Accounting" feature (Crossbar [commit](https://github.com/crossbario/crossbar/commit/57a2393f62c32723b72cc9fcb5a26eb0a5365f57) and [OSX issue](https://github.com/crossbario/crossbar/issues/1543)), I would expect it to be able to run on a Linux system where where the CONFIG_TASK_IO_ACCOUNTING kernel option was disabled. But instead an error is raised:

![3312b065-e6d8-45e9-a745-4cef84540dd0](https://user-images.githubusercontent.com/15091591/63057253-baa7f800-bee1-11e9-8b73-1a4f13195d5c.png)

Crossbar was being tested on a VirtualBox VM running a balenaOS release (2.38.0) for the Intel NUC device:
* https://www.balena.io/docs/learn/getting-started/intel-nuc/nodejs/
* https://www.balena.io/blog/no-hardware-use-virtualbox/

A balena issue was also created for CONFIG_TASK_IO_ACCOUNTING to be enabled in the Linux kernel: https://github.com/balena-os/meta-balena/issues/1617
... but a new Crossbar release addressing this issue might benefit a larger number of users, sooner.

Thanks!

oberstet:
Ok, I see: could you pls post the complete traceback (just paste the whole log eg)?

mvenilton:
@oberstet, this is the complete traceback:

[Controller     15] Could not start node: Traceback (most recent call last):
    File "/usr/local/lib/python3.7/dist-packages/twisted/internet/base.py", line 447, in _continueFiring
      callable(*args, **kwargs)
    File "/usr/local/lib/python3.7/dist-packages/crossbar/node/main.py", line 878, in start_crossbar
      d = node.start()
    File "/usr/local/lib/python3.7/dist-packages/twisted/internet/defer.py", line 1613, in unwindGenerator
      return _cancellableInlineCallbacks(gen)
    File "/usr/local/lib/python3.7/dist-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
      _inlineCallbacks(None, g, status)
  --- <exception caught here> ---
    File "/usr/local/lib/python3.7/dist-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
      result = g.send(result)
[...]

PR #1826, opened 2020-11-09
Check for io_counters feature

lmcgartland:
This resolves [#1618](https://github.com/crossbario/crossbar/issues/1618) by checking for the attribute 'io_counters' before trying to access it on Linux systems.


oberstet:
thanks for fixing this!

the CI fails are unrelated to this issue: seems like pypy/cryptography now wants a more recent openssl on the travis buildbox https://travis-ci.com/github/crossbario/crossbar/jobs/430892567#L461

.. we'll fix that in a different change

PR #1833, opened 2020-12-14
Unpin urllib3

sjustas:
Please consider increasing pinned version pin, 1.25+ is a requirement to more and more libraries recently.  For example:  [botocore](https://github.com/boto/botocore/blob/develop/setup.py#L32), [responses](https://github.com/getsentry/responses/blob/master/setup.py#L26)

This PR is probably an incorrect/incomplete way to unpin it.


oberstet:
crossbar does not depend on botocore, responses, and we don't support installing crossbar into a python environment shared with other stuff. one reason is exactly such version conflicts. it is not a viable approach to try to figure out a mix of library versions that works with crossbar and an unknown set of other user applications for us.

sjustas:
Makes sense.  Thanks for tanking your time to answer!

PR #1855, opened 2021-02-11
An attempt to fix dependency hell

fijal:


oberstet:
rgd installing latest autobahn from master .. not sure, maybe it was related: I've fixed a bad bug yesterday: https://github.com/crossbario/autobahn-python/commit/500898710c10e9f0eb92b424b40031ecd2c9dd6d

AB "xbr" flavor had listed AB itself (at least a certain version) as a dependency!

PR #1869, opened 2021-03-24
Add the ability to run crossbar docker images on x86 (not recommended)

fijal:
A few notes:

* I can't build the aarch64 images on x86 due to rust issues. I'm not sure if this is a problem with my machine or a deeper problem, I did not investigate
* 32bit arm is untested, maybe we remove the arm target completely?
* qemu is now copied from docker and not from system. That works better (works at all on my machine) and makes it independent of version of qemu installed. There is probably some cleanup to be done on github workflows for that
* binaries are installed in /opt/pypy/bin and not /usr/local/bin for crossbar wheel, I did not investigate why

### DIFF
--- a/docker/Dockerfile.cpy-slim
+++ b/docker/Dockerfile.cpy-slim
@@ -2,6 +2,8 @@
 # however, CROSSBAR_ARCH will _not_ be available further down the line because of this ordering!
 ARG CROSSBAR_ARCH
 
+FROM multiarch/qemu-user-static:x86_64-aarch64 as qemu
+FROM multiarch/qemu-user-static:x86_64-arm as qemuarm
 FROM ${CROSSBAR_ARCH}/python:3.9-slim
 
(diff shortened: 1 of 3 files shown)


oberstet:
> 32bit arm is untested, maybe we remove the arm target completely?

yes, we could remove that altogether (not needed)

> I can't build the aarch64 images on x86 due to rust issues.

not sure I understand .. yes, I am aware that a recent rust is needed. I use upstream (not distro pkg), and that works for me:

https://github.com/crossbario/crossbarfx/blob/d8f969f003a2fbabd400d4ac9774c778c393149f/docker/Dockerfile.cpy-slim#L67

> binaries are installed in /opt/pypy/bin and not /usr/local/bin for crossbar wheel,

not a problem - as long as /opt/pypy/bin is on the PATH then

> qemu is now copied from docker and not from system

ok, also good .. if it works

PR #1873, opened 2021-04-07
Add a meta-event for specific registration URI

stephenmdangelo:
I'm not sure if this is way out-of-line as a feature request, as it does get into WAMP specification territory. Just figured this was the easiest way to see if this was something that would be considered or not. Feel free to toss it out if it's a bad idea!

Mainly, I have Components that use a combination of "wamp.registration.lookup" and subscribing to the "wamp.registration.on_create/delete" events to determine if an end-point is available to be called. The deletion/creation of an end-point is used for a Component to know it may need to re-load some data from the other Component. Unfortunately, the "wamp.registration.on_create" event is very noisy in the case that many Components are coming online at once. Ultimately, I only care about a few select end-points, so it would be nice to be able to subscribe to specific end-point registration events and let Crossbar do the filtering instead of the client Components.

Other options are to either add a single Component that translates the "on_create" into the individual per-URI events for my specific use-case, but then I have to manage that Component, and ensure it scales as well as Crossbar, or I can have the individual Components emit application-specific "I'm ready" events when they register their end-points, but the "on_create" meta-event is already there and feels so close to what I need.

Maybe my issue is just that I'm relying on these introspection/debug events for actual application logic, and ultimately application-specific events are the way to go.

Corresponding on_delete, on_subscribe, on_unsubscribe could, of course, be added if this is actually something you are interested in doing.

### DIFF
--- a/crossbar/router/dealer.py
+++ b/crossbar/router/dealer.py
@@ -524,6 +524,12 @@ class Dealer(object):
                                     registration_details,
                                     options=options
                                 )
+                                service_session.publish(
+                                    'wamp.registration.on_create.' + registration.uri,
+                                    session._session_id,
+                                    registration_details,
+                                    options=options
+                                )
 
                         if not was_already_registered:
                             if options:


oberstet:
thanks for the detailed description of background/context and the "why?". much appreciated!

> Maybe my issue is just that I'm relying on these introspection/debug events for actual application logic, and ultimately application-specific events are the way to go.

Yes, agreed, this is an important non-goal to avoid and keep in mind. There is no "yes / no" answer .. it depends.

Having said that: 

> Ultimately, I only care about a few select end-points, so it would be nice to be able to subscribe to specific end-point registration events and let Crossbar do the filtering instead of the client Components.

seems like a natural, next-step wish given the nature of the WAMP meta API.

only subscribing to the WAMP meta API events associated with a given single subscription (for pubsub) or registration (for rpc) rather than _all_ such events in a realm, which can be overwhelming.

couple of aspects to discuss (well, apart from also spec text):

* `wamp.registration.on_create.<registration-id>` vs `wamp.registration.object.<registration-id>.on_create`
* `registration.uri` vs `registration.id`

stephenmdangelo:
> registration.uri vs registration.id

In regards to this, registration.uri is the important thing here, as that's the fixed identifier that the client will be attempting to call. As the registration.id is ephemeral, it doesn't help much if that's what it in the topic.

> only subscribing to the WAMP meta API events associated with a given single subscription (for pubsub) or registration (for rpc) rather than all such events in a realm, which can be overwhelming.

Yeah, I feel like what I am attempting to do here is move from a common TCP-based workflow, where the TCP connection status is the notifier of a service being ready or not. I'm trying to avoid having to add concepts that are traditionally handled by TCP already such as "ping" and "connection established" to the application layer.

oberstet:
> In regards to this, registration.uri is the important thing here, as that's the fixed identifier that the client will be attempting to call. As the registration.id is ephemeral, it doesn't help much if that's what it in the topic.

including the URI doesn't work, since subscriptions in WAMP can also be for URI patterns.

instead, one can use this meta API

* wamp.registration.match
* wamp.registration.lookup
* wamp.subscription.match
* wamp.subscription.lookup

to deal with IDs

Issue #1226, opened 2017-11-23
Document systemd notification mechanism


oberstet:
Some time ago Crossbar added support for the systemd notify mechanism. This 
mechanism tells systemd to wait for the service to notify systemd that it 
is up and running before loading dependent services. Normally systemd just 
starts the process and assumes that's enough, but in the case of Crossbar 
and a client Autobahn application this isn't good enough, at least not on 
all systems.

Here's the PR related to this change: 
https://github.com/crossbario/crossbar/pull/637

The example pages for setting up Crossbar to run as a service does not 
mention using this feature: 
https://crossbar.io/docs/Automatic-Startup-and-Restart/ 

The example probably ought to show using "type=notify" instead of 
type="simple".

CanNuhlar:
Hello, I hope to solve this with #1883

PR #1883, opened 2021-06-28
Add systemd-notify support to docs

CanNuhlar:
This PR aims to solve issue #1226


CanNuhlar:
Any idea why pipeline has failed? Took a look but I don't think it's related to this PR.

Issue #1887, opened 2021-08-02
master node fails to start while merging lists


om26er:
Running the command `crossbar master start` fail with the below error. However if used with the exact config available here https://github.com/crossbario/crossbar/blob/master/crossbar/master/node/config.json it starts fine.
```
2021-08-03T03:35:45+0500 [Controller 238438] 
2021-08-03T03:35:45+0500 [Controller 238438]     :::::::::::::::::
2021-08-03T03:35:45+0500 [Controller 238438]           :::::          _____                 __              _____  __
2021-08-03T03:35:45+0500 [Controller 238438]     :::::   :   :::::   / ___/______  ___ ___ / /  ___ _____  / __/ |/_/
2021-08-03T03:35:45+0500 [Controller 238438]     :::::::   :::::::  / /__/ __/ _ \(_-<(_-</ _ \/ _ `/ __/ / _/_>  <
2021-08-03T03:35:45+0500 [Controller 238438]     :::::   :   :::::  \___/_/  \___/___/___/_.__/\_,_/_/   /_/ /_/|_|
2021-08-03T03:35:45+0500 [Controller 238438]           :::::
2021-08-03T03:35:45+0500 [Controller 238438]     :::::::::::::::::   Crossbar.io FX v21.6.1.dev1 [19000101-0000000]
2021-08-03T03:35:45+0500 [Controller 238438] 
2021-08-03T03:35:45+0500 [Controller 238438]     Copyright (c) 2013-2021 Crossbar.io Technologies GmbH. All rights reserved.
2021-08-03T03:35:45+0500 [Controller 238438] 
2021-08-03T03:35:45+0500 [Controller 238438] Booting master node .. <crossbar.node.main._run_command_start>
2021-08-03T03:35:45+0500 [Controller 238438] Node key files exist and are valid. Node public key is 0xf24b4e22f449e0ad14a17936ac58f53534ed5cb0427b9d1fc6e2a05797e9b982
2021-08-03T03:35:45+0500 [Controller 238438] Node key loaded from /home/om26er/scm/crossbario/markhor/online-services/.crossbar/key.priv
2021-08-03T03:35:45+0500 [Controller 238438] Expanding built-in node configuration from local file /home/om26er/scm/crossbario/markhor/online-services/.crossbar/config.json
2021-08-03T03:35:45+0500 [Controller 238438] Traceback (most recent call last):
[...]

PR #1888, opened 2021-08-03
Add default template for master config

om26er:
Adds a default config for crossbar master when `crossbar master init` is run. Note the json config is a copy of https://github.com/crossbario/crossbar/blob/master/crossbar/master/node/config.json

One alternative could be to create a `template` directory inside https://github.com/crossbario/crossbar/blob/master/crossbar/master/node/ and put the existing config.json there inside a `.crossbar` subdirectory.

If this is an acceptable change, a similar thing could be done for `edge` as well.

### DIFF
--- a/crossbar/node/main.py
+++ b/crossbar/node/main.py
@@ -554,7 +554,10 @@ def _run_command_init(options, reactor, personality):
 
     log.info("Initializing application directory '{options.appdir}' ..", options=options)
 
-    get_started_hint = Templates.init(options.appdir, template='default')
+    if personality.NAME == 'master':
+        get_started_hint = Templates.init(options.appdir, template='master')
+    else:
+        get_started_hint = Templates.init(options.appdir, template='default')
 
     _maybe_generate_key(cbdir)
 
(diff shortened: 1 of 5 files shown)


om26er:
After deeper dive, I think this PR doesn't make much sense as the "master" config is fundamentally different from both `edge` and `standalone` and it's parsing is done very differently.

For the master node the default [config](https://github.com/crossbario/crossbar/blob/master/crossbar/master/node/config.json) is merged with the user provided config. However not many values can be overridden. The config merging code does have a [bug](https://github.com/crossbario/crossbar/issues/1887) OR the error could at least be more informative.

oberstet:
^ The treatment of config for different node types are indeed confusing, and lacks. The reasons are historic changes, and multiple stages of development. I agree rgd the change here: it doesn't really make sense - we should clean up this stuff once and completely.

Rgd the node merging functionality: this is one generic approach I tried to allow a user to override some things like TLS certs, while still using most of the built in standard config of the respective node (master, edge, core, ..). I even added unit tests for the merging (it does work "ok" ..). I still think this makes sense, but if there is a better way, np ..

om26er:
This would need a new issue and more discussions. Closing the PR now

Issue #1902, opened 2021-11-11
rlinks: Allow custom cryptosign key


om26er:
Currently the cryptosign key used for router to router links is of the Crossbar node. This should be configurable. When a router link is being started, it should be possible to pass private key hex.

oberstet:
yes, this is by design:

* the identity of the node at the WAMP rlink authentication level is via the node key (there is exactly 1 per node, always)
* the identity of the node at the cross-domain (inter-operator) rlink level is via an Ethereum private key - pls see page 2 here https://github.com/wamp-proto/wamp-proto/blob/457b92a47e2615d4cdf67dee07d82eceb054bb94/wep/wep002/wamp_ap_r2r.pdf

I've marked the issue as "for discussion" nevertheless. Rgd the requirements I had in mind, a good question to start with is:

* Say you operate node-1, node-2 and node-3, and want to establish a full mesh of rlinks. all nodes are managed as one logical units (as eg in crossbar master). the pubkeys accepted and used should be tied to the nodes - otherwise it'll lead to potentially quadratic number of keys .. and why (for what gain)?
* Say someone else operates node-4 and you want to establish an rlink from each of your nodes (or any one) to that node-4. How do you know the pubkey of node-4 to accept, how does node-4 know the pubkeys of your nodes?

om26er:
One scenario for that is the Edge, where the private key of the hardware (through secure element) is it's identity. In such scenarios it can be useful to have requested flexibility.

oberstet:
no, private keys should not be part of the rlink api. and it would work like that anyways for hardware wallets:

* if the private key is in a hardware wallet, exporting that key is usually impossible, and importing the private key (even if it would be known) is wrong anyways, since the hardware wallet should sign
* hence, what we need is support for signing by hardware wallet in autobahn

oberstet:
hardware signing can be added to CB via a new (sub)class for SigningKey

https://github.com/crossbario/autobahn-python/blob/a35f22eeaafca7568f1deb35c4a1b82ae78f77d4/autobahn/wamp/cryptosign.py#L357

om26er:
@oberstet So there are two scenarios that I can think of

1. A real hardware wallet (for production)
2. A mock hardware wallet (for development and testing. This basically incorporates a private key hex)

Once we have something like that in place in autobahn. I guess CB would need to be changed somehow.

Would it make sense for CB to optionally support setting the "signing key" object on the controller ?

Below is what I am thinking
```
[...]

PR #1906, opened 2021-11-16
implement dynamic node key

om26er:
This is work in progress as it needs to be extended so that crossbar workers also read the "correct" private key.

### DIFF
--- a/crossbar/node/controller.py
+++ b/crossbar/node/controller.py
@@ -23,8 +23,9 @@ from twisted.internet.error import ProcessExitedAlready
 from twisted.python.runtime import platform
 
 from autobahn.util import utcnow, utcstr
+from autobahn.wamp.cryptosign import format_challenge, sign_challenge
 from autobahn.wamp.exception import ApplicationError
-from autobahn.wamp.types import PublishOptions, ComponentConfig
+from autobahn.wamp.types import PublishOptions, ComponentConfig, Challenge
 from autobahn import wamp
 
 import crossbar
@@ -415,6 +416,17 @@ class NodeController(NativeProcess):
             raise ApplicationError('crossbar.error.no_such_worker', emsg)
 
         return self._workers[worker_id].getlog(limit)
+
+    @wamp.register(None)
+    def sign_challenge(self, challenge_method, challenge_extra, channel_id_raw, channel_id_type='tls-unique',
+                       details=None):
+        challenge = Challenge(challenge_method, challenge_extra)
+        data = format_challenge(challenge, channel_id_raw, channel_id_type)
+        return sign_challenge(data, self._node._node_key.sign)
+
+    @wamp.register(None)
+    def get_public_key(self, details=None):
+        return self._node._node_key.public_key()
 
     def _start_native_worker(self, worker_type, worker_id, worker_options=None, details=None):
 
--- a/crossbar/node/controller.py
+++ b/crossbar/node/controller.py
@@ -418,7 +418,11 @@ class NodeController(NativeProcess):
         return self._workers[worker_id].getlog(limit)
 
     @wamp.register(None)
-    def sign_challenge(self, challenge_method, challenge_extra, channel_id_raw, channel_id_type='tls-unique',
+    def sign_challenge(self,
+                       challenge_method,
+                       challenge_extra,
+                       channel_id_raw,
+                       channel_id_type='tls-unique',
                        details=None):
         challenge = Challenge(challenge_method, challenge_extra)
         data = format_challenge(challenge, channel_id_raw, channel_id_type)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -136,9 +136,6 @@ jobs:
       run: |
         pip install -U .
         pip install -r requirements-latest.txt
-        cat /opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/autobahn/wamp/cryptosign.py
-        python -c 'from autobahn.wamp.cryptosign import HAS_CRYPTOSIGN; print("HAS_CRYPTOSIGN =", HAS_CRYPTOSIGN)'
-        python -c 'from autobahn.wamp.cryptosign import SigningKeyBase'
 
     - name: Display Crossbar.io version
       run: |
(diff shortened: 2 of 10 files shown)


oberstet:
yeah, this is how to expand the CLI to worker options passing. unfortunately, it won't work .. pls see https://github.com/crossbario/crossbar/issues/1902#issuecomment-971165831 =(

crossbar/node/controller.py
```diff
@@ -416,6 +417,21 @@ def get_worker_log(self, worker_id, limit=100, details=None):
...
         return self._workers[worker_id].getlog(limit)
 
+    @wamp.register(None)
+    def sign_challenge(self,
+                       challenge_method,
+                       challenge_extra,
+                       channel_id_raw,
+                       channel_id_type='tls-unique',
+                       details=None):
+        challenge = Challenge(challenge_method, challenge_extra)
```
om26er, line 427: This change was made to address https://github.com/crossbario/autobahn-python/pull/1501/files#r752058376 i.e. without changing autobahn-python

om26er:
This PR is still not ready. However, after https://github.com/crossbario/autobahn-python/pull/1501 lands, this PR at least allows rlinks to use the key instance defined in Crossbar json config.

om26er:
The CI is blocked because it seems crossbar is not able to find the new class that was introduced in latest autobahn-python. Specifically `SigningKeyBase`.

The relevant traceback looks like
```
ImportError: cannot import name 'SigningKeyBase' from 'autobahn.wamp.cryptosign' 
```

oberstet:
>  is not able to find the new class that was introduced in latest autobahn-python

yeah, the test needs to run on an installation of AB from the GH master branch for that to work .. not sure, the CI setup

https://github.com/crossbario/crossbar/blob/93f3931eb3fb34bae59c6a8d3d4bcbbef7131d67/.github/workflows/main.yml#L137

installs

https://github.com/crossbario/crossbar/blob/93f3931eb3fb34bae59c6a8d3d4bcbbef7131d67/requirements-latest.txt#L2

`SigningKeyBase` is on AB master now, right?

oberstet:
does it work if you pip install in a fresh venv `git+https://github.com/crossbario/autobahn-python.git@master#egg=autobahn[twisted,encryption,compress,serialization,scram,xbr]`?

another trap: above works, but not in the CI - maybe because in the CI the AB package is already installed when it hits above requirements-latest? such stuff can be fun ... not sure if it applies here

om26er:
Yes, here https://github.com/crossbario/autobahn-python/blob/062e5e6492beb92292934455ae6a95bce4a0da19/autobahn/wamp/cryptosign.py#L432

I took a deeper look it seems the file `/opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/autobahn/wamp/cryptosign.py` on matterhorn doesn't really have that class. So it's likely being imported from the wrong location or there is a difference between which interpreter is being used

oberstet:
you could add the failing import in the CI before the actual testsuite starts here https://github.com/crossbario/autobahn-python/blob/062e5e6492beb92292934455ae6a95bce4a0da19/.github/workflows/main.yml#L108

om26er:
> does it work if you pip install in a fresh venv `git+https://github.com/crossbario/autobahn-python.git@master#egg=autobahn[twisted,encryption,compress,serialization,scram,xbr]`?

just tried in a clean env, it works yes. I believe the issue is with the environment.

> you could add the failing import in the CI before the actual testsuite starts here https://github.com/crossbario/autobahn-python/blob/062e5e6492beb92292934455ae6a95bce4a0da19/.github/workflows/main.yml#L108

tried that and it's imported fine https://github.com/crossbario/autobahn-python/runs/4267541701?check_suite_focus=true#step:7:16

om26er:
To verify a few things and probably find a fix, I would probably need SSH access to matterhorn, if you are comfortable with that

om26er:
> To verify a few things and probably find a fix, I would probably need SSH access to matterhorn, if you are comfortable with that

Not needed. I was able to fix this. It seems the issue was caused due to the hard-coded version of pip and autobahn not updating when installed from github because it also expects a change in version inside setup.py (https://stackoverflow.com/a/68432209).

oberstet:
@om26er pip again. we will see how the now current pip version behaves - I added the pip version pinning because pip changed how multiple confluent deps are resolved, which broke xbr related stuff. forgot the details. we will see. python deps are like the GIL .. endless, never really fixed.

oberstet:
I had a look through the changes again at this point, and this already looks quite good! this will work, and fits into nicely=)

The one thing we need to discuss/change IMO is `check_node_key`. The node config item for the configuration of the way and the details of accessing node key/signing. Here is a proposal:

*File-based*:

```
{
   "controller": {
      "nodekey": {
         "type": "file",
         "path": "/home/homer/key.bin"
      }
   }
}
```

*HW-based*:

```
{
   "controller": {
      "nodekey": {
         "type": "hsm",
         "driver": "nxp-se050",
         "port": "/dev/ttyACM0"
      }
   }
}
```

om26er:
Updated the PR to support file-based and hardware keys. This does reduce the flexibility and future extensibility (but is probably more secure ?) i.e. to support a new type of hardware key, we would have to make changes to both crossbar and autobahn.

Also for the NXP chip, it seems there are no python bindings, so that's something we'd probably end up writing and eventually integrated into CB/AB. On our hardware the secure element is actually accessible on the I2C interface (specifically /dev/i2c-6)

oberstet:
> This does reduce the flexibility and future extensibility (but is probably more secure ?)

no, would disagree with this,  it doesn't reduce future extensibility, but properly defines a narrow interface to extend the system

eg allowing to specify a class name is bad design (in this case .. for a container worker wamp class, that's different)

a configuration item is an interface. the purpose of an interface is primarily to structure a big system into loosely coupled parts. narrow/strict interfaces make it loosely coupled

a configuration item of course also transports some info .. the minimal info, constrained to the maximum

> Also for the NXP chip, it seems there are no python bindings

really? forgot everything already;) in any case, should be easy .. just use a I2C library (take care of async or wrappers) and bit bang the chip's registers.

eg a I2C lib like this https://github.com/kplindegaard/smbus2 or similar .. depends also on hosting env etc

om26er:
Do you think we should land this now. A follow up PR that I will propose tomorrow will take care of making sure all workers use the same mechanism to access the `key`.

This would allow us to also land a few other PRs that are blocked due to failing CI

oberstet:
sure! AB is merged, this one runs green .. merged. thanks a lot =)

PR #1912, opened 2021-11-21
fix cookiestore.py utc

antonymott:
match utc creation method in _create() and _clean_cookie_file()

### DIFF
--- a/crossbar/router/cookiestore.py
+++ b/crossbar/router/cookiestore.py
@@ -85,7 +85,7 @@ class CookieStore(object):
         # cookie tracking data
         cbtData = {
             # UTC timestamp when the cookie was created
-            'created': util.utcnow(),
+            'created': datetime.datetime.now(datetime.timezone.utc),
 
             # maximum lifetime of the tracking/authenticating cookie
             'max_age': self._cookie_max_age,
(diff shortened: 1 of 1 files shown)


oberstet:
no,  the format must be ISO 8601, and the current code is correct

https://github.com/crossbario/autobahn-python/blob/9de497273c6db50d8697019844f6a7fd3a867789/autobahn/util.py#L166

```
(cpy39_1) oberstet@intel-nuci7:~$ python
Python 3.9.4 (default, Apr 29 2021, 02:30:43) 
[GCC 9.3.0] on linux
>>> import datetime
>>> datetime.timezone
<class 'datetime.timezone'>
>>> datetime.timezone.utc
datetime.timezone.utc
>>> datetime.datetime.now(datetime.timezone.utc)
datetime.datetime(2021, 11, 22, 8, 32, 52, 286101, tzinfo=datetime.timezone.utc)
>>> str(datetime.datetime.now(datetime.timezone.utc))
'2021-11-22 08:33:07.006501+00:00'
>>> from autobahn import util
>>> util.utcnow()
'2021-11-22T08:33:30.639Z'
>>> 
```

Issue #1916, opened 2021-11-24
rlinks: duplicate events after reconnect


asimfarooq5:
If the connection between instance A and B is lost, On connection resume/resubscribe the event causes duplicate results for number of times it reconnects.

The same configuration can be used to reproduce the bug #1909  

* Start both Instances A and B
* Subscribe event on instance A and B
* Publish event to instance A and B
* get the same event on both instances
* stop/restart instance B then resubscribe
* Now, when the event is published to instance A, this will result in duplicate events on instance B

om26er:
This issue is real, I can reproduce that. It seems the local leg has the registrations even when the remote end disconnected/stopped. In such event the following code is not called https://github.com/crossbario/crossbar/blob/01ec53a100b08a4b78adabe26ab0abb12ba21d5a/crossbar/worker/rlink.py#L167

So we probably need to update `on_remote_leave` to unsubscribe to the remote events https://github.com/crossbario/crossbar/blob/01ec53a100b08a4b78adabe26ab0abb12ba21d5a/crossbar/worker/rlink.py#L204

oberstet:
As a note, it can be daunting/confusing to think about and keep all message interactions for a certain scenario  in mind with this rlink stuff. In particular, since those WAMP messages happen at the meta level, and at the app level.

* pls consider https://github.com/crossbario/crossbar/blob/master/docs-cfx/_static/crossbar-rlinks.pdf - I tried to make this stuff approachable, and ..
* This is supposed to intro one graphical notation for documenting what is supposed to happen with rlinks at the WAMP message level in an accurate way.
* If that notation makes sense and is welcome, we could add all message interactions and scenarios

asimfarooq5:
It seems to be fixed with proposed branch #1919.

PR #1919, opened 2021-11-26
Unsubscribe forwarded events on remote disconnect

om26er:
When remote leg disconnects, make sure to unsubscribe all active subscriptions on local leg.

Fixes https://github.com/crossbario/crossbar/issues/1916


oberstet:
fwiw, I'm slightly worried that our current approach of modifying the code in response to problems identified in adhoc, random tests, and without adding both "positive, supposed to work" and CI/unit/regression tests isn't sustainable ...

PR #1923, opened 2021-12-07
dont construct the signing key in worker controller

om26er:
The worker controller file now has `get_public_key` and `sign_challenge` methods. Don't try to construct the SingingKey object manually as that's redundant and may cause issues


om26er:
@oberstet lets land it ?

oberstet:
sure, great! only missed the notification ..

PR #1954, opened 2022-03-02
Polish clustering

oberstet:


om26er:
What about the banner, is that something to update as well ? https://github.com/crossbario/crossbar/blob/76bd11e82d6b8a0307b49c8ca9b34a856197220e/crossbar/edge/personality.py#L355

oberstet:
yes, good point! https://github.com/crossbario/crossbar/pull/1954/commits/2b0bcc0f54bb044dcd5f6b7dbb700284b81805e6

PR #1968, opened 2022-03-23
Fix proxy controller has_realm() logic

om26er:
`ProxyController.has_realm()`, checks if the route for a realm exists or not. Initially that function works correctly when no route exists, however when a route is started and later stopped, that function still returns `True` because the code only checks for the existence of a key. It also needs to check if the key returns an empty dictionary or not.

Here is the relevant part which actually removes a route https://github.com/crossbario/crossbar/blob/e88876f47f12a830d76d225723e0b8d13fdc589c/crossbar/worker/proxy.py#L1653.

### DIFF
--- a/crossbar/worker/proxy.py
+++ b/crossbar/worker/proxy.py
@@ -1190,7 +1190,7 @@ class ProxyController(TransportController):
         :returns: True if a route to the realm (for any role) exists.
         :rtype: bool
         """
-        result = realm in self._routes and len(self._routes[realm]) > 0
+        result = realm in self._routes
         self.log.debug('{func}(realm="{realm}") -> {result}',
                        func=hltype(ProxyController.has_realm),
                        realm=hlid(realm),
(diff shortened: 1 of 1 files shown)


oberstet:
> If all routes are stopped, clear the realm from routes map

no, this is wrong

om26er:
ok, in that case, we can close the PR.

The discrepancy here is that the proxy worker has very basic knowledge of a realm. So when both the proxy connection and the proxy realm route is stopped, then at that point, the proxy worker should just forget the realm. However currently it tells the application code something like
```
wamp.error.no_such_role message=realm "omer-home" has no role "owner"
```

So basically when the route is stopped, it "forgets" the role but not the realm itself. expected ?

oberstet:
no, sorry for the confusion: the PR is fine after the change you did, as:

>  "having a realm" == "having at least one route configured for that realm"

om26er:
ah, just saw the other comment, I can do the doc fix

om26er:
Looking at it again, I think the current doc string is already clear enough
```
        """
        Check if a route to a realm with the given name is currently running.

        :param realm: Realm name (the WAMP name, _not_ the run-time object ID).
        :type realm: str

        :returns: True if a route to the realm (for any role) exists.
        :rtype: bool
        """
```

Issue #1971, opened 2022-03-25
proxy routes broken with multiple realm roles


om26er:
If a realm route has only one role, everything works fine. However if there are multiple roles for a realm and the route is configured, the round robin logic is broken
https://github.com/crossbario/crossbar/blob/fa0e418ea143efe6ab360fb5f438e5ebf46e8f15/crossbar/worker/proxy.py#L1386-L1388

In the current implementation, a new realm route is started for each role
https://github.com/crossbario/crossbar/blob/fa0e418ea143efe6ab360fb5f438e5ebf46e8f15/crossbar/node/node.py#L962-L978

However, the round robin logic above doesn't check if a route has the specific role and tries to access it, resulting in a KeyError.

om26er:
Here is a super basic crossbar config for a proxy worker, enough to reproduce the bug described above
```
{
   "version": 2,
   "workers": [
      {
         "type": "proxy",
         "connections": {
            "conn1": {
               "transport": {
                  "type": "rawsocket",
                  "endpoint": {
                     "type": "tcp",
                     "host": "127.0.0.1",
                     "port": 9001
                  },
                  "url": "ws://localhost",
                  "serializer": "cbor"
               }
            }
         },
         "routes": {
            "realm1": {
               "system": "conn1",
               "admin": "conn1"
            }
         },
         "transports": [
            {
               "type": "universal",
               "endpoint": {"type": "tcp", "port": 8080},
               "rawsocket": {
                  "serializers": ["json", "cbor"],
                  "auth": {
                     "anonymous": {
                        "type": "static",
                        "role": "admin"
                     }
                  }
               }
            }
         ]
      }
   ]
}
```

om26er:
Since the code runs as a roundrobin, it "succeeds" one first attempt and fails on the other
```
om26er@P1:~/Documents/backend_bug$ wick publish --url rs://localhost:8080 --realm realm1 hello
received unexpected ABORT message when expecting WELCOME: wamp.error.authentication_failed message=Frontend connection accept failed (Connection was refused by other side: 111: Connection refused.)
om26er@P1:~/Documents/backend_bug$ wick publish --url rs://localhost:8080 --realm realm1 hello
received unexpected ABORT message when expecting WELCOME: wamp.error.authentication_failed message=Frontend connection accept failed ('admin')
om26er@P1:~/Documents/backend_bug$ wick publish --url rs://localhost:8080 --realm realm1 hello
[...]

PR #1972, opened 2022-03-25
Fix proxy routes when multiple roles configured

om26er:
Fixes https://github.com/crossbario/crossbar/issues/1971

Separately, I think the round robin logic there is completely broken.

### DIFF
--- a/crossbar/worker/proxy.py
+++ b/crossbar/worker/proxy.py
@@ -1384,6 +1384,10 @@ class ProxyController(TransportController):
         assert self.has_role(realm_name, role_name)
 
         routes = self._routes[realm_name]
+        # Since we have all the routes for the requested realm, we
+        # need to filter the routes which are for the requested role.
+        # Fix for https://github.com/crossbario/crossbar/issues/1971
+        routes = {route_id: route for route_id, route in routes.items() if role_name in route.config}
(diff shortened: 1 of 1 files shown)


om26er:
> Separately, I think the round robin logic there is completely broken.

What I mean by this is that the existing roundrobin logic assumes that the number of routes is always static. We need to think an alternate approach here because the routes can potentially be dynamic hence it might make sense to have a roundrobin index per realm per role. I can work on that in a follow up branch, if that approach makes sense

oberstet:
> I think the round robin logic there is completely broken.

do you have a unit test that demonstrates that it is broken? without a unit test, broken is undefined, as it works for me: clients can do RPC and PubSub, so what's wrong?

om26er:
Just a clarification, I am talking about the round robin logic inside the ProxyController which decides what proxy route to use for a connection request. This is going to be a difficult task to create an automated test for, do you have any pointers on writing automated tests for this stuff of features in Crossbar ?

And regarding the PR, there is a crossbar config in issue https://github.com/crossbario/crossbar/issues/1971, which makes the relevant bug 100% reproducible.

oberstet:
> do you have any pointers on writing automated tests for this stuff of features in Crossbar 

yes, in this repo in the test folder

PR #1994, opened 2022-04-18
Update main.yml

myoung34:


myoung34:
Ignore =)

myoung34:
Yep, just looking for public repos with self hosted to flag potential issues. I cant tell what gates are in front without opening a PR though

Nothing seems insecure about yours =)

oberstet:
yeah, I figured after checking your profile it would likely be a "sec test";) but thanks for replying. rgd the self-hosted CI from GH stuff: I consider this "semi secure": a first time contributor needs to have an admin look into the (first) PR before that PR is even run. after that, not anymore. which means, a PR can change the workflow and do stuff as the CI user on that box. that user has limited rights. and the credentials for publishing to PyPI or similar are _not_ on the box, so this needs to be done by an admin from his own box. I would consider the credentials handling in workflows the area where stuff could go wrong the "most likely" .. anyways, have a good one!

myoung34:
You too!

What you described is definitely a best practice but a lot of people tend to enable things without due diligence on the gates.

Issue #1908, opened 2021-11-20
rlinks: Implement `stop_link`


om26er:
The `stop_link` API is not implemented. This can be very useful when writing tests for rlinks.

Imagine a static router config with three different router workers, all started but not connected to each other. The ability to start and stop links dynamically could make that we can tests a whole bunch of different scenarios, without too much hassle.

om26er:
Here is the relevant API https://github.com/crossbario/crossbar/blob/b9799d2ac6667a97373db0549d2055fff97d784c/crossbar/worker/rlink.py#L891

oberstet:
Yes, that would be useful to have, agreed. Eg it would also allow to orchestrate more complex integration tests ... some already exist in commented form, eg https://github.com/crossbario/crossbar/blob/b9799d2ac6667a97373db0549d2055fff97d784c/test/test_cli_1.sh#L104

PR #2019, opened 2022-05-31
Implement rlink stop

om26er:
Fix for https://github.com/crossbario/crossbar/issues/1908

### DIFF
--- a/crossbar/worker/rlink.py
+++ b/crossbar/worker/rlink.py
@@ -674,13 +674,14 @@ class RLinkRemoteSession(BridgeSession):
 
 
 class RLink(object):
-    def __init__(self, id, config, started=None, started_by=None, local=None, remote=None):
+    def __init__(self, id, config, started=None, started_by=None, local=None, remote=None, remote_runner=None):
         assert type(id) == str
         assert isinstance(config, RLinkConfig)
         assert started is None or type(started) == int
         assert started_by is None or isinstance(started_by, RLinkConfig)
         assert local is None or isinstance(local, RLinkLocalSession)
         assert remote is None or isinstance(remote, RLinkLocalSession)
+        assert remote_runner is None or isinstance(remote_runner, ApplicationRunner)
 
         # link ID
         self.id = id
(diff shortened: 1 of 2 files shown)


oberstet:
ah, right, stopping rlinks;) couple of notes/feedback:

The PR expands the constructor of `RLink` to forward a reference to `ApplicationRunner` - it could use `ComponentConfig.runner`:

https://github.com/crossbario/autobahn-python/blob/master/autobahn/twisted/wamp.py#L234
https://github.com/crossbario/autobahn-python/blob/0090f300fa9e2d4a1ef2a44ee15241a9ad9ba4f3/autobahn/wamp/types.py#L109

The actual stopping doesn't run async (it will always immediately return) and doesn't wait for stuff actually stopped. It also doesn't care much if errors during stopping occurs ..

Stopping rlinks should be added (uncommented) in tests. In general, tests for this functionality? eg

```
/home/oberstet/scm/crossbario/crossbar/test/functests/cfctests/test_cfc_rlink_mesh.py:            u"crossbarfabriccenter.remote.router.start_router_realm_link",
/home/oberstet/scm/crossbario/crossbar/test/functests/cfctests/test_cfc_rlinks.py:            result = yield management_session.call('crossbarfabriccenter.remote.router.start_router_realm_link', node_id, worker_id, realm_id, link_id, link_config)
```

Stopping of rlinks (start_router_realm) should be added to the orchestration code
https://github.com/crossbario/crossbar/blob/541ffa448225f3ec6d63fea83a997a601270e54f/crossbar/master/arealm/arealm.py#L970

That is: once an **arealm** is stopped via the master node, which first should only marks the arealm as "stopping"

https://github.com/crossbario/crossbar/blob/541ffa448225f3ec6d63fea83a997a601270e54f/crossbar/master/arealm/arealm.py#L1527
https://github.com/crossbario/crossbar/blob/541ffa448225f3ec6d63fea83a997a601270e54f/crossbar/master/arealm/arealm.py#L1566

and then stop all rlinks, and the actual application realm in the respective workers. and while doing so publish meta events

Issue #2002, opened 2022-04-26
Make vmprof a soft-dependency


oberstet:
vmprof is currently a hard dependency

https://github.com/crossbario/crossbar/blob/8521972e03e20b56193046343a9f5209ac81ac5c/requirements-min.txt#L74

however, there are issues on

* python 3.10 (both cpy and pypy)
* freebsd (arm64)

see https://github.com/crossbario/autobahn-python/issues/1550

we should work around above and make vmprof a run-time checked soft-dependency - vmprof is an advanced, optional feature

PR #2035, opened 2022-07-10
make vmprof a soft dependency

Bastian-Krause:
Since vmprof has trouble releasing a python 3.10 capable version [1] and it is only needed for profiling in crossbar, make it a soft dependency.

All occurrences of vmprof in actual code are already guarded by try except constructs around imports and `_HAS_VMPROF` checks around vmprof use.

So it should be safe to drop vmprof from the various reqirements/setup/Makefile files as well as corresponding licensing files.

[1] https://github.com/vmprof/vmprof-python/issues/240

Fixes: #2002


oberstet:
thanks for fixing this - for now. as you mention, unfortunately, if vmprof hasn't catched up yet, let's make it optional (until it is there).

> one thing to remember, we now need to add and explicit vmprof dependency for Docker image building ..

Issue #2053, opened 2022-10-11
PubSub not working between RLink connected routers


Skully17:
We are having an issue with publish in crossbar io 22.6.1 and previous versions with Rlinks. We require Rlinks to forward local topic publications and RPC’s to cloud and vice versa.

Setup:
There is a crossbar router running on a local machine and another running on an AWS machine. Both have connected JS clients using Autobahn-JS which have 4 buttons to:
1. Subscribe to “com.browser.hub”
2. Subscribe to “com.browser.cloud”
3. Publish to “com.browser.hub”
4. Publish to “com.browser.cloud”

Steps:
1. On the cloud machine, I subscribe to “com.browser.cloud” then subscribe to “com.browser.hub”
2. On the local machine, I subscribe to “com.browser.cloud” then subscribe to “com.browser.hub”
3. On the cloud machine, I publish to “com.browser.cloud” then publish to “com.browser.hub”
4. On the local machine, I publish to “com.browser.cloud” then publish to “com.browser.hub”

You can see from the logs that the first 3 publishes are successfully sent to 2 receivers (sub on local and cloud) but the last one only goes to 1 receiver (sub on local):
![image](https://user-images.githubusercontent.com/49399704/195123420-f5c90fde-8a41-4873-9eb2-0f277754f9f5.png)
![image](https://user-images.githubusercontent.com/49399704/195123943-254b42f4-784b-4abe-a0e5-4fdf839f4236.png)
![image](https://user-images.githubusercontent.com/49399704/195124022-b17a340e-2d63-4b27-82b8-9ad4c6866867.png)
![image](https://user-images.githubusercontent.com/49399704/195124070-1f7b021b-e81a-4b70-8104-78465fe347c3.png)
This means that the second subscription is not being forwarded over the RLink.
When making the subscriptions, this error message always accompanies subscriptions that are failed to be forwarded:
![image](https://user-images.githubusercontent.com/49399704/195123791-dc18af64-b185-4d37-9c34-045d4c08c361.png)
I believe that there is an issue with the syncing of sub IDs over RLinks.

Skully17:
I found a poblem with how the subscription was being handled in rlink.py. I believe the wrong ID was being handled and stored for subscriptions being sent over the rlink.
I changed the subscription code to work similar to registering, which was fixed in #1789, and it works for me.
I have also made a PR for this change: #2055

PR #2055, opened 2022-10-13
Fix for #2053

Skully17:
- Fixed PubSub over RLinks by using "sub_details['id']" instead of "sub_id".
- changed "sub_id" to "sub_session"

These changes follow the fix made to registering over RLinks in #1789

### DIFF
--- a/crossbar/worker/rlink.py
+++ b/crossbar/worker/rlink.py
@@ -67,7 +67,7 @@ class BridgeSession(ApplicationSession):
             other=other)
 
         @inlineCallbacks
-        def on_subscription_create(sub_id, sub_details, details=None):
+        def on_subscription_create(sub_session, sub_details, details=None):
             """
             Event handler fired when a new subscription was created on this router.
 
@@ -79,14 +79,14 @@ class BridgeSession(ApplicationSession):
             :param details:
             :return:
             """
-            if sub_id in self._subs:
+            if sub_details["id"] in self._subs:
                 # this should not happen actually, but not sure ..
                 self.log.error('on_subscription_create: sub ID {sub_id} already in map {method}',
-                               sub_id=sub_id,
+                               sub_id=sub_details["id"],
                                method=hltype(BridgeSession._setup_event_forwarding))
                 return
 
-            self._subs[sub_id] = sub_details
+            self._subs[sub_details["id"]] = sub_details
 
             uri = sub_details['uri']
             ERR_MSG = [None]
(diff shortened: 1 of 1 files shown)


oberstet:
Wow, kudos for tracking this down! Also thanks for contributing in general=)

----

since this was your first contribution in the repo here: guess you are David, right? in which case, thanks for CAA!

----

rgd the PR: minor whitespace formatting (yapf is unhappy), and the docs failure is network related. I'll just restart that one later.

oberstet:
ok, the change seems to indeed match how it should work:

* RLinks use `BridgeSession` https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/worker/rlink.py#L37 which
* sets up a meta API handler https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/worker/rlink.py#L191 for `wamp.subscription.on_create` and uses `wamp.subscription.get` to get subscription details which
* returns a subscription **object**: https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/router/service.py#L642

oberstet:
> Fixed PubSub over RLinks by using "sub_details['id']" instead of "sub_id".

actually, I'm not sure anymore I understand the issue, because: https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/router/service.py#L643

Skully17:
> guess you are David, right?

Yes :)

> minor whitespace formatting (yapf is unhappy), and the docs failure is network related. I'll just restart that one later.

Cheers. I have changed the formatting and the yapf test is now passing locally (copied the test's command from the error).

> actually, I'm not sure anymore I understand the issue, because:

I don't think I follow what you mean but I have noticed a few things when following the steps on how it works:
First, I noticed that "forward_current_subs" passes "sub_id" into "on_subscription_create" which should, I assume, be the same as the id stored in "sub".
https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/worker/rlink.py#L199

However, when debugging before, the "sub_id" and "sub_details['id']" were not the same when I printed them from "on_subscription_create". Perhaps "on_subscription_create" was being called elsewhere?
This lead me to find that "on_subscription_create" is assigned to a subscription and that the subscription is published to here:
https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/router/broker.py#L873

Here "on_subscription_create" is being given "session._session_id", not "sub_id" like in "forward_current_subs".

I propose that "forward_current_subs" should also be changed to pass session id, instead of "sub_id", into "on_subscription_create". This would mean that is works same as the comparable function for regesters:
https://github.com/crossbario/crossbar/blob/ccbae0a56ef331fe00192455837aa72578eadc54/crossbar/worker/rlink.py#L397

There are also other slight differences, such as there being an assert in "register_current":
image

What further changes would be appropriate for me to make in this PR?
1. Change "sub_id" to "self.session_id" in "on_subscription_create"
2. Add assert to check that "sub_id" and "sub['id']" are the same
3. Move check to see if the URI starts with "wamp." to be done at the start of "on_subscription_create" like it is in "on_registration_create"

Skully17:
I have made the changes I suggested to in my last comment.
They are based off of the changes that were made to the handling of registrations but seemed to have be forgotten to be applied to the handling of subscriptions.
Please let me know if there is anything I've missed or is not quite right

PR #2059, opened 2022-11-08
Refactoring "identity_realm_name_category" to "identify_realm_name_category"

Skully17:
"identity_realm_name_category" was changed to "identify_realm_name_category" in autobahn:
[https://github.com/crossbario/autobahn-python/commit/8c57aff8f0d01b64b0e3fa32ddd5780cd43dd1bd](url)

### DIFF
--- a/crossbar/common/checkconfig.py
+++ b/crossbar/common/checkconfig.py
@@ -20,7 +20,7 @@ txaio.use_twisted()
 
 from autobahn.websocket.util import parse_url
 from autobahn.wamp.message import _URI_PAT_STRICT_NON_EMPTY, _URI_PAT_STRICT_LAST_EMPTY, \
-    _URI_PAT_REALM_NAME, _URI_PAT_LOOSE_EMPTY, identity_realm_name_category
+    _URI_PAT_REALM_NAME, _URI_PAT_LOOSE_EMPTY, identify_realm_name_category
 from autobahn.wamp.uri import convert_starred_uri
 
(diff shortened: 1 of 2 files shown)


Bastian-Krause:
This is a breaking change, so we need to bump the minimum autobahn version, right?

Bastian-Krause:
Ah, seems to be already covered by #2060.

oberstet:
@Bastian-Krause yeah, I am currently working to consolidate and test a new set of releases .. I am still working on crossbar, but autobahn-python and txaio should be good already (if nothing pops up from crossbar requiring another patch .. but I don't expect that)

oberstet:
another release related note: both autobahn-python and txaio are now tested up to python 3.11 in various flavors https://github.com/crossbario/autobahn-python/actions/runs/3925457332

as said, I am still hashing out crossbar, but it will be compatible and tested on the same set fwiw, because of the amount of deps, and because python dep mgmt is seemi working, this is annoying and tricky, but I'm getting there ..

oberstet:
yeah, we should do a release .. and fact, I was on it, but got stopped by the upstream stuff mentioned here https://github.com/crossbario/autobahn-python/pull/1617 before a new autobahn release (which should come first, then crossbar ..). sorry

janrinze:
Looks like it's not easy to keep both autobahn and crossbar in sync.
Keep up the good work and hopefully we will see a new release soon.

PR #2067, opened 2023-02-26
doc: notes on updating certificates and restarting.

### LINKED ISSUE #1122: Note about need for restart when updating TLS certs/keys
The following should be added to http://crossbar.io/docs/TLS-Certificates/

"""
After updating Crossbar.io TLS server certificates and/or keys, you must restart Crossbar.io for the new certificates/keys to take effect. The certificates/keys are cached when loading the first time, and hence without restarting Crossbar.io, the old certificates/keys would still be used despite the new ones already being stored on disk.
"""

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46791752-note-about-need-for-restart-when-updating-tls-certs-keys?utm_campaign=plugin&utm_content=tracker%2F462544&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F462544&utm_medium=issues&utm_source=github).
</bountysource-plugin>

2e0byo:
The first paragraph is lifted from #1122

Closes #1122.

Certbot can actually get the original certificates too.  I don't recall ever calling `letsencrypt` manually.  And the installation notes here are very server specific: letsencrypt has a (slightly obsessive) interface to get the 'right' docs up for a particular platform and it's probably better to defer to that.

### DIFF
--- a/docs/TLS-Certificates.rst
+++ b/docs/TLS-Certificates.rst
@@ -488,3 +488,32 @@ fit your system .. this works for Ubuntu):
         Verify return code: 0 (ok)
     ---
     ^C
+
+Updating Certificates
+=====================
+
+After updating Crossbar.io TLS server certificates and/or keys, you must restart
+Crossbar.io for the new certificates/keys to take effect. The certificates/keys
+are cached when loading the first time, and hence without restarting
+Crossbar.io, the old certificates/keys would still be used despite the new ones
+already being stored on disk.
+
+Updating LetsEncrypt with certbot
+---------------------------------
+
+Once you have configured LetsEncrypt, you can periodically run `certbot
+<https://certbot.eff.org/>`_ to update your certificates.  You may wish to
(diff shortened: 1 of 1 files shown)


2e0byo:
Only one of the CI failures is related, but it seems I can't write valid rst any more.  I'll fix it and force-push later.

2e0byo:
I think that's it---I forgot the blank line after the code directive.  Would you like me to overhaul the section on getting the certificate too?  I really think it's better to let certbot handle the whole thing, so I'd just add a link to certbot's own docs and then an example walkthrough (since it's the same once you've got it installed).

oberstet:
looks good, it is an improvement .. even though the docs in general/total is a bit of a mess:( anyways, thanks for contributing!

PR #2091, opened 2023-08-23
Updating dependencies

oberstet:

Issue #2092, opened 2023-08-23
Move CI fully to hosted GH runners again


oberstet:
currently, the CI is (mostly) still setup to run on a self-hosted runner which was deployed to a machine operated by Crossbar.io GmbH - which no longer exists.

PR #2093, opened 2023-08-24
common/reloader: import reload from importlib (Python 3.12 compatibility)

Bastian-Krause:
The imp module is deprecated since Python 3.4 [1]. The import was added even after it was deprecated in [2]. In Python 3.12 the imp module will be removed.

Use the reload function from importlib, which is the designated replacement [1].

With #2091 and this change, crossbar master can be installed and run on Python 3.12 with:
- numpy>=1.26.0b1
- wsaccel @ git+https://github.com/methane/wsaccel@v0.6.4 (assuming cython>=3.0.0 is installed, tarball from pypi won't work [3])
- `AIOHTTP_NO_EXTENSIONS=1` provided for pip install, needed until aiohttp 4.0.0 is released to pypi.

[1] https://docs.python.org/3.11/library/imp.html
[2] https://github.com/crossbario/crossbar/commit/21910b070c79016f7aa5ba62f897ba4f978f593f
[3] https://github.com/methane/wsaccel/issues/30

### DIFF
--- a/crossbar/common/reloader.py
+++ b/crossbar/common/reloader.py
@@ -14,7 +14,7 @@ try:
     reload
 except NameError:
     # Python 3
-    from imp import reload
+    from importlib import reload
 
 __all__ = ('TrackingModuleReloader', )
 


oberstet:
thanks a lot for the detailed infos in the PR!

so just to be sure: crossbar supports python 3.7+ (https://github.com/crossbario/crossbar/blob/03d7057ee18a0d1e59b33bf1633ee246c1cf908a/setup.py#L107), this change is expected to just work, and hence shouldn't break stuff, right?

if so, I'd just merge it - the CI won't run on GH because of https://github.com/crossbario/crossbar/issues/2092

Bastian-Krause:
> so just to be sure: crossbar supports python 3.7+ (
> 
> https://github.com/crossbario/crossbar/blob/03d7057ee18a0d1e59b33bf1633ee246c1cf908a/setup.py#L107
> 
> ), this change is expected to just work, and hence shouldn't break stuff, right?

Right, `importlib.reload` is part of the Python standard library [since 3.4](https://docs.python.org/3/library/importlib.html#importlib.reload). No breakage expected.

> if so, I'd just merge it - the CI won't run on GH because of #2092

Alright.

threexc:
@oberstet do you know if there will be a new release soon containing these changes? Fedora 39 is having issues, as will any other distros that start shipping Python 3.12.

oberstet:
@threexc here is the long explanation: there definitely will be a new release, and Python 3.12 will definitely be fully supported of course. as will be PyPy. and ARM64. and development will continue (e.g. I'm also involved in WAMP spec level work and new and polished things there).

the main work tbd here isn't in the unmerged PRs, but in again fixing the CI. the reason why this has turned out to be a significant source of work over the years is: the amount and breadth of dependencies this package includes, the fact that it is quite a non-trivial package, and the deficiencies in Python dependency management in general (which is IMO one of the biggest language/run-time level issues remaining after the move to Unicode and async) in the ecosystem, but that's only a personal opinion and side note. 

the latter has over the years quite disillusioned myself in any hope that someone (eg myself) would be able to fix all this _once and for all_ in whatever big effort required. once and for all. _If_ I would believe in such a possibility, I might get my act and motivation together and just do it. contrary I do know it is significant, nerve wrecking work and it will remain so. sadly I am not paid for it and unfortunately I am living in a world where I have to pay bills, so I am busy with other things that do in fact earn bucks.

at some point I will personally feel the pain and get fed up by lack of Python 3.12, and then I will fix it no matter what.
since I personally run Crossbar.io on either vanilla upstream Python source, Docker or snaps, on both x86 and ARM, I am not bound to any distro cycles though. IOW: I cannot give you a date. sorry.

threexc:
> @threexc here is the long explanation: there definitely will be a new release, and Python 3.12 will definitely be fully supported of course. as will be PyPy. and ARM64. and development will continue (e.g. I'm also involved in WAMP spec level work and new and polished things there).
> 
> the main work tbd here isn't in the unmerged PRs, but in again fixing the CI. the reason why this has turned out to be a significant source of work over the years is: the amount and breadth of dependencies this package includes, the fact that it is quite a non-trivial package, and the deficiencies in Python dependency management in general (which is IMO one of the biggest language/run-time level issues remaining after the move to Unicode and async) in the ecosystem, but that's only a personal opinion and side note.
> 
> the latter has over the years quite disillusioned myself in any hope that someone (eg myself) would be able to fix all this _once and for all_ in whatever big effort required. once and for all. _If_ I would believe in such a possibility, I might get my act and motivation together and just do it. contrary I do know it is significant, nerve wrecking work and it will remain so. sadly I am not paid for it and unfortunately I am living in a world where I have to pay bills, so I am busy with other things that do in fact earn bucks.
> 
> at some point I will personally feel the pain and get fed up by lack of Python 3.12, and then I will fix it no matter what. since I personally run Crossbar.io on either vanilla upstream Python source, Docker or snaps, on both x86 and ARM, I am not bound to any distro cycles though. IOW: I cannot give you a date. sorry.

No worries, thank you for the summary!

PR #2090, opened 2023-08-23
Fix for #2079

oberstet:
- Corrected param name
- Added variable for repetitively accessed dictionary value
- extended if statement to also check if "sub" is not None to see if the subscription has actually been made.
- Added on_remote_leave to empty dictionary of forwarded subscriptions
- extended if statement to also check if "reg" is not None to see if the registration has actually been made.
- Made "create forwarding registration" log debug level to match the subscription one. This is expected functionality and there can be large amounts of registrations which would cause log spam.

Issue #2079, opened 2023-05-19
Subscriptions And Registration Not Being Forwarded Over RLink On Reconnect Or Late Join


Skully17:
Hello,

I have a system that uses a local Crossbar router which connects, using RLink, to a remote Crossbar router on the cloud. The remote router needs to be able to disconnect and reconnect at any time and to be able to connect later after the initial boot. I tested these scenarios out and found that WAMP tasks on the remote side did not work after reconnecting or when connecting after boot.

After poking around at the code, all in rlink.py, I found two issues:

With the first issue, here is what I think happens. When a subscription is made, on_subscription_create is called and the sub’s ID is stored in the _subs dictionary. The actual subscription is only made and stored in this dictionary if there is an active  RLink connection. When an RLink connection is made, for example after a temporary disconnection, forward_current_subs is called and on_subscription_create is called for every single existing subscription. There is then a check in on_subscription_create that stops the creation if the sub ID already exists in _subs. This means that, even though the sub was not made in the first place due to the remote router not being connected, it assumes that the subscription has already been made and doesn’t try to create it.
The same problem happens for registrations as well.

The second issue was much simpler. When the remote router disconnects the subscriptions aren’t removed from the _subs dictionary. When the remote router reconnects, no subscriptions are forwarded to the remote router because they already exist in _subs.
This problem was fixed for registrations by adding an on_leave that just sets _regs to an empty dictionary when the remote router disconnects. I have done a similar thing for subscriptions.

oberstet:
first of all, kudos for diving that deep into rlinks and router-to-router operations;)

the code for rlinks, including the features that it depends on (such as the meta API) is pretty complex. there will be bugs. however, it is - I would say - quite organized, and in fact, the problems you described are likely in the file you mentioned

https://github.com/crossbario/crossbar/blob/master/crossbar/worker/rlink.py

I'm currently busy with other things and hence didn't look deeply into what you raised and mention, however, from my point of view, one general comment:

The bugs still there are related to complex situations including recovery of node failures in a network of nodes and such, and _reliably_ fixing them _without_ introducing new bugs will _require_ automated tests.

Poking around in the code without automated test cases will unlikely succeed (because of the complexity).

My overarching, general main worry hence is: how do we come up with a suitable set of test scenarios and automated tests to cover router-to-router app messaging ... to test the promises r2r makes wrt apps?

Issue #2076, opened 2023-05-02
Internal Error Attempting Rlink Forwarding


Skully17:
Hello,

I have a simple setup of a “local” Crossbar router which connects to a “remote” Crossbar router using Rlink. When calling/publishing from the “remote” router to a URL that was registered/subscribed to, I am getting this error:
`RLinkRemoteSession.onUserError(): "RuntimeError: Internal error attempting rlink forwarding"`

I can see that this is being raised here and it’s because details does not have a caller, caller_authrole or caller_authid:
https://github.com/crossbario/crossbar/blob/03d7057ee18a0d1e59b33bf1633ee246c1cf908a/crossbar/worker/rlink.py#L292

I can’t work out why these values are not being set or even how they are set at all. How are details passed to the on_call method here:
https://github.com/crossbario/crossbar/blob/03d7057ee18a0d1e59b33bf1633ee246c1cf908a/crossbar/worker/rlink.py#L284

I can see that this method is registered here but details isn't passed there, just reference in RegisterOptions:
https://github.com/crossbario/crossbar/blob/03d7057ee18a0d1e59b33bf1633ee246c1cf908a/crossbar/worker/rlink.py#L344

Does RegisterOptions supply the details?

Just to be clear, the same process works when the “local” router calls/registers to a URI that has been registered/subscribed to on the “remote” router and I can see that the caller variables in details have values.

Skully17:
This comment made here describes exactly what tripped me up:
https://github.com/crossbario/crossbar/issues/1909#issuecomment-974705712

Would it be agreeable if the error message "Internal error attempting rlink forwarding" was edited to include a suggestion to change the role's disclosures to "true" if this is not expected behaviour?

Issue #1959, opened 2022-03-16
rlinks: support wildcard registrations and subscription


om26er:
Rlinks currently don't support wildcard subscriptions and registrations. I think it's an important missing feature and something we should implement.

om26er:
@oberstet can you share your thoughts on the implementation design for this one ?

oberstet:
I haven't thought about this at the code/design level and, fwiw, I would first solve https://github.com/crossbario/crossbar/issues/1903, as otherwise testing will become very hard

om26er:
> I would first solve https://github.com/crossbario/crossbar/issues/1903, as otherwise testing will become very hard
Agreed. I will start working on that and will continue the conversation on that issue. Once we have some sort of automation in place, then will come back to this issue

oberstet:
also, there are some existing automated tests which are currently deactivated (because sth with the CI wasn't working):

https://github.com/crossbario/crossbar/blob/5c3d0e44e7e2b72357b7364ab8e945e08a994b27/.github/workflows/main.yml#L156
https://github.com/crossbario/crossbar/tree/master/test

PR #2095, opened 2023-09-01
RLink fixes and enhancements

DZabavchik:
Although this has been tested extensively this PR is for evaluation/discussion only.  

This PR is superseding #2090 and #2091

Issues that it addresses for some setups (tested with single route all-to-all mesh (not forwarding from other RLinks)

#2079 Forwarding subscriptions after reconnects
#2076 Internal error when attempting forwarding
#1894 Loop in invocation

Adds:
- #1959 Support for prefix and wildcard registrations and subscriptions
- Forced re-registration 
- Role override for RLink on local router:
   - When RLinkLocalSession is `trusted` it can "see" registrations and subs from other RLinks - making it an RLink forwarding other RLinks 
   - When RLinkLocalSession is 'rlink` it doesn't see registrations/subscriptions from other RLinks

### DIFF
--- a/crossbar/router/broker.py
+++ b/crossbar/router/broker.py
@@ -124,7 +124,7 @@ class Broker(object):
 
             for subscription in self._session_to_subscriptions[session]:
 
-                was_subscribed, was_last_subscriber = self._subscription_map.drop_observer(session, subscription)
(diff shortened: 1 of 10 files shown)


DZabavchik:
Will refile after bug-fix PRs to dealer/broker

Issue #2096, opened 2023-09-04
Slow memory leak in dealer.py


DZabavchik:
Finally found this leak that was haunting for 7+ years

https://github.com/crossbario/crossbar/blob/0089c1ef6fbbb87fc7316088a91f1859fa84eeb0/crossbar/router/dealer.py#L1135-L1139

Zombie invocations remain in `_invocations_by_call` with references to callee, caller, authorization that will not be GC'd

Testing changes that preserve original caller session ID in `InvocationRequest` object, so it can be deleted by key from  `_invocations_by_call` with
`del self._invocations_by_call[invocation_request.caller_session_id, invocation_request.call.request]`

Will submit PR shortly

PR #2097, opened 2023-09-04
Bug #2096 memory leak in dealer.py

DZabavchik:
Fixes memory leak in dealer where `InvocationRequest` are not removed from `_invocations_by_call` dict when caller is already gone.  Invocation requests retain references to session objects and others, exacerbating the leak.

```
Partition of a set of 6356 objects. Total size = 779728 bytes.
 Index  Count   %     Size   % Cumulative  % Kind (class / dict of class)
     0   1158  18   227160  29    227160  29 dict (no owner)
     1    163   3    72072   9    299232  38 set
     2    735  12    64030   8    363262  47 str
     3     50   1    58800   8    422062  54 dict of
                                             crossbar.router.protocol.WampRawSocketServerProtocol
     4    887  14    54280   7    476342  61 tuple
     5    339   5    21696   3    498038  64 types.MethodType
     6     54   1    21600   3    519638  67 dict of crossbar.router.session.RouterSession
     7    270   4    21600   3    541238  69 functools.partial
     8    339   5    21520   3    562758  72 list
     9     32   1    15736   2    578494  74 types.FrameType
```

### DIFF
--- a/crossbar/router/dealer.py
+++ b/crossbar/router/dealer.py
@@ -40,6 +40,7 @@ class InvocationRequest(object):
         'id',
         'registration',
         'caller',
+        'caller_session_id',
         'call',
         'callee',
         'forward_for',
@@ -53,6 +54,7 @@ class InvocationRequest(object):
         self.id = id
         self.registration = registration
         self.caller = caller
+        self.caller_session_id = caller._session_id
         self.call = call
         self.callee = callee
         self.forward_for = forward_for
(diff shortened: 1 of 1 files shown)


DZabavchik:
There were a lot more resource leaks found (memory, sockets)

oberstet:
> There were a lot more resource leaks found (memory, sockets)

when you say "were found": did you search for and find leaks? or who did? and how? what's the tool the produced above output? leaks are bad obviously and would need fixing ..

PR #2102, opened 2024-04-03
Fix pydantic version

baileydayson:
Force eth-account to use 0.11.0 so that pydantic doesn't need to be v2


oberstet:
not sure .. open, then close again? ;)

in general, I haven't had time recently to follow eth packages, such as eth-account, and whether they have caught up with some issues / fallout they come with .. which I had "quick fixed" in my non-merged fork ... sorry .. usual OSS / fast moving projects things ...

> crossbar/common/keys/crossbar-24-2.pub

sorry, I would not merge such a change, since you are not the "official source" of the package ..

PR #2121, opened 2025-01-13
fix WAMPCRA auth with salt

Mahad-10:


DZabavchik (approved):

oberstet:
so this is passing all tests including on 3.11/amd64 - and only fails 3.9/amd64 - so I'd assume this is sth we want to ignore at this point - hence, I'm going to force-merge ..

sloe10:
Hello, I think this change actually breaks salting functionality in WAMPCRA. 
Feature was misunderstood by author of the patch. 
Hash must be computed at the client side, router actually know only hashed string, not the real password. 

By this change the password is hashed twice or must be stored in database as a plain text what is really dangerous.

om26er:
> I wanted to say, that this "fix" breaks salting possibility. It should be removed.

please share a piece of code so we can verify that as well.

sloe10:
Please check https://github.com/crossbario/crossbar/issues/2123.

This is wrong: 
  "client2": {
                                            **"secret": "new-secret",**
                                            "role": "frontend",
                                            "salt": "salt123",
                                            "iterations": 100,
                                            "keylen": 16
                                        }

Secret must be already hashed password, otherwise salting does not make any sense.

PR #2122, opened 2025-01-17
Add test to verify WAMPCRA salt authentication

Mahad-10:


oberstet:
thanks a lot!

this adds a new direct, low-level CI driven test for WAMP-CRA _including salting_ against Crossbar.io router authentication, as in against modules

```python
from crossbar.router.role import RouterRoleStaticAuth
from crossbar.router.auth import cryptosign, wampcra, ticket, tls, anonymous

from autobahn.wamp import types
from autobahn.wamp import auth as autobahn_auth
```

and it's added to the right file, so thanks a lot! hope it works. I mean the test in itself, and ideally quite easily then shows a bug in Crossbar.io with salting ..

oberstet:
also, I am wondering whether we have examples / test vectors which include salts in the WAMP spec https://wamp-proto.org/spec.html?

oberstet:
rgd

https://github.com/crossbario/crossbar/actions/runs/12828901013/job/35790689804?pr=2122#step:8:579 

so what _is_ expected?

and for what reason is the value that is expected claimed to be valid?

is it a "test vector" ... from where?

Issue #2128, opened 2025-06-18
RemoteSession not unregistering on leave


Skully17:
When the RemoteSession leaves it is unsubscribing from everything but doesn't unregister. This means that new (correct) registerations can't be made when the rlink connection is re-established.

PR #2129, opened 2025-06-18
Unregistering when RemoteSession leaves

Skully17:
Fix for #2128


oberstet:
thanks for the fix! LGTM. wondering how why/how this obvious code gap could be missed in the first place - with very similar / analog code for subscriptions right above. well, who knows, should be good now;)

DZabavchik:
I'm happy to see activity here.  Will be creating few bugs and PRs.

PR #2153, opened 2026-01-03
Test/caller disconnect callee crash

markope:
a rundown on what the issue is and how it impacts the callee:

### Exception Flow: Caller Disconnect Causes Callee Session Crash

When a caller disconnects during an in-flight RPC, the callee's session can be terminated due to unhandled exception propagation:

### The Problem

1. **Caller disconnects** during an in-flight RPC call
2. **Callee completes work** and sends `YIELD` message with the result
3. **Dealer receives YIELD** but the invocation tracking data is stale (points to disconnected caller)
4. **`processYield()` in `dealer.py`** tries to send result to `caller._transport` which is `None`
5. **Exception is raised** → propagates up to `router.process()`
6. **`router.process()` re-raises `ProtocolError`** (see [router.py#L355-356](src/crossbar/router/router.py#L355-356))
7. **Exception bubbles to Autobahn's `WampWebSocketProtocol.onMessage()`**
8. **Autobahn catches `ProtocolError`** and calls `_bailout()` with `CLOSE_STATUS_CODE_PROTOCOL_ERROR`
9. **WebSocket connection is closed** → **Callee's session is terminated**

### Code Path

```
Callee sends YIELD
  → RouterSession.onMessage()
    → router.process()
      → dealer.processYield()
        → Exception (stale invocation / None transport)
      → ProtocolError re-raised
    → Exception propagates (no catch)
  → WampWebSocketProtocol.onMessage() catches ProtocolError
    → _bailout() → _fail_connection()
      → WebSocket closed → Callee session LOST
```

### Impact

Any `ProtocolError` raised in the dealer causes the **entire session's WebSocket connection to be closed**, which is catastrophic for the callee who did nothing wrong.

### The Fix

Clean up invocation tracking when the caller detaches, so that when the callee later calls `processYield()`:
- The invocation is already gone from `_invocations`
- `processYield()` simply logs a debug message and returns `None`
- No exception is raised
- Callee's session remains intact

### The Test

The test creates the scenario sketched above and expects the dealer.py to not raise an error in step 4 (processYield).
Currently without the fix the test does not pass.

### DIFF
--- a/src/crossbar/router/dealer.py
+++ b/src/crossbar/router/dealer.py
@@ -196,24 +196,8 @@ class Dealer(object):
                 if invoke.callee is invoke.caller:  # if the calling itself - no need to notify
                     continue
                 callee = invoke.callee
-                if (
-                    "callee" not in callee._session_roles
-                    or not callee._session_roles["callee"]
-                    or not callee._session_roles["callee"].call_canceling
-                ):
-                    self.log.debug(
-                        "INTERRUPT not supported on in-flight INVOKE with id={request} on"
-                        " session {session} (caller went away)",
-                        request=invoke.id,
-                        session=session._session_id,
-                    )
-                    continue
-                self.log.debug(
-                    "INTERRUPTing in-flight INVOKE with id={request} on session {session} (caller went away)",
-                    request=invoke.id,
-                    session=session._session_id,
-                )
 
+                # Always clean up the invocation tracking regardless of call_canceling support
                 if invoke.timeout_call:
                     invoke.timeout_call.cancel()
                     invoke.timeout_call = None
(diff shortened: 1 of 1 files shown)


markope:
all tests pass now see [here](https://github.com/crossbario/crossbar/actions/runs/20679011162/job/59370561885?pr=2153)

oberstet:
fantastic! this is a showcase example for the workflow!

it's perfect - both for me reviewing, as well as (IMO) for users.

thanks also in particular for the detailed comments on the new test `test_caller_detach_cleans_up_invocation_tracking_without_cancel_support()`!

ah, and the remaining failed code quality check was completely unrelated to this PR.

merging.

oberstet:
love it. thanks so much for the fix! and for taking so much care following the workflow.

fwiw, now (together with fixed typo wrt check-typing), workflows are 100% green & happy:

image

---

the `release.yml` - correctly - posted a new nightly/development github release: https://github.com/crossbario/crossbar/releases/tag/master-202601040913

image

---

PR #2255, opened 2026-07-15
Fix 2250

### LINKED ISSUE #2250: [BUG] WebSocket size-limit config: fill the empty compression validator, clarify compressed-vs-uncompressed docs, fix the zip-bomb functest and put it in CI
## Problem

Three related gaps on the Crossbar side of the transport size-limit story
(depends on the Autobahn fixes in META #TBD / draft 00):

1. **Empty compression validator.** `check_websocket_compression()` is a stub —
   its body is just `# FIXME` (`src/crossbar/common/checkconfig.py:1593-1600`), so
   WebSocket `compression` options are effectively unvalidated.
2. **Ambiguous docs.** `max_message_size` / `max_frame_size` are documented only
   as "incoming WebSocket messages/frames accepted" (`docs/WebSocket-Options.rst:61,63`)
   with no statement of whether the limit is on **compressed (wire)** or
   **uncompressed (application)** size. After the Autobahn fix the answer is
   well-defined — `max_message_size` bounds the **uncompressed reassembled**
   message (the serialized WAMP message), `max_frame_size` bounds the **wire**
   frame — and the docs should say so. The compression docs
   (`docs/WebSocket-Compression.rst`) never mention decompression-bomb risk.
3. **Zip-bomb functest passes for the wrong reason and is not in CI.**
   `test/functests/cbtests/test_cb_zip_bomb.py` asserts that an oversized
   decompressed message drops the connection — but with the current (broken)
   deflate cap that drop comes from a `zlib -3` decode error on truncated data
   (see Autobahn child 01), **not** a clean size-limit rejection. The test's own
   comments ("turning on Deflate seems .. hard", "why do I have to specify max_...
   in at least 2 places?") flag the API confusion the Autobahn work resolves.

## Fix (approach)

- **Validator:** implement `check_websocket_compression` to validate the
  `deflate` (and any other) compression option blocks — permitted keys, types,
  ranges — consistent with what `set_websocket_options` actually reads
  (`router/protocol.py:83-115`).
- **Docs:** state the semantics explicitly in `WebSocket-Options.rst` and
  `RawSocket-Transport.rst` — `max_message_size` = max uncompressed serialized
  WAMP message; `max_frame_size` = max wire frame — and add a short
  decompression-bomb note to `WebSocket-Compression.rst` pointing operators at
  `max_message_size` as the effective guard once Autobahn enforces it uncompressed.
- **Functest:** rewrite `test_cb_zip_bomb.py` to assert a **clean** rejection
  (peer closes with `MESSAGE_TOO_BIG` / `wasMaxMessagePayloadSizeExceeded`), not a
  decode error; drive it against the fixed Autobahn. Add it to a CI job (ties into
  the Track T "functests into CI" work).

## Red/green test plan (TDD, on this PR)

- **RED first:** land the rewritten `test_cb_zip_bomb.py` (asserting clean
  `MESSAGE_TOO_BIG`) and a `checkconfig` unit test for the compression validator,
  and prove them red — red on the functest requires the Autobahn fixes (children
  01/02) to be available in the pinned Autobahn; sequence this PR **after** those
  land (or pin a dev build). Document the dependency in the PR.
- **GREEN:** validator implemented + functest passing against fixed Autobahn +
  job wired into CI.
- Note the Autobahn-version dependency explicitly (this is the one child that
  crosses the repo boundary and needs the coordinated Autobahn floor).

## Acceptance criteria

- `check_websocket_compression` validates real config and rejects malformed
  compression blocks (unit-tested).
- Docs state compressed-vs-uncompressed semantics for both transports and note
  the zip-bomb guard.
- `test_cb_zip_bomb.py` asserts a clean size-limit rejection and runs in CI.

## References

- META https://github.com/crossbario/autobahn-python/issues/1907. Depends on Autobahn children
  01 + 02 (and benefits from 03). Ties into the Track T testing epic.

_This work is being completed with AI assistance (Claude Code)._

oberstet:
fixes https://github.com/crossbario/crossbar/issues/2250

### DIFF
--- a/.audit/oberstet_fix_2250.md
+++ b/.audit/oberstet_fix_2250.md
@@ -1,0 +1,8 @@
+- [ ] I did **not** use any AI-assistance tools to help create this pull request.
+- [x] I **did** use AI-assistance tools to *help* create this pull request.
+- [x] I have read, understood and followed the projects' [AI Policy](https://github.com/crossbario/autobahn-python/blob/main/AI_POLICY.md) when creating code, documentation etc. for this pull request.
+
+Submitted by: @oberstet
+Date: 2026-07-15
+Related issue(s): #2250
+Branch: oberstet:fix_2250
--- a/src/crossbar/common/checkconfig.py
+++ b/src/crossbar/common/checkconfig.py
@@ -27,6 +27,7 @@ from autobahn.wamp.message import (
     identify_realm_name_category,
 )
 from autobahn.wamp.uri import convert_starred_uri
+from autobahn.websocket.compress_deflate import PerMessageDeflateMixin
 from autobahn.websocket.util import parse_url
 from yaml import Dumper, Loader, SafeDumper, SafeLoader
 from yaml.constructor import ConstructorError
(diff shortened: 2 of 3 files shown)


oberstet:
image

